PR #1, opened 2024-04-29
Package review PR

KarthikSubbarao:
This is a PR which merges current changes in `unstable` into an empty branch in order to help with a review code of the entire codebase.


src/commands/bloom.rs
```diff
@@ -0,0 +1,357 @@
...
+    let argc = input_args.len();
+    if !(4..=6).contains(&argc) {
+        return Err(RedisError::WrongArity);
+    }
+    let mut curr_cmd_idx = 1;
+    // Parse the filter name
+    let filter_name = &input_args[curr_cmd_idx];
+    curr_cmd_idx += 1;
+    // Parse the error rate
+    let fp_rate = match input_args[curr_cmd_idx].to_string_lossy().parse::<f32>() {
```
hpatro: Does rust panic on invalid value being passed?

src/commands/bloom.rs
```diff
@@ -0,0 +1,357 @@
...
+        {
+            "NONSCALING" if argc == 5 => {
+                expansion = 0;
+            }
+            "EXPANSION" if argc == 6 => {
+                curr_cmd_idx += 1;
+                expansion = match input_args[curr_cmd_idx].to_string_lossy().parse::<u32>() {
+                    Ok(num) if num > 0 && num <= BLOOM_EXPANSION_MAX => num,
+                    _ => {
+                        return Err(RedisError::Str("ERR bad expansion"));
```
hpatro: Like we discussed, do we need to prefix with `ERR`.

src/commands/bloom.rs
```diff
@@ -0,0 +1,357 @@
...
+                Ok(_v) => REDIS_OK,
+                Err(_) => Err(RedisError::Str(ERROR)),
+            }
+        }
+    }
+}
+
+pub fn bloom_filter_insert(ctx: &Context, input_args: &[RedisString]) -> RedisResult {
+    let argc = input_args.len();
+    // At the very least, we need: BF.INSERT <key> ITEM <item>
```
hpatro: Why do we need `ITEM` keyword before the actual item value? Is this how the Redis API is designed?

src/commands/bloom.rs
```diff
@@ -0,0 +1,357 @@
...
+            }
+            let mut bf = BloomFilterType::new_reserved(fp_rate, capacity, expansion);
+            for item in input_args.iter().take(argc).skip(idx) {
+                result.push(RedisValue::Integer(bf.add_item(item.as_slice())));
+            }
+            match filter_key.set_value(&BLOOM_FILTER_TYPE, bf) {
+                Ok(_) => Ok(RedisValue::Array(result)),
+                Err(_) => Err(RedisError::Str(ERROR)),
+            }
+        }
```
hpatro: Could we check the bloom filter exists or not first and then insert the data in a single flow. Don't like the code duplication around insertion.

src/commands/bloom_util.rs
```diff
@@ -0,0 +1,125 @@
+use bloomfilter;
+
+pub const ERROR: &str = "ERROR";
+
+/// The BloomFilterType structure. 32 bytes.
+/// Can contain one or more filters.
+pub struct BloomFilterType {
+    pub expansion: u32,
+    pub fp_rate: f32,
+    pub filters: Vec<BloomFilter>,
```
hpatro: Does the underlying support scalable bloom filters or do we need to support this mechanism?

src/commands/bloom_util.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub fn capacity(&self) -> i64 {
+        let mut capacity = 0;
+        // Check if item exists already.
+        for filter in &self.filters {
+            capacity += filter.capacity;
+        }
+        capacity as i64
+    }
+
+    pub fn add_item(&mut self, item: &[u8]) -> i64 {
```
hpatro: Looks like the response could be bool instead of i64.

src/commands/bloom_util.rs
```diff
@@ -0,0 +1,125 @@
...
+                filter.num_items += 1;
+                return 1;
+            }
+            if filter.num_items >= filter.capacity {
+                let new_capacity = filter.capacity * self.expansion;
+                let mut new_filter = BloomFilter::new(self.fp_rate, new_capacity);
+                // Add item.
+                new_filter.bloom.set(item);
+                new_filter.num_items += 1;
+                self.filters.push(new_filter);
```
hpatro: Shouldn't we fail the request if the expansion is much higher than we can support?

src/wrapper/bloom_callback.rs
```diff
@@ -0,0 +1,79 @@
...
+use redis_module::raw;
+use std::os::raw::{c_char, c_int, c_void};
+use std::ptr::null_mut;
+
+// Note: methods in this mod are for the bloom module data type callbacks.
+// The reason they are unsafe is because the callback methods are expected to be
+// "unsafe extern C" based on the Rust module API definition
+
+/// # Safety
+pub unsafe extern "C" fn bloom_rdb_save(rdb: *mut raw::RedisModuleIO, value: *mut c_void) {
```
hpatro, line 27: I think we should introduce a magic number/version number to easily decode the data based on that. If in future, we introduce some new information to be stored in RDB, we can bump it.

src/commands/bloom_data_type.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        mem_usage2: None,
+        free_effort2: None,
+        unlink2: None,
+        copy2: None,
+    },
+);
+
+pub fn bloom_rdb_load_data_object(
+    rdb: *mut raw::RedisModuleIO,
```
hpatro: This is just one item, right?

src/wrapper/bloom_callback.rs
KarthikSubbarao: Currently, we have already specified an encoding number for this new Module Data type.

In bloom_data_type.rs,  the version is specified as `BLOOM_FILTER_TYPE_ENCODING_VERSION` and is used during module data type creation on module load. We can fetch the version from the rdb `RedisModuleIO` object here and use it for custom logic in the future when needed

src/commands/bloom_util.rs
KarthikSubbarao: `expansion` really means expansion_rate. I can rename this variable if that would help clarify this.

I don't think there is a limit on the number of sub filters an object can have. But (if we want) we can define a config for this to either silently fail expansion (by allowing a `set`) beyond a limit of X sub filters per object OR explicitly return an error (without a `set`).

src/commands/bloom_util.rs
KarthikSubbarao: It could be `bool`. I was just using this to directly return the result of this into the command handler as a response instead of performing a bool to `i64` conversion operation

src/commands/bloom_util.rs
KarthikSubbarao: This particular library does not have auto scaling

https://docs.rs/bloomfilter/1.0.13/bloomfilter/struct.Bloom.html

src/commands/bloom_data_type.rs
KarthikSubbarao: Yes, this is one object / item

src/commands/bloom.rs
KarthikSubbarao: I can create a separate function for multi adds and call it from both flows. I wanted to handle both through the same flow, however it will be out of reference scope and result in a moved error

src/commands/bloom.rs
KarthikSubbarao: No, it will not panic here. If I did not use a `match` operation and instead did an `unwrap` on a Result, it would have panicked and crashed on any invalid input

src/commands/bloom.rs
KarthikSubbarao: Using this `RedisError::Str`, yes, we need to provide a prefix. But I will check for alternatives if any have the prefix auto filled in

src/commands/bloom.rs
KarthikSubbarao: Yes, this is the command syntax for the `BF.INSERT` API

hpatro: Just checked it's `ITEMS` actually.

src/wrapper/bloom_callback.rs
hpatro: In the load flow, I didn't see any failure condition on unsupported encoding version. Shall we add it there?

KarthikSubbarao: Yes, good point. I will add a check to return early if there is an unsupported encoding version

src/commands/bloom.rs
KarthikSubbarao: Good catch. Thanks! I will update it

KarthikSubbarao: Yes, I am already using "ITEMS" in the parsing logic.

But the documentation highlighted here needs to be updated

src/commands/bloom.rs
KarthikSubbarao: I looked at the core module logic and they do not auto populate the `-ERR` prefix. The rust module crate also does not have an API which does this.

So we need to either add `ERR` ourselves (using the existing API) or create an API in the rust module crate which does this

```
/* Reply with the error 'err'.
 *
 * Note that 'err' must contain all the error, including
 * the initial error code. The function only provides the initial "-", so
 * the usage is, for example:
 *
 *     ValkeyModule_ReplyWithError(ctx,"ERR Wrong Type");
 *
 * and not just:
 *
 *     ValkeyModule_ReplyWithError(ctx,"Wrong Type");
 *
 * The function always returns VALKEYMODULE_OK.
 */
int VM_ReplyWithError(ValkeyModuleCtx *ctx, const char *err) {
    client *c = moduleGetReplyClient(ctx);
    if (c == NULL) return VALKEYMODULE_OK;
    addReplyErrorFormat(c,"-%s",err);
    return VALKEYMODULE_OK;
}
```

src/commands/bloom_util.rs
KarthikSubbarao: One other aspect we should handle this checking / rejecting based on memory overhead of every operation that creates a new BloomFilter object (BF.ADD, BF.MADD, BD.RESERVE, BF.INSERT, RDB Load). Before any of these operations, we should probably check the memory usage and reject the operations if there is not sufficient space. We have a mechanism to compute the estimated additional memory overhead per creation.

src/wrapper/bloom_callback.rs
```diff
@@ -0,0 +1,109 @@
...
+use std::ptr::null_mut;
+use valkey_module::raw;
+use valkey_module::{RedisModuleDefragCtx, RedisModuleString};
+
+// Note: methods in this mod are for the bloom module data type callbacks.
+// The reason they are unsafe is because the callback methods are expected to be
+// "unsafe extern C" based on the Rust module API definition
+
+/// # Safety
+pub unsafe extern "C" fn bloom_rdb_save(rdb: *mut raw::RedisModuleIO, value: *mut c_void) {
```
parthpatel, line 27: Why are we not using "https://docs.rust-embedded.org/book/interoperability/rust-with-c.html#no_mangle" as recommended by Rust documentation?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,103 @@
+name: ci
```
parthpatel, line 1: If I am a new developer looking at this file, how do I understand the intent of this file without documentation? Are these configurations obvious? What documentation did you read to write this file?

Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+description = "A bloom filter module for Valkey"
+homepage = "https://github.com/valkey-io/valkey-bloom"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+valkey-module = "0.1.2"
+bloomfilter = "1.0.13"
+lazy_static = "1.4.0"
+libc = "0.2"
```
parthpatel, line 20: Why do we need libc? 
Why do we not depend on "valkey-module-macros" package? (valkey-module-rs github repo)

Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+crate-type = ["cdylib"]
+name = "valkey_bloom"
+
+[profile.dev]
+opt-level = 0
+debug = 2
+debug-assertions = true
+
+[features]
+enable-system-alloc = ["valkey-module/enable-system-alloc"]
```
parthpatel, line 39: What is the intention with this configuration?

requirements.txt
```diff
@@ -0,0 +1,2 @@
+valkey
```
parthpatel, line 1: What does this file do?

src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+fn initialize(_ctx: &Context, _args: &[ValkeyString]) -> Status {
+    Status::Ok
+}
+
+fn deinitialize(_ctx: &Context) -> Status {
+    Status::Ok
+}
+
+/// Command handler for BF.EXISTS <key> <item>
+fn bloom_exists_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
```
parthpatel, line 46: Why do we need these wrapper functions? Can't we feed the commands from command_handler module directly in the macro below?

src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+//////////////////////////////////////////////////////
+
+valkey_module! {
+    name: MODULE_NAME,
+    version: 1,
+    allocator: (valkey_module::alloc::ValkeyAlloc, valkey_module::alloc::ValkeyAlloc),
+    data_types: [
+        BLOOM_FILTER_TYPE,
+    ],
+    init: initialize,
```
parthpatel, line 109: Does the valkey module not have a default? What happens if we don't specify this?

src/bloom/data_type.rs
```diff
@@ -0,0 +1,116 @@
...
+use crate::configs::{
+    FIXED_SIP_KEY_ONE_A, FIXED_SIP_KEY_ONE_B, FIXED_SIP_KEY_TWO_A, FIXED_SIP_KEY_TWO_B,
+};
+use crate::wrapper::bloom_callback;
+use crate::MODULE_NAME;
+use std::os::raw::c_int;
+use valkey_module::native_types::ValkeyType;
+use valkey_module::{logging, raw};
+
+const BLOOM_FILTER_TYPE_ENCODING_VERSION: i32 = 0;
```
parthpatel: Why 0 ?

src/bloom/data_type.rs
```diff
@@ -0,0 +1,116 @@
...
+
+pub static BLOOM_FILTER_TYPE: ValkeyType = ValkeyType::new(
+    "bloomfltr",
+    BLOOM_FILTER_TYPE_ENCODING_VERSION,
+    raw::RedisModuleTypeMethods {
+        version: raw::REDISMODULE_TYPE_METHOD_VERSION as u64,
```
parthpatel, line 22: u64? that seem like an overkill.

parthpatel (commented):
Mostly just questions - trying to understand this code base.

PR #2, opened 2024-09-11
rdb format optimization: using fixed seed for bloom filters

YueTang-Vanessa:
#### Description

Fixed seeds and sip keys can help user create bloom objects using the same seed and to restore (from RDB save and load) the bloom objects with the same hashers(sip keys are generated based on the seed). This will save RDB space (32 bytes per filter and there can be multiple filters per object).

This commit have following changes:
* Introduced a seed constant and sip keys constant which generated separately using the `getrandom` method in rust `bloomfilter` crate
* Only save the "number of items in filter" count in the final bloom filter of a bloom object's vector 

#### Test

Loading the module with `valkey-server`, verified that the rdb can load successfully with same `bf.info key item` and `bf.exists key item` output

```
127.0.0.1:6379> info keyspace
# Keyspace
127.0.0.1:6379> set hello world
OK
127.0.0.1:6379> info keyspace
# Keyspace
db0:keys=1,expires=0,avg_ttl=0
127.0.0.1:6379> bf.add key item
(integer) 1
127.0.0.1:6379> info keyspace
# Keyspace
db0:keys=2,expires=0,avg_ttl=0
127.0.0.1:6379> keys *
1) "hello"
2) "key"
127.0.0.1:6379> bf.exists key item
(integer) 1
127.0.0.1:6379> bf.info key
 1) Capacity
 2) (integer) 100000
 3) Size
 4) (integer) 179952
 5) Number of filters
 6) (integer) 1
 7) Number of items inserted
 8) (integer) 1
 9) Expansion rate
10) (integer) 2
127.0.0.1:6379> bgsave
Background saving started
127.0.0.1:6379> shutdown
not connected> info keyspace
# Keyspace
db0:keys=2,expires=0,avg_ttl=0
127.0.0.1:6379> keys *
1) "key"
2) "hello"
127.0.0.1:6379> bf.exists key item
(integer) 1
127.0.0.1:6379> bf.info key
 1) Capacity
 2) (integer) 100000
 3) Size
 4) (integer) 179952
 5) Number of filters
 6) (integer) 1
 7) Number of items inserted
 8) (integer) 1
 9) Expansion rate
10) (integer) 2
```
Also passed python tests and unit tests
```
running 3 tests
test bloom::utils::tests::test_sip_keys ... ok
test bloom::utils::tests::test_non_scaling_filter ... ok
test bloom::utils::tests::test_scaling_filter ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 9.92s


valkey-bloom/tests/test_basic.py::TestBloomBasic::test_basic PASSED           [ 50%]
valkey-bloom/tests/test_save_and_restore.py::TestBloomSaveRestore::test_basic_save_and_restore PASSED [100%]

================================================================ 2 passed in 3.61 seconds ================================================================
Build and Integration Tests succeeded
```


src/bloom/utils.rs
```diff
@@ -154,7 +154,15 @@ pub struct BloomFilter {
 impl BloomFilter {
     /// Instantiate empty BloomFilter object.
     pub fn new(fp_rate: f32, capacity: u32) -> BloomFilter {
-        let bloom = bloomfilter::Bloom::new_for_fp_rate(capacity as usize, fp_rate as f64);
+        let fixed_seed: [u8; 32] = hex::decode(FIXED_SEED_IN_HEX)
+            .expect("Failed to decode")
+            .try_into()
+            .unwrap();
```
KarthikSubbarao: Could we save the [u8, 32] itself as a constant and simply use it from here? This will help avoid a decode operation which is in the data path of bloom filter creations. It will also help remove the dependency on `hex`

src/configs.rs
```diff
@@ -18,3 +18,10 @@ lazy_static! {
 /// Constants
 pub const BLOOM_FP_RATE_DEFAULT: f32 = 0.001;
 pub const TIGHTENING_RATIO: f32 = 0.8515625;
+// generate using rust crate bloomfilter https://crates.io/crates/bloomfilter
```
KarthikSubbarao: It will help to add a `///` documentation block here to explain that these are fixed seeds and sip keys to help us create bloom objects using the same seed and to restore (from RDB save and load) the bloom objects with the same hashers because we are using sip keys which are computed from the fixed seed. This will save RDB space (32 bytes per filter and there can be multiple filters per object).

Also, can we add this motivation and reasoning to the PR description?

src/bloom/data_type.rs
```diff
@@ -72,28 +75,15 @@ pub fn bloom_rdb_load_data_object(
...
-        };
-        let Ok(sip_key_two_a) = raw::load_unsigned(rdb) else {
-            return None;
-        };
-        let Ok(sip_key_two_b) = raw::load_unsigned(rdb) else {
-            return None;
-        };
         let Ok(num_items) = raw::load_unsigned(rdb) else {
             return None;
         };
```
KarthikSubbarao: We only write `num_items` on the final bloom filter in the vector (based on the logic of RDB Save in this PR). We should not be loading an unsigned integer for num_items unless this is the final bloom filter.

For all the "intermediate" filters, we can use capacity as the num_items because they will be filled to capacity before any scale out occurs.

src/bloom/utils.rs
```diff
@@ -154,7 +154,11 @@ pub struct BloomFilter {
 impl BloomFilter {
     /// Instantiate empty BloomFilter object.
     pub fn new(fp_rate: f32, capacity: u32) -> BloomFilter {
-        let bloom = bloomfilter::Bloom::new_for_fp_rate(capacity as usize, fp_rate as f64);
+        let bloom = bloomfilter::Bloom::new_for_fp_rate_with_seed(
+            capacity as usize,
+            fp_rate as f64,
+            &FIXED_SEED,
```
KarthikSubbarao, line 172: Can we add a unit test in this file to validate that when we use `FIXED_SEED` to create bloom filters, we can use the `sip_keys` function and it will return the same sip_keys values (FIXED_SIP_KEY_ONE_A, ONE_B, TWO_A, TWO_B) every time?

I updated the package version and we can now run unit tests using: `cargo test --features enable-system-alloc`

src/bloom/utils.rs
```diff
@@ -349,7 +355,36 @@ mod tests {
...
+        // Verify restore
+        let mut restore_bf = BloomFilterType::create_copy_from(&bf);
+        assert_eq!(
+            restore_bf.add_item(b"new_item"),
+            Err(BloomError::NonScalingFilterFull)
+        );
+        assert_eq!(restore_bf.capacity(), initial_capacity as i64);
+        assert_eq!(restore_bf.cardinality(), initial_capacity as i64);
+        assert_eq!(restore_bf.free_effort(), 1);
+        assert!(restore_bf.memory_usage() > 0);
```
KarthikSubbarao: Can we compare that they have the same memory usage? (I am guessing they should be equal)

src/bloom/utils.rs
```diff
@@ -349,7 +355,36 @@ mod tests {
...
+
+        // Verify restore
+        let mut restore_bf = BloomFilterType::create_copy_from(&bf);
+        assert_eq!(
+            restore_bf.add_item(b"new_item"),
+            Err(BloomError::NonScalingFilterFull)
+        );
+        assert_eq!(restore_bf.capacity(), initial_capacity as i64);
+        assert_eq!(restore_bf.cardinality(), initial_capacity as i64);
+        assert_eq!(restore_bf.free_effort(), 1);
```
KarthikSubbarao: For all of these, we can compare with the original object (`bf`)

src/bloom/utils.rs
```diff
@@ -410,6 +445,54 @@ mod tests {
...
+        );
+        assert_eq!(
+            restore_bloom_filter_type.cardinality(),
+            expected_restore_total_capacity as i64
+        );
+        assert_eq!(
+            restore_bloom_filter_type.free_effort(),
+            num_filters_to_scale as usize
+        );
+        assert!(restore_bloom_filter_type.memory_usage() > 0);
```
KarthikSubbarao: For all of these, we can compare with the original object (`bf`)

src/bloom/utils.rs
```diff
@@ -349,7 +355,36 @@ mod tests {
...
+        );
+        item_exists_test(
+            &restore_bf,
+            add_operation_idx + 1,
+            add_operation_idx * 2,
+            expected_fp_rate,
+            fp_margin,
+            false,
+            &rand_prefix,
+        );
```
KarthikSubbarao, line 446: Comparing the inner properties (number of hash functions, bits) between the original and restore objects will validate the correctness of the restoration. Could we add this?

YueTang-Vanessa: Added comparison of sip_keys, number_of_hash_functions and bitmap.

zackcam (approved):

src/bloom/utils.rs
```diff
@@ -349,7 +355,49 @@ mod tests {
...
+            Err(BloomError::NonScalingFilterFull)
+        );
+        assert_eq!(restore_bf.capacity(), bf.capacity());
+        assert_eq!(restore_bf.cardinality(), bf.cardinality());
+        assert_eq!(restore_bf.free_effort(), bf.free_effort());
+        assert_eq!(restore_bf.memory_usage(), bf.memory_usage());
+        assert!(restore_bf.filters.iter().all(|restore_filter| bf
+            .filters
+            .iter()
+            .any(|filter| filter.bloom.sip_keys() == restore_filter.bloom.sip_keys())));
```
KarthikSubbarao: Can we also check that this matches the fixed sip_keys by comparing with `expected_sip_keys`?
We can update the IF statement to: A == B and B == C
```
        let expected_sip_keys = [
            (FIXED_SIP_KEY_ONE_A, FIXED_SIP_KEY_ONE_B),
            (FIXED_SIP_KEY_TWO_A, FIXED_SIP_KEY_TWO_B),
        ];
```

src/bloom/utils.rs
```diff
@@ -410,6 +458,64 @@ mod tests {
...
+        assert_eq!(restore_bloom_filter_type.cardinality(), bf.cardinality(),);
+        assert_eq!(restore_bloom_filter_type.free_effort(), bf.free_effort());
+        assert_eq!(restore_bloom_filter_type.memory_usage(), bf.memory_usage());
+        assert!(restore_bloom_filter_type
+            .filters
+            .iter()
+            .all(|restore_filter| bf
+                .filters
+                .iter()
+                .any(|filter| filter.bloom.sip_keys() == restore_filter.bloom.sip_keys())));
```
KarthikSubbarao: Can we also check that this matches the fixed sip_keys by comparing with `expected_sip_keys`?
We can update the IF statement to: A == B and B == C
```
        let expected_sip_keys = [
            (FIXED_SIP_KEY_ONE_A, FIXED_SIP_KEY_ONE_B),
            (FIXED_SIP_KEY_TWO_A, FIXED_SIP_KEY_TWO_B),
        ];
```

src/bloom/utils.rs
```diff
@@ -410,6 +458,64 @@ mod tests {
...
+                .iter()
+                .any(|filter| filter.bloom.number_of_hash_functions()
+                    == restore_filter.bloom.number_of_hash_functions())));
+        assert!(restore_bloom_filter_type
+            .filters
+            .iter()
+            .all(|restore_filter| bf
+                .filters
+                .iter()
+                .any(|filter| filter.bloom.bitmap() == restore_filter.bloom.bitmap())));
```
KarthikSubbarao: This works but it looks like a lot of repeated code (for both unit tests). Can we simply iterate over each filter using a loop and then compare all the properties per iteration?

KarthikSubbarao (approved):

PR #3, opened 2024-09-17
Add unit testing for scaling & non scaling filters

KarthikSubbarao:
PR to add unit testing of the BloomFilter structure's behavior in case of scaling and non scaling filter and also to test for false positive rate correctness. Non scaling and scaling filters are created at scale to filled with items to capacity and exists operations are used to validate fp_rate correctness.


YueTang-Vanessa (approved):
LGTM

zackcam (approved):

PR #7, opened 2024-10-03
Integration tests for commands behavior, deletion, expiry, compatibility

YueTang-Vanessa:
This commit includes following changes:

* Add test for bloom command arity, behavior and basic error
* Set up a base case for valkey bloom filter module
* Set info of expansion to Null for NONSCALING mode
* Optimize bf.insert error handling
* Add test for basic valkey command
* Add test for compatibility

### DIFF
--- a/src/bloom/utils.rs
+++ b/src/bloom/utils.rs
@@ -10,13 +10,14 @@ pub const ERROR: &str = "ERROR";
 pub const NON_SCALING_FILTER_FULL: &str = "ERR non scaling filter is full";
 pub const NOT_FOUND: &str = "ERR not found";
 pub const ITEM_EXISTS: &str = "ERR item exists";
-pub const INVALID_INFO_VALUE: &str = "ERR Invalid information value";
+pub const INVALID_INFO_VALUE: &str = "ERR invalid information value";
 pub const BAD_EXPANSION: &str = "ERR bad expansion";
 pub const BAD_CAPACITY: &str = "ERR bad capacity";
 pub const BAD_ERROR_RATE: &str = "ERR bad error rate";
 pub const ERROR_RATE_RANGE: &str = "ERR (0 < error rate range < 1)";
 pub const CAPACITY_LARGER_THAN_0: &str = "ERR (capacity should be larger than 0)";
 pub const MAX_NUM_SCALING_FILTERS: &str = "ERR max number of scaling filters reached";
+pub const UNKNOWN_ARGUMENT: &str = "ERR unknown argument received";
 
 #[derive(Debug, PartialEq)]
 pub enum BloomError {
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -1,18 +1,13 @@
 import time
 import pytest
+from util.waiters import *
 from valkey import ResponseError
-from valkeytests.valkey_test_case import ValkeyTestCase
+from valkey_bloom_test_case import ValkeyBloomTestCaseBase
 from valkeytests.conftest import resource_port_tracker
 import logging
 import os
 
-class TestBloomBasic(ValkeyTestCase):
-
-    def get_custom_args(self):
-        self.set_server_version(os.environ['SERVER_VERSION'])
-        return {
-            'loadmodule': os.getenv('MODULE_PATH'),
-        }
+class TestBloomBasic(ValkeyBloomTestCaseBase):
 
     def test_basic(self):
         client = self.server.get_new_client()
@@ -31,3 +26,98 @@ class TestBloomBasic(ValkeyTestCase):
         assert bf_exists_result == 1
         bf_exists_result = client.execute_command('BF.EXISTS filter1 item2')
         assert bf_exists_result == 0
+
+    def test_bloom_modification(self):
+        client = self.server.get_new_client()
+        # check bloom filter with basic valkey command
+        # cmd touch
+        assert client.execute_command('BF.ADD key1 val1') == 1
+        assert client.execute_command('BF.ADD key2 val2') == 1
+        assert client.execute_command('TOUCH key1 key2') == 2
+        assert client.execute_command('TOUCH key3') == 0
+        self.verify_key_number(client, 2)
+
+    def test_bloom_transaction(self):
+        client = self.server.get_new_client()
+        # cmd multi, exec
+        assert client.execute_command('MULTI') == b'OK'
+        assert client.execute_command('BF.ADD M1 V1') == b'QUEUED'
+        assert client.execute_command('BF.ADD M2 V2') == b'QUEUED'
+        assert client.execute_command('BF.EXISTS M1 V1') == b'QUEUED'
+        assert client.execute_command('DEL M1') == b'QUEUED'
+        assert client.execute_command('BF.EXISTS M1 V1') == b'QUEUED'
+        assert client.execute_command('EXEC') == [1, 1, 1, 1, 0]
+        self.verify_bloom_filter_existence(client, 'M2', 'V2')
+        self.verify_bloom_filter_existence(client, 'M1', 'V1', should_exist=False)
+        self.verify_key_number(client, 1)
+
+    def test_bloom_lua(self):
+        client = self.server.get_new_client()
+        # lua
+        load_filter = """
+        redis.call('BF.ADD', 'LUA1', 'ITEM1');
+        redis.call('BF.ADD', 'LUA2', 'ITEM2');
+        redis.call('BF.MADD', 'LUA2', 'ITEM3', 'ITEM4', 'ITEM5');
+        """
+        client.eval(load_filter, 0)
+        assert client.execute_command('BF.MEXISTS LUA2 ITEM1 ITEM3 ITEM4') == [0, 1, 1]
+        self.verify_key_number(client, 2)
+
+    def test_bloom_deletes(self):
+        client = self.server.get_new_client()
+        # delete
+        assert client.execute_command('BF.ADD filter1 item1') == 1
+        self.verify_bloom_filter_existence(client, 'filter1', 'item1')
+        self.verify_key_number(client, 1)
+        assert client.execute_command('DEL filter1') == 1
+        self.verify_bloom_filter_existence(client, 'filter1', 'item1', should_exist=False)
+        self.verify_key_number(client, 0)
+
+        # flush
+        self.insert_bloom_filter(client, number_of_bf=10)
+        self.verify_key_number(client, 10)
+        assert client.execute_command('FLUSHALL')
+        self.verify_key_number(client, 0)
+
+        # unlink
+        assert client.execute_command('BF.ADD A ITEMA') == 1
+        assert client.execute_command('BF.ADD B ITEMB') == 1
+        self.verify_bloom_filter_existence(client, 'A', 'ITEMA')
+        self.verify_bloom_filter_existence(client, 'B', 'ITEMB')
+        self.verify_bloom_filter_existence(client, 'C', 'ITEMC', should_exist=False)
+        self.verify_key_number(client, 2)
+        assert client.execute_command('UNLINK A B C') == 2
+        assert client.execute_command('BF.MEXISTS A ITEMA ITEMB') == [0, 0]
+        self.verify_bloom_filter_existence(client, 'A', 'ITEMA', should_exist=False)
+        self.verify_bloom_filter_existence(client, 'B', 'ITEMB', should_exist=False)
+        self.verify_key_number(client, 0)
+
+    def test_bloom_expiration(self):
+        client = self.server.get_new_client()
+        # expiration
+        # cmd object idletime
+        self.verify_key_number(client, 0)
+        assert client.execute_command('BF.ADD TEST_IDLE val3') == 1
+        self.verify_bloom_filter_existence(client, 'TEST_IDLE', 'val3')
+        self.verify_key_number(client, 1)
+        time.sleep(1)
+        assert client.execute_command('OBJECT IDLETIME test_idle') == None
+        assert client.execute_command('OBJECT IDLETIME TEST_IDLE') > 0
+        # cmd ttl, expireat
+        assert client.execute_command('BF.ADD TEST_EXP ITEM') == 1
+        assert client.execute_command('TTL TEST_EXP') == -1
+        self.verify_bloom_filter_existence(client, 'TEST_EXP', 'ITEM')
+        self.verify_key_number(client, 2)
+        curr_time = int(time.time())
+        assert client.execute_command(f'EXPIREAT TEST_EXP {curr_time + 5}') == 1
+        wait_for_equal(lambda: client.execute_command('BF.EXISTS TEST_EXP ITEM'), 0)
+        self.verify_key_number(client, 1)
+        # cmd persist
+        assert client.execute_command('BF.ADD TEST_PERSIST ITEM') == 1
+        assert client.execute_command('TTL TEST_PERSIST') == -1
+        self.verify_bloom_filter_existence(client, 'TEST_PERSIST', 'ITEM')
+        self.verify_key_number(client, 2)
+        assert client.execute_command(f'EXPIREAT TEST_PERSIST {curr_time + 100000}') == 1
+        assert client.execute_command('TTL TEST_PERSIST') > 0
+        assert client.execute_command('PERSIST TEST_PERSIST') == 1
+        assert client.execute_command('TTL TEST_PERSIST') == -1
--- a/tests/test_bloom_command.py
+++ b/tests/test_bloom_command.py
@@ -1,0 +1,148 @@
+import pytest
+from valkey_bloom_test_case import ValkeyBloomTestCaseBase
+from valkeytests.conftest import resource_port_tracker
+
+class TestBloomCommand(ValkeyBloomTestCaseBase):
+
+    def verify_command_arity(self, command, expected_arity): 
+        command_info = self.client.execute_command('COMMAND', 'INFO', command)
+        actual_arity = command_info.get(command).get('arity')
+        assert actual_arity == expected_arity, f"Arity mismatch for command '{command}'"
+
+    def test_bloom_command_arity(self):
+        self.verify_command_arity('BF.EXISTS', -1)
+        self.verify_command_arity('BF.ADD', -1)
+        self.verify_command_arity('BF.MEXISTS', -1)
+        self.verify_command_arity('BF.MADD', -1)
+        self.verify_command_arity('BF.CARD', -1)
+        self.verify_command_arity('BF.RESERVE', -1)
+        self.verify_command_arity('BF.INFO', -1)
+        self.verify_command_arity('BF.INSERT', -1)
+
+    def test_bloom_command_error(self):
+        # test set up
+        assert self.client.execute_command('BF.ADD key item') == 1
+        assert self.client.execute_command('BF.RESERVE bf 0.01 1000') == b'OK'
+        # non scaling filter
+        assert self.client.execute_command('BF.RESERVE bf_non 0.01 2 NONSCALING') == b'OK'
+        assert self.client.execute_command('BF.ADD bf_non 0') == 1
+        assert self.client.execute_command('BF.ADD bf_non 1') == 1
+
+        basic_error_test_cases = [
+            # not found
+            ('BF.INFO TEST404', 'not found'),
+            # incorrect syntax and argument usage
+            ('BF.ADD bf_non 2', 'non scaling filter is full'),
+            ('bf.info key item', 'invalid information value'),
+            ('bf.insert key CAPACITY 10000 ERROR 0.01 EXPANSION 0.99 NOCREATE NONSCALING ITEMS test1 test2 test3', 'bad expansion'),
+            ('BF.INSERT KEY HELLO WORLD', 'unknown argument received'),
+            ('BF.INSERT KEY error 2 ITEMS test1', '(0 < error rate range < 1)'),
+            ('BF.INSERT TEST_LIMIT ERROR 0.999999999 ITEMS ERROR_RATE', '(0 < error rate range < 1)'),
+            ('BF.INSERT TEST_LIMIT CAPACITY 4394967295 ITEMS CAP', 'bad capacity'),
+            ('BF.INSERT TEST_LIMIT CAPACITY 0 ITEMS CAP0', '(capacity should be larger than 0)'),
+            ('BF.INSERT TEST_LIMIT EXPANSION 11 ITEMS EXPAN', 'bad expansion'),
+            ('BF.INSERT TEST_NOCREATE NOCREATE ITEMS A B', 'not found'),
+            ('BF.RESERVE KEY String 100', 'bad error rate'),
+            ('BF.RESERVE KEY 0.999999999 3000', '(0 < error rate range < 1)'),
+            ('BF.RESERVE KEY 2 100', '(0 < error rate range < 1)'),
+            ('BF.RESERVE KEY 0.01 String', 'bad capacity'),
+            ('BF.RESERVE KEY 0.01 0.01', 'bad capacity'),
+            ('BF.RESERVE KEY 0.01 -1', 'bad capacity'),
+            ('BF.RESERVE KEY 0.01 4394967295', 'bad capacity'),
+            ('BF.RESERVE bf 0.01 1000', 'item exists'),
+            ('BF.RESERVE TEST_CAP 0.50 0', '(capacity should be larger than 0)'),
+
+            # wrong number of arguments
+            ('BF.ADD TEST', 'wrong number of arguments for \'BF.ADD\' command'),
+            ('BF.ADD', 'wrong number of arguments for \'BF.ADD\' command'),
+            ('BF.ADD HELLO TEST WORLD', 'wrong number of arguments for \'BF.ADD\' command'),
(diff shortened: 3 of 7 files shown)


src/bloom/utils.rs
```diff
@@ -10,8 +10,8 @@ pub const ERROR: &str = "ERROR";
 pub const NON_SCALING_FILTER_FULL: &str = "ERR non scaling filter is full";
 pub const NOT_FOUND: &str = "ERR not found";
 pub const ITEM_EXISTS: &str = "ERR item exists";
-pub const INVALID_INFO_VALUE: &str = "ERR Invalid information value";
-pub const BAD_EXPANSION: &str = "ERR bad expansion";
+pub const INVALID_INFO_VALUE: &str = "Invalid information value";
+pub const BAD_EXPANSION: &str = "Bad expansion";
```
zackcam: Should we also capitalize the other error messages first words?

tests/test_basic.py
```diff
@@ -1,18 +1,12 @@
 import time
 import pytest
 from valkey import ResponseError
-from valkeytests.valkey_test_case import ValkeyTestCase
+from valkey_bloom_test_case import ValkeyBloomTestCaseBase
 from valkeytests.conftest import resource_port_tracker
```
zackcam, line 6: Could you test if we move this import to valkey_bloom_test_case.py if tests still work? Cause if so we wont need to add this import to every test in the future

tests/test_basic.py
```diff
@@ -31,3 +25,56 @@ def test_basic(self):
...
+        assert client.execute_command('EXEC') == [1, 1, 1, 1, 0]
+        
+        # lua
+        load_filter = """
+        redis.call('BF.ADD', 'LUA1', 'ITEM1');
+        redis.call('BF.ADD', 'LUA2', 'ITEM2');
+        redis.call('BF.MADD', 'LUA2', 'ITEMS3', 'ITEMS4', 'ITEMS5');
+        """
+        client.eval(load_filter, 0)
+        assert client.execute_command('BF.MEXISTS LUA2 ITEMS2 ITEMS3 ITEMS4') == [0, 1, 1]
```
zackcam: Maybe just for readability but the checking against ITEMS2 when ITEM2 is in was kinda easy to gloss over the S, so confused me for a bit on why it didn't exist. Would ITEM1 work instead as thats probably more readable?

src/bloom/utils.rs
YueTang-Vanessa: > Should we also capitalize the other error messages first words?

Nah, this is by design.

tests/test_basic.py
YueTang-Vanessa: > Could you test if we move this import to valkey_bloom_test_case.py if tests still work? Cause if so we wont need to add this import to every test in the future

I tried to move this line `from valkeytests.conftest import resource_port_tracker` to `valkey_bloom_test_case.py` before, running tests locally won't have any issues, but when I push that in the github, I can see the workflow fail. See [this workflow](https://github.com/YueTang-Vanessa/valkey-bloom/actions/runs/10951218250) I think we can probably work on that later

tests/test_basic.py
YueTang-Vanessa: > Maybe just for readability but the checking against ITEMS2 when ITEM2 is in was kinda easy to gloss over the S, so confused me for a bit on why it didn't exist. Would ITEM1 work instead as thats probably more readable?

Sounds good. Will do it

tests/valkey_bloom_test_case.py
```diff
@@ -0,0 +1,28 @@
...
+            assert_error_msg = f"Actual error message: '{str(e)}' is different from expected error message '{expected_err_reply}'"
+            assert str(e) == expected_err_reply, assert_error_msg
+
+    def verify_command_behavior_as_expected(self, client, cmd, expected_result):
+        try:
+            cmd_actual_result = client.execute_command(cmd)
+            assert_error_msg = f"Actual command response '{cmd_actual_result}' is different from expected response '{expected_result}'"
+            assert cmd_actual_result == expected_result, assert_error_msg
+        except:
+            print("Something went wrong in command behavior verification")
```
KarthikSubbarao: Can we add a new line here?

tests/test_bloom_command.py
```diff
@@ -0,0 +1,136 @@
...
+            ('BF.CARD TEST', 5),
+            ('bf.card HELLO', 0),
+            ('BF.RESERVE bf 0.01 1000', b'OK'),
+            ('BF.RESERVE bf_exp 0.01 1000 EXPANSION 2', b'OK'),
+            ('BF.RESERVE bf_non 0.01 1000 NONSCALING', b'OK'),
+            ('bf.info bf_exp expansion', 2),
+            ('BF.INFO bf_non expansion', None),
+        ]
+
+        assert self.client.execute_command('FLUSHALL')
```
KarthikSubbarao: Do we need to execute flushall at the beginning of the test? Doesn't a new server startup at this point?

tests/test_bloom_command.py
```diff
@@ -0,0 +1,136 @@
...
+            ('BF.MEXISTS hello world5', [0]),
+            ('BF.MADD hello world5', [1]),
+            ('BF.MEXISTS hello world5 world6 world7', [1, 0, 0]),
+            ('BF.INSERT TEST ITEMS ITEMS', [1]),
+            ('BF.INSERT TEST CAPACITY 1000 ITEMS ITEMS', [0]),
+            ('BF.INSERT TEST CAPACITY 200 error 0.99 ITEMS ITEMS ITEMS1 ITEMS2', [0, 1, 1]),
+            ('BF.INSERT TEST CAPACITY 300 ERROR 0.90 EXPANSION 1 ITEMS ITEMS FOO', [0, 1]),
+            ('BF.INSERT TEST ERROR 0.80 EXPANSION 3 NOCREATE items BOO', [1]), 
+            ('BF.INSERT TEST ERROR 0.80 EXPANSION 1 NOCREATE NONSCALING items BOO', [0]),
+            ('BF.INFO TEST Capacity', 100),
```
KarthikSubbarao, line 116: Can we test other variations? `BF.INFO TEST filters` and `BF.INFO TEST`

tests/test_bloom_command.py
```diff
@@ -0,0 +1,136 @@
...
+            ('BF.EXISTS key item2', 1),
+            ('BF.EXISTS key item3', 0),
+            ('BF.MADD hello world1 world2 world3', [1, 1, 1]),
+            ('BF.MADD hello world1 world2 world3 world4', [0, 0, 0, 1]),
+            ('BF.MEXISTS hello world5', [0]),
+            ('BF.MADD hello world5', [1]),
+            ('BF.MEXISTS hello world5 world6 world7', [1, 0, 0]),
+            ('BF.INSERT TEST ITEMS ITEMS', [1]),
+            ('BF.INSERT TEST CAPACITY 1000 ITEMS ITEMS', [0]),
+            ('BF.INSERT TEST CAPACITY 200 error 0.99 ITEMS ITEMS ITEMS1 ITEMS2', [0, 1, 1]),
```
KarthikSubbarao: Nit: Creating a filter with 0.99 error rate will likely make this test flaky in terms of false positives (it can indicate that these items exist even though we did not add them). If you are adding a test where you are asserting that an item add succeeds (returning 1), we should probably use a smaller error rate

tests/test_bloom_command.py
```diff
@@ -0,0 +1,136 @@
...
+
+        assert self.client.execute_command('FLUSHALL')
+        assert self.client.info_obj().num_keys() == 0
+        assert self.client.execute_command('BF.ADD key item') == 1
+        assert self.client.execute_command('BF.RESERVE bf 0.01 1000') == b'OK'
+
+        # non scaling filter
+        assert self.client.execute_command('BF.RESERVE bf_non 0.01 2 NONSCALING') == b'OK'
+        assert self.client.execute_command('BF.ADD bf_non 0') == 1
+        assert self.client.execute_command('BF.ADD bf_non 1') == 1
```
KarthikSubbarao: Let's move this block of code to the start of the test so that we know it is a "setup" for this particular test

tests/test_basic.py
KarthikSubbarao: Would be good if this is in a separate test for transactions and lua

tests/test_basic.py
```diff
@@ -31,3 +25,56 @@ def test_basic(self):
...
+        curr_time = int(time.time())
+        assert client.execute_command(f'EXPIREAT TEST_EXP {curr_time - 1}') == 1
+        assert client.execute_command('BF.EXISTS TEST_EXP ITEM') == 0
+        assert client.execute_command('BF.ADD TEST_PERSIST ITEM') == 1
+        assert client.execute_command('TTL TEST_PERSIST') == -1
+        assert client.execute_command(f'EXPIREAT TEST_PERSIST {curr_time + 100000}') == 1
+        assert client.execute_command('BF.EXISTS TEST_PERSIST ITEM') == 1
+        assert client.execute_command('TTL TEST_PERSIST') > 0
+        assert client.execute_command('PERSIST TEST_PERSIST') == 1
+        assert client.execute_command('TTL TEST_PERSIST') == -1
```
KarthikSubbarao: We can move this to a expiration test

tests/test_basic.py
```diff
@@ -31,3 +25,56 @@ def test_basic(self):
...
+        assert client.execute_command('BF.EXISTS filter1 item1') == 0
+        assert client.execute_command('FLUSHALL')
+        assert client.info_obj().num_keys() == 0
+        assert client.execute_command('BF.ADD A ITEMA') == 1
+        assert client.execute_command('BF.ADD B ITEMB') == 1
+        assert client.execute_command('UNLINK A B C') == 2
+        assert client.execute_command('BF.MEXISTS A ITEMA ITEMB') == [0, 0]
+        assert client.execute_command('BF.ADD key1 val1') == 1
+        assert client.execute_command('BF.ADD key2 val2') == 1
+        assert client.execute_command('TOUCH key1 key2') == 2
```
KarthikSubbarao: Let's have one test for all delete operations called test_bloom_deletes. DEL, UNLINK, FLUSHALL

tests/test_bloom_command.py
```diff
@@ -0,0 +1,136 @@
...
+        self.verify_command_arity('BF.INSERT', -1)
+
+    def test_bloom_command_error(self):
+        basic_error_test_cases = [
+            # not found
+            ('BF.INFO TEST404', 'not found'),
+            # incorrect syntax and argument usage
+            ('BF.ADDX TEST KEY', 'unknown command \'BF.ADDX\', with args beginning with: \'TEST\' \'KEY\' '),
+            ('BF.EXIST KEY', 'unknown command \'BF.EXIST\', with args beginning with: \'KEY\' '),
+            ('bf.reserv', 'unknown command \'bf.reserv\', with args beginning with: '),
```
KarthikSubbarao: We can remove this case. Invalid command names are not really a bloom specific behavior (the core engine will reject this)

tests/test_basic.py
```diff
@@ -31,3 +25,56 @@ def test_basic(self):
...
         bf_exists_result = client.execute_command('BF.EXISTS filter1 item2')
         assert bf_exists_result == 0
+
+    def test_bloom_with_basic_valkey_command(self):
+        client = self.server.get_new_client()
+        # check bloom filter with basic valkey command
+        # modify and delete
+        assert client.execute_command('BF.ADD filter1 item1') == 1
+        assert client.execute_command('DEL filter1') == 1
+        assert client.execute_command('BF.EXISTS filter1 item1') == 0
```
KarthikSubbarao: 1. Can we also test the `EXISTS` command here? We can use EXISTS before deleting and it should succeed. After deleting, the command should return 0 / indicate that it does not exist.

2. Similarly, we can check num_keys before and after the delete operation. 

If we can do this for all operations including: DEL, UNLINK, FLUSHALL, and for expiration that will help validate correct creation and deletion

tests/test_basic.py
```diff
@@ -31,3 +25,56 @@ def test_basic(self):
...
+        assert client.execute_command('BF.ADD B ITEMB') == 1
+        assert client.execute_command('UNLINK A B C') == 2
+        assert client.execute_command('BF.MEXISTS A ITEMA ITEMB') == [0, 0]
+        assert client.execute_command('BF.ADD key1 val1') == 1
+        assert client.execute_command('BF.ADD key2 val2') == 1
+        assert client.execute_command('TOUCH key1 key2') == 2
+
+        # expiration
+        assert client.execute_command('BF.ADD key3 val3') == 1
+        time.sleep(1)
```
KarthikSubbarao, line 103: Instead of using sleep, if we use wait_for, it will help reduce flakiness

tests/test_basic.py
KarthikSubbarao: Nit: ITEMS2 confused me with ITEM2 :D 

Can we use either ITEM or ITEMS for the entire test? If we are using ITEM, we can just use ITEM1, ITEM2 ...  instead of adding ITEMS2

KarthikSubbarao (changes requested):

tests/valkey_bloom_test_case.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def verify_error_response(self, client, cmd, expected_err_reply):
+        try:
+            client.execute_command(cmd)
+            assert False
+        except ResponseError as e:
+            assert_error_msg = f"Actual error message: '{str(e)}' is different from expected error message '{expected_err_reply}'"
+            assert str(e) == expected_err_reply, assert_error_msg
+
+    def verify_command_behavior_as_expected(self, client, cmd, expected_result):
```
KarthikSubbarao: Nit: A shorter name would be `verify_command_success_reply` or `verify_command_success_response`

tests/test_bloom_command.py
```diff
@@ -0,0 +1,136 @@
...
+            ('bf.info key item', 'Invalid information value'),
+            ('bf.insert key CAPACITY 10000 ERROR 0.01 EXPANSION 0.99 NOCREATE NONSCALING ITEMS test1 test2 test3', 'Bad expansion'),
+            ('BF.INSERT KEY HELLO WORLD', 'wrong number of arguments for \'BF.INSERT\' command'),
+            ('BF.RESERVE KEY SSS SSS', 'bad error rate'),
```
KarthikSubbarao: [For BF.RESERVE and BF.INSERT] Could we add limit tests for the capacity, expansion, and false positive rate? For example, expansion is only allowed within a certain limit (1 to 10). Capacity (1 to u32::MAX). 

false positive rate has a precision of f32. Example:  Parsing `0.99999999` as f32 becomes 1

tests/test_basic.py
YueTang-Vanessa: > Instead of using sleep, if we use wait_for, it will help reduce flakiness

As discussed offline, will keep the sleep here for the OBJECT IDLETIME test, will use wait_for from util for the EXPIREAT command

src/bloom/utils.rs
YueTang-Vanessa: As discussed offline with Karthik, will add bad the ERR prefix to keep our message consistent.

tests/test_redis_bloom_compatibility.py
```diff
@@ -0,0 +1,137 @@
+import pytest
+import valkey
+from valkey.exceptions import ValkeyError
+from valkey_bloom_test_case import ValkeyBloomTestCaseBase
+from valkeytests.conftest import resource_port_tracker
+
+class TestRedisBloomCompatibility(ValkeyBloomTestCaseBase):
```
KarthikSubbarao: We can update to: `TestValkeyBloomCompatibility`

tests/test_redis_bloom_compatibility.py
```diff
@@ -0,0 +1,137 @@
...
+    #     assert decoded_r.bf().create("pipeline", 0.01, 1000)
+    #     for i in range(100):
+    #         pipeline.add("pipeline", i)
+    #     for i in range(100):
+    #         assert not (decoded_r.bf().exists("pipeline", i))
+    #
+    #     pipeline.execute()
+    #
+    #     for i in range(100):
+    #         assert decoded_r.bf().exists("pipeline", i)
```
KarthikSubbarao: Why is this commented? Was this from the original valkey-py test file for bloom?

tests/test_bloom_command.py
```diff
@@ -0,0 +1,138 @@
...
+        # non scaling filter
+        assert self.client.execute_command('BF.RESERVE bf_non 0.01 2 NONSCALING') == b'OK'
+        assert self.client.execute_command('BF.ADD bf_non 0') == 1
+        assert self.client.execute_command('BF.ADD bf_non 1') == 1
+
+        basic_error_test_cases = [
+            # not found
+            ('BF.INFO TEST404', 'not found'),
+            # incorrect syntax and argument usage
+            ('BF.ADDX TEST KEY', 'unknown command \'BF.ADDX\', with args beginning with: \'TEST\' \'KEY\' '),
```
KarthikSubbarao: We can remove this test case since the `BF.ADDX` command does not exist

tests/test_bloom_command.py
```diff
@@ -0,0 +1,138 @@
...
+            ('BF.ADD bf_non 2', 'non scaling filter is full'),
+            ('BF.EXIST KEY', 'unknown command \'BF.EXIST\', with args beginning with: \'KEY\' '),
```
KarthikSubbarao: We can remove this test case since the `BF.EXIST` command does not exist

tests/test_bloom_command.py
```diff
@@ -0,0 +1,138 @@
...
+            ('BF.INSERT TEST_LIMIT ERROR 0.999999999 ITEMS ERROR_RATE', '(0 < error rate range < 1)'),
+            ('BF.INSERT TEST_LIMIT CAPACITY 4394967295 ITEMS CAP', 'bad capacity'),
+            ('BF.INSERT TEST_LIMIT EXPANSION 11 ITEMS EXPAN', 'bad expansion'),
+            ('BF.RESERVE KEY String 100', 'bad error rate'),
+            ('BF.RESERVE KEY 0.01 String', 'bad capacity'),
+            ('BF.RESERVE KEY 0.01 0.01', 'bad capacity'),
+            ('BF.RESERVE KEY 0.01 -1', 'bad capacity'),
+            ('BF.RESERVE TT 0.01 1 NONSCALING EXPANSION 1', 'wrong number of arguments for \'BF.RESERVE\' command'),
+            ('BF.RESERVE bf 0.01 1000', 'item exists'),
+            ('BF.RESERVE TEST_CAP 0.50 0', '(capacity should be larger than 0)'),
```
KarthikSubbarao, line 53: Can we test the same scenario for `BF.INSERT`?

tests/test_bloom_command.py
```diff
@@ -0,0 +1,138 @@
...
+            ('BF.INSERT KEY error 2 ITEMS test1', '(0 < error rate range < 1)'),
+            ('BF.INSERT TEST_LIMIT ERROR 0.999999999 ITEMS ERROR_RATE', '(0 < error rate range < 1)'),
+            ('BF.INSERT TEST_LIMIT CAPACITY 4394967295 ITEMS CAP', 'bad capacity'),
+            ('BF.INSERT TEST_LIMIT EXPANSION 11 ITEMS EXPAN', 'bad expansion'),
+            ('BF.RESERVE KEY String 100', 'bad error rate'),
+            ('BF.RESERVE KEY 0.01 String', 'bad capacity'),
+            ('BF.RESERVE KEY 0.01 0.01', 'bad capacity'),
+            ('BF.RESERVE KEY 0.01 -1', 'bad capacity'),
+            ('BF.RESERVE TT 0.01 1 NONSCALING EXPANSION 1', 'wrong number of arguments for \'BF.RESERVE\' command'),
+            ('BF.RESERVE bf 0.01 1000', 'item exists'),
```
KarthikSubbarao, line 52: Can we test a NOCREATE scenario for BF.INSERT?

tests/test_bloom_command.py
```diff
@@ -0,0 +1,138 @@
...
+            # not found
+            ('BF.INFO TEST404', 'not found'),
+            # incorrect syntax and argument usage
+            ('BF.ADDX TEST KEY', 'unknown command \'BF.ADDX\', with args beginning with: \'TEST\' \'KEY\' '),
+            ('BF.ADD bf_non 2', 'non scaling filter is full'),
+            ('BF.EXIST KEY', 'unknown command \'BF.EXIST\', with args beginning with: \'KEY\' '),
+            ('bf.info key item', 'invalid information value'),
+            ('bf.insert key CAPACITY 10000 ERROR 0.01 EXPANSION 0.99 NOCREATE NONSCALING ITEMS test1 test2 test3', 'bad expansion'),
+            ('BF.INSERT KEY HELLO WORLD', 'wrong number of arguments for \'BF.INSERT\' command'),
+            ('BF.INSERT KEY error 2 ITEMS test1', '(0 < error rate range < 1)'),
```
KarthikSubbarao, line 39: Can we test this same scenario (error rate outside range) for BF.RESERVE?

tests/test_redis_bloom_compatibility.py
```diff
@@ -0,0 +1,137 @@
+import pytest
```
KarthikSubbarao: Let's rename the file to `test_valkeypy_bloom_compatibility.py` or `test_valkeypy_bloom_compat.py`

tests/test_bloom_command.py
```diff
@@ -0,0 +1,138 @@
...
+            ('BF.MADD hello world5', [1]),
+            ('BF.MEXISTS hello world5 world6 world7', [1, 0, 0]),
+            ('BF.INSERT TEST ITEMS ITEM', [1]),
+            ('BF.INSERT TEST CAPACITY 1000 ITEMS ITEM', [0]),
+            ('BF.INSERT TEST CAPACITY 200 error 0.50 ITEMS ITEM ITEM1 ITEM2', [0, 1, 1]),
+            ('BF.INSERT TEST CAPACITY 300 ERROR 0.50 EXPANSION 1 ITEMS ITEM FOO', [0, 1]),
+            ('BF.INSERT TEST ERROR 0.50 EXPANSION 3 NOCREATE items BOO', [1]), 
+            ('BF.INSERT TEST ERROR 0.50 EXPANSION 1 NOCREATE NONSCALING items BOO', [0]),
+            ('BF.INSERT TEST_EXPANSION EXPANSION 9 ITEMS ITEM', [1]),
+            ('BF.INSERT TEST_CAPACITY CAPACITY 2000 ITEMS ITEM', [1]),
```
KarthikSubbarao, line 114: Can we also add this test case?
```
BF.INSERT <key> ITEMS 1 2 3 EXPANSION 2
```
Here, `1 2 3 EXPANSION 2` are all treated as items. So we should have an array reply with 5 integers

tests/test_bloom_command.py
```diff
@@ -0,0 +1,138 @@
...
+        basic_error_test_cases = [
+            # not found
+            ('BF.INFO TEST404', 'not found'),
+            # incorrect syntax and argument usage
+            ('BF.ADDX TEST KEY', 'unknown command \'BF.ADDX\', with args beginning with: \'TEST\' \'KEY\' '),
+            ('BF.ADD bf_non 2', 'non scaling filter is full'),
+            ('BF.EXIST KEY', 'unknown command \'BF.EXIST\', with args beginning with: \'KEY\' '),
+            ('bf.info key item', 'invalid information value'),
+            ('bf.insert key CAPACITY 10000 ERROR 0.01 EXPANSION 0.99 NOCREATE NONSCALING ITEMS test1 test2 test3', 'bad expansion'),
+            ('BF.INSERT KEY HELLO WORLD', 'wrong number of arguments for \'BF.INSERT\' command'),
```
KarthikSubbarao: This should return a unknown arg error and looks like it will need a small code change in the BF.INSERT handler

tests/test_bloom_command.py
```diff
@@ -0,0 +1,138 @@
...
+
+        basic_error_test_cases = [
+            # not found
+            ('BF.INFO TEST404', 'not found'),
+            # incorrect syntax and argument usage
+            ('BF.ADDX TEST KEY', 'unknown command \'BF.ADDX\', with args beginning with: \'TEST\' \'KEY\' '),
+            ('BF.ADD bf_non 2', 'non scaling filter is full'),
+            ('BF.EXIST KEY', 'unknown command \'BF.EXIST\', with args beginning with: \'KEY\' '),
+            ('bf.info key item', 'invalid information value'),
+            ('bf.insert key CAPACITY 10000 ERROR 0.01 EXPANSION 0.99 NOCREATE NONSCALING ITEMS test1 test2 test3', 'bad expansion'),
```
KarthikSubbarao, line 37: Can we add a test case for the command variation below?
```
BF.INSERT <key> EXPANSION 2
```

This should return a WRONG ARITY error and looks like it will need a small code change in the BF.INSERT handler

tests/test_redis_bloom_compatibility.py
YueTang-Vanessa: > Why is this commented? Was this from the original valkey-py test file for bloom?

Yes, this is from the [valkey-py/test-bloom.py](https://github.com/valkey-io/valkey-py/blob/main/tests/test_bloom.py#L492-L505). I have tried this test, it failed with message `AttributeError: 'BFBloom' object has no attribute 'pipeline'`, I guess this is valkey-py intended to comment

PR #14, opened 2024-10-10
Updating _ to -in lib.rs. Also updating loading from rdb to use a trait instead

zackcam:
General refactoring changes to use a trait instead for RDB loads will eventually also have trait be updated for valkey save as well.


KarthikSubbarao:
Thanks, looks good.

PR #15, opened 2024-10-12
Add Integration Testing for correctness of scaling and non scaling filters and other core functionality validation

KarthikSubbarao:
PR to add integration tests for validating the correctness of bloom filters (scaling and non scaling) and also testing other core Valkey operations on bloom module objects.

### DIFF
--- a/tests/test_correctness.py
+++ b/tests/test_correctness.py
@@ -1,105 +1,11 @@
-import time
 import pytest
-from valkey import ResponseError
-from valkeytests.valkey_test_case import ValkeyTestCase
-from valkeytests.conftest import resource_port_tracker
-import logging
-import os
-import random
-import string
+from valkey_bloom_test_case import ValkeyBloomTestCaseBase
 
-def generate_random_string(length=7):
-    characters = string.ascii_letters + string.digits
-    random_string = ''.join(random.choice(characters) for _ in range(length))
-    return random_string
-
-def add_items_till_capacity(client, filter_name, capacity_needed, starting_item_idx, rand_prefix, batch_size=1000):
-    new_item_idx = starting_item_idx
-    fp_count = 0
-    cardinality = client.execute_command(f'BF.CARD {filter_name}')
-    while cardinality < capacity_needed:
-        # Calculate how many more items we need to add. 
-        remaining_capacity = capacity_needed - cardinality
-        batch_to_add = min(batch_size, remaining_capacity)        
-        # Prepare a batch of items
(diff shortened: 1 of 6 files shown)


tests/test_correctness.py
```diff
@@ -0,0 +1,224 @@
...
+            add_operation_idx + 1,
+            add_operation_idx * 2,
+            False,
+            item_prefix,
+        )
+        fp_assert(error_count, num_operations, expected_fp_rate, fp_margin)
+
+class TestBloomCorrectness(ValkeyTestCase):
+
+    def get_custom_args(self):
```
YueTang-Vanessa: If you are using `ValkeyBloomTestCaseBase`, you don't need to write this function in every test file.

tests/test_correctness.py
```diff
@@ -0,0 +1,224 @@
...
+            client,
+            copy_filter_name,
+            add_operation_idx + 1,
+            add_operation_idx * 2,
+            False,
+            item_prefix,
+        )
+        fp_assert(error_count, num_operations, expected_fp_rate, fp_margin)
+
+class TestBloomCorrectness(ValkeyTestCase):
```
YueTang-Vanessa: Can you use `ValkeyBloomTestCaseBase` instead of `ValkeyTestCase`?

tests/test_correctness.py
```diff
@@ -0,0 +1,224 @@
...
+        for item_result in result:
+            if item_result != expected_result:
+                error_count += 1
+    return error_count, num_operations
+
+def fp_assert(error_count, num_operations, expected_fp_rate, fp_margin):
+    real_fp_rate = error_count / num_operations
+    fp_rate_with_margin = expected_fp_rate + fp_margin
+    
+    assert real_fp_rate < fp_rate_with_margin, f"The actual fp_rate, {real_fp_rate}, is greater than the configured fp_rate with margin. {fp_rate_with_margin}."
```
YueTang-Vanessa: Nitpick: maybe move these methods to [valkey_bloom_test_case.py](https://github.com/valkey-io/valkey-bloom/blob/unstable/tests/valkey_bloom_test_case.py) so that we can reuse this method for future bloom integration tests.

tests/test_correctness.py
KarthikSubbarao: Nice! I forgot you added that wrapper class. I will use it

tests/test_correctness.py
KarthikSubbarao: Sure. I will move it there

YueTang-Vanessa (approved):

Issue #8, opened 2024-10-07
Support for BF.LOAD command to handle AOF Rewrite


KarthikSubbarao:
Modules can support AOF Rewrite through a Module data type callback. valkey-bloom implements a module data type callback and during AOF Rewrite, we would want to write a "BF.LOAD" command that contains the dump of the existing bloom object with all the properties and bit vector dump so that it can be recreated when the command is used later on from the AOF File.

Syntax: `BF.LOAD <key-name> <dump>`

This issue will implement the data type callback here (https://github.com/valkey-io/valkey-bloom/blob/unstable/src/bloom/data_type.rs#L22) and add support for a new BF.LOAD command

wuranxx:
@KarthikSubbarao I'd like to contribute this feature. Is this already under development?

Is BF.LOAD the equivalent of redisbloom's BF.LOADCHUNK? Do we need to implement BF.LOAD and BF.DUMP (BF.DUMP replaces BF.SCANDUMP) at the same time? Do the two commands need to be consistent with the command names of redisbloom to ensure API compatibility?

KarthikSubbarao:
@wuranxx - Thank you. This feature is not yet under development, and you can contribute it. 

BF.LOAD is intended to provide similar functionality (to restore a bloom object), but it is different from BF.LOADCHUNK. 
* We only plan to use this during AOF Rewrite. You can take a look at the "AOF Rewrite handling" section in the ValkeyBloom RFC [here](https://github.com/valkey-io/valkey-rfc/pull/4/files#diff-734f99a027bf48242220e1e0cf89cce8913501d524513a69ef945c26ed586859)
* It does not load in iterations. It loads / restores a bloom object in one go.

For save (equivalent to BF.SCANDUMP), valkey-bloom provides RDB Save functionality which can be used. Alternatively, when using AOF, we will write the BF.LOAD command as mentioned above

PR #17, opened 2024-10-19
add bf.load to support bloom filter aofrewrite.

wuranxx:
this pr is for #8  https://github.com/valkey-io/valkey-bloom/issues/8.

# design proposal

I use serde  and bincode to  deserialize and deserialize BloomFilterType. add data version in first element.

bf.load not support override old key.

### DIFF
--- a/src/bloom/command_handler.rs
+++ b/src/bloom/command_handler.rs
@@ -490,3 +490,79 @@ pub fn bloom_filter_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
         _ => Err(ValkeyError::Str(utils::NOT_FOUND)),
     }
 }
+
+pub fn bloom_filter_load(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
+    let argc = input_args.len();
+    if argc != 3 {
+        return Err(ValkeyError::WrongArity);
+    }
+    let mut idx = 1;
+    let filter_name = &input_args[idx];
+    idx += 1;
+    let value = &input_args[idx];
+    // find filter
+    let filter_key = ctx.open_key_writable(filter_name);
+
+    let filter = match filter_key.get_value::<BloomFilterType>(&BLOOM_FILTER_TYPE) {
+        Ok(v) => v,
+        Err(_) => {
+            // error
+            return Err(ValkeyError::Str(utils::ERROR));
+        }
+    };
+    match filter {
+        Some(_) => {
+            // if bloom exists, return exists error.
+            Err(ValkeyError::Str(utils::KEY_EXISTS))
+        }
+        None => {
+            // if filter not exists, create it.
+            let hex = value.to_vec();
+            let bf = match BloomFilterType::decoder_bloom_filter(&hex) {
+                Ok(v) => v,
+                Err(_) => {
+                    return Err(ValkeyError::Str(utils::ERROR));
+                }
+            };
+            match filter_key.set_value(&BLOOM_FILTER_TYPE, bf) {
+                Ok(_) => {
+                    replicate_and_notify_events(ctx, filter_name, false, true);
+                    VALKEY_OK
+                }
+                Err(_) => Err(ValkeyError::Str(utils::ERROR)),
+            }
+        }
+    }
+}
+
+pub fn bloom_filter_dump(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
+    let argc = input_args.len();
+    if argc != 2 {
+        return Err(ValkeyError::WrongArity);
+    }
+    let idx = 1;
+    let filter_name = &input_args[idx];
+    // find filter
+    let filter_key = ctx.open_key_writable(filter_name);
+
+    let value = match filter_key.get_value::<BloomFilterType>(&BLOOM_FILTER_TYPE) {
+        Ok(v) => v,
+        Err(_) => {
+            // not found return error
+            return Err(ValkeyError::Str(utils::INVALID_ARGUMENT));
+        }
+    };
+    match value {
+        Some(bf) => {
+            let hex = match bf.encoder_bloom_filter() {
+                Ok(val) => val,
+                Err(err) => {
+                    println!("encode bloom filter failed. {err}");
+                    return Err(ValkeyError::Str(utils::ERROR));
+                }
+            };
+            Ok(ValkeyValue::StringBuffer(hex))
+        }
+        None => Ok(ValkeyValue::Null),
+    }
+}
--- a/src/bloom/command_handler.rs
+++ b/src/bloom/command_handler.rs
@@ -518,10 +518,11 @@ pub fn bloom_filter_load(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
         None => {
             // if filter not exists, create it.
             let hex = value.to_vec();
-            let bf = match BloomFilterType::decode_bloom_filter(&hex) {
+            let validate_size_limit = !ctx.get_flags().contains(ContextFlags::REPLICATED);
+            let bf = match BloomFilterType::decode_bloom_filter(&hex, validate_size_limit) {
                 Ok(v) => v,
-                Err(_) => {
-                    return Err(ValkeyError::Str(utils::ERROR));
+                Err(err) => {
+                    return Err(ValkeyError::Str(err.as_str()));
                 }
             };
             match filter_key.set_value(&BLOOM_FILTER_TYPE, bf) {
--- a/src/bloom/command_handler.rs
+++ b/src/bloom/command_handler.rs
@@ -518,7 +518,7 @@ pub fn bloom_filter_load(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
         None => {
             // if filter not exists, create it.
             let hex = value.to_vec();
-            let bf = match BloomFilterType::decoder_bloom_filter(&hex) {
+            let bf = match BloomFilterType::decode_bloom_filter(&hex) {
                 Ok(v) => v,
                 Err(_) => {
                     return Err(ValkeyError::Str(utils::ERROR));
@@ -532,37 +532,5 @@ pub fn bloom_filter_load(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
                 Err(_) => Err(ValkeyError::Str(utils::ERROR)),
             }
         }
-    }
-}
-
-pub fn bloom_filter_dump(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
-    let argc = input_args.len();
-    if argc != 2 {
-        return Err(ValkeyError::WrongArity);
-    }
-    let idx = 1;
-    let filter_name = &input_args[idx];
-    // find filter
-    let filter_key = ctx.open_key_writable(filter_name);
-
-    let value = match filter_key.get_value::<BloomFilterType>(&BLOOM_FILTER_TYPE) {
-        Ok(v) => v,
-        Err(_) => {
-            // not found return error
-            return Err(ValkeyError::Str(utils::INVALID_ARGUMENT));
-        }
-    };
-    match value {
-        Some(bf) => {
-            let hex = match bf.encoder_bloom_filter() {
-                Ok(val) => val,
-                Err(err) => {
-                    println!("encode bloom filter failed. {err}");
-                    return Err(ValkeyError::Str(utils::ERROR));
-                }
-            };
-            Ok(ValkeyValue::StringBuffer(hex))
-        }
-        None => Ok(ValkeyValue::Null),
     }
 }
--- a/src/bloom/data_type.rs
+++ b/src/bloom/data_type.rs
@@ -12,9 +12,9 @@ use std::os::raw::c_int;
 use valkey_module::native_types::ValkeyType;
 use valkey_module::{logging, raw};
 
-use super::utils::BLOOM_TYPE_VERSION;
-
-use super::utils::BLOOM_TYPE_VERSION;
+/// Use for decode an encode `BloomFilterType`.
+/// This value must increased when `BloomFilterType` struct change.
+pub const BLOOM_TYPE_VERSION: u8 = 1;
 
 const BLOOM_FILTER_TYPE_ENCODING_VERSION: i32 = 1;
 
--- a/src/bloom/data_type.rs
+++ b/src/bloom/data_type.rs
@@ -12,7 +12,7 @@ use std::os::raw::c_int;
 use valkey_module::native_types::ValkeyType;
 use valkey_module::{logging, raw};
 
-/// Use for decode an encode `BloomFilterType`.
+/// Used for decoding and encoding `BloomFilterType`. Currently used in AOF Rewrite.
 /// This value must increased when `BloomFilterType` struct change.
 pub const BLOOM_TYPE_VERSION: u8 = 1;
 
@@ -25,7 +25,6 @@ pub static BLOOM_FILTER_TYPE: ValkeyType = ValkeyType::new(
         version: raw::REDISMODULE_TYPE_METHOD_VERSION as u64,
         rdb_load: Some(bloom_callback::bloom_rdb_load),
         rdb_save: Some(bloom_callback::bloom_rdb_save),
-        // TODO
         aof_rewrite: Some(bloom_callback::bloom_aof_rewrite),
 
         mem_usage: Some(bloom_callback::bloom_mem_usage),
--- a/src/bloom/data_type.rs
+++ b/src/bloom/data_type.rs
@@ -12,6 +12,10 @@ use std::os::raw::c_int;
 use valkey_module::native_types::ValkeyType;
 use valkey_module::{logging, raw};
 
+use super::utils::BLOOM_TYPE_VERSION;
+
+use super::utils::BLOOM_TYPE_VERSION;
+
 const BLOOM_FILTER_TYPE_ENCODING_VERSION: i32 = 1;
 
 pub static BLOOM_FILTER_TYPE: ValkeyType = ValkeyType::new(
@@ -22,7 +26,7 @@ pub static BLOOM_FILTER_TYPE: ValkeyType = ValkeyType::new(
         rdb_load: Some(bloom_callback::bloom_rdb_load),
         rdb_save: Some(bloom_callback::bloom_rdb_save),
         // TODO
-        aof_rewrite: None,
+        aof_rewrite: Some(bloom_callback::bloom_aof_rewrite),
 
         mem_usage: Some(bloom_callback::bloom_mem_usage),
         // TODO
@@ -117,6 +121,7 @@ impl ValkeyDataType for BloomFilterType {
         );
         BLOOM_NUM_OBJECTS.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
         let item = BloomFilterType {
+            version: BLOOM_TYPE_VERSION,
             expansion: expansion as u32,
             fp_rate,
             filters,
--- a/src/bloom/data_type.rs
+++ b/src/bloom/data_type.rs
@@ -121,7 +121,6 @@ impl ValkeyDataType for BloomFilterType {
         );
         BLOOM_NUM_OBJECTS.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
         let item = BloomFilterType {
-            version: BLOOM_TYPE_VERSION,
             expansion: expansion as u32,
             fp_rate,
             filters,
--- a/src/bloom/utils.rs
+++ b/src/bloom/utils.rs
@@ -809,7 +809,6 @@ mod tests {
         let _ = bf.add_item(key.as_bytes(), true);
         let origin_expansion = bf.expansion;
         let origin_fp_rate = bf.fp_rate;
-        let origin_size = configs::BLOOM_MEMORY_LIMIT_PER_FILTER.load(Ordering::SeqCst);
         // unsupoort expansion
         bf.expansion = 0;
 
@@ -837,14 +836,14 @@ mod tests {
         );
         bf.fp_rate = origin_fp_rate;
 
-        // 3. unsupport filter size:
-        configs::BLOOM_MEMORY_LIMIT_PER_FILTER.store(1, Ordering::SeqCst);
-        let vec = bf.encode_bloom_filter().unwrap();
+        // 3. build a larger than 64mb filter
+        let extra_large_filter =
+            BloomFilterType::new_reserved(0.01_f64, 57000000, 2, false).unwrap();
+        let vec = extra_large_filter.encode_bloom_filter().unwrap();
         // should return error
         assert_eq!(
             BloomFilterType::decode_bloom_filter(&vec, true).err(),
             Some(BloomError::ExceedsMaxBloomSize)
         );
-        configs::BLOOM_MEMORY_LIMIT_PER_FILTER.store(origin_size, Ordering::SeqCst);
     }
 }
--- a/src/bloom/utils.rs
+++ b/src/bloom/utils.rs
@@ -1,5 +1,6 @@
 use crate::{configs, metrics};
 use bloomfilter;
+use serde::{Deserialize, Serialize};
 use std::{mem, sync::atomic::Ordering};
 
 /// KeySpace Notification Events
@@ -21,6 +22,10 @@ pub const MAX_NUM_SCALING_FILTERS: &str = "ERR bloom object reached max number o
 pub const UNKNOWN_ARGUMENT: &str = "ERR unknown argument received";
 pub const EXCEEDS_MAX_BLOOM_SIZE: &str =
     "ERR operation results in filter allocation exceeding size limit";
+pub const INVALID_ARGUMENT: &str = "ERR invalid argument received";
+pub const KEY_EXISTS: &str = "-BUSYKEY Target key name already exists.";
+
+pub const BLOOM_TYPE_VERSION: u8 = 1;
 
 #[derive(Debug, PartialEq)]
 pub enum BloomError {
@@ -42,7 +47,9 @@ impl BloomError {
 /// The BloomFilterType structure. 40 bytes.
 /// Can contain one or more filters.
 /// This is a generic top level structure which is not coupled to any bloom crate.
+#[derive(Serialize, Deserialize)]
 pub struct BloomFilterType {
+    pub version: u8,
     pub expansion: u32,
     pub fp_rate: f64,
     pub filters: Vec<BloomFilter>,
@@ -71,6 +78,7 @@ impl BloomFilterType {
         let bloom = BloomFilter::new(fp_rate, capacity);
         let filters = vec![bloom];
         let bloom = BloomFilterType {
+            version: BLOOM_TYPE_VERSION,
             expansion,
             fp_rate,
             filters,
@@ -91,6 +99,7 @@ impl BloomFilterType {
             filters.push(new_filter);
         }
         BloomFilterType {
+            version: BLOOM_TYPE_VERSION,
             expansion: from_bf.expansion,
             fp_rate: from_bf.fp_rate,
             filters,
@@ -188,6 +197,41 @@ impl BloomFilterType {
         }
         Ok(0)
     }
+
+    pub fn encoder_bloom_filter(&self) -> Result<Vec<u8>, Box<bincode::ErrorKind>> {
+        let vec = bincode::serialize(self)?;
+        Ok(vec)
+    }
+
+    pub fn decoder_bloom_filter(
+        decoded_bytes: &[u8],
+    ) -> Result<BloomFilterType, Box<dyn std::error::Error>> {
+        if decoded_bytes.is_empty() {
+            return Err(bincode::Error::new(bincode::ErrorKind::Custom(
+                String::from("invalid data length"),
+            )));
+        }
+        let version = decoded_bytes[0];
+        match version {
+            1 => {
+                // always use new version to init bloomFilterType.
+                // This is to ensure that the new fields can be recognized when the object is serialized and deserialized in the future.
+                let (_, expansion, fp_rate, filters): (u8, u32, f32, Vec<BloomFilter>) =
+                    bincode::deserialize(decoded_bytes)?;
+                let filter = BloomFilterType {
+                    version: BLOOM_TYPE_VERSION,
+                    expansion,
+                    fp_rate,
+                    filters,
+                };
+                Ok(filter)
+            }
+
+            _ => Err(bincode::Error::new(bincode::ErrorKind::Custom(
+                String::from("not support version"),
+            ))),
+        }
+    }
 }
 
 /// Structure representing a single bloom filter. 200 Bytes.
@@ -196,6 +240,7 @@ impl BloomFilterType {
 /// we have a limit on the memory usage of a `BloomFilter` to be 64MB.
 /// Based on this, we expect the number of items on the `BloomFilter` to be
 /// well within the u32::MAX limit.
+#[derive(Serialize, Deserialize)]
 pub struct BloomFilter {
     pub bloom: bloomfilter::Bloom<[u8]>,
     pub num_items: u32,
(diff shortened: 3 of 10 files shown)


wuranxx:
@KarthikSubbarao Please help review this PR.

src/lib.rs
```diff
@@ -77,6 +89,8 @@ valkey_module! {
         ["BF.RESERVE", bloom_reserve_command, "write fast deny-oom", 1, 1, 1],
         ["BF.INFO", bloom_info_command, "readonly fast", 1, 1, 1],
         ["BF.INSERT", bloom_insert_command, "write fast deny-oom", 1, 1, 1],
+        ["BF.LOAD", bloom_load_command, "write fast deny-oom", 1, 1, 1],
+        ["BF.DUMP", bloom_dump_command, "readonly fast", 1, 1, 1]
```
KarthikSubbarao: In the [ValkeyBloom RFC](https://github.com/valkey-io/valkey-rfc/pull/4), we discussed that the BF.DUMP command was not needed because for save and restore, the user can use RDB Save and Load. And for AOF Rewrite, we only need the BF.LOAD command to be saved. This BF.LOAD command can be replayed later on to re-create the object.

So, can we remove the BF.DUMP command? Or is it needed for any specific purpose?

src/bloom/command_handler.rs
```diff
@@ -441,3 +441,98 @@ pub fn bloom_filter_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
...
+            return Err(ValkeyError::Str(utils::ERROR));
+        }
+    };
+    match filter {
+        Some(_) => {
+            // if bloom exists, replace it.
+            match filter_key.delete() {
+                Ok(v) => v,
+                Err(_) => return Err(ValkeyError::Str(utils::ERROR)),
+            };
```
KarthikSubbarao: Question: Did we decide to overwrite (delete + recreate) based on any existing examples? Is any existing "LOAD" functionality where we overwrite?

The `RESTORE` command returns an error if the key already exists.
```
127.0.0.1:6379> set key v
OK
127.0.0.1:6379> dump key
"\x00\x01v\x0b\x00\xf8\xd7s\xc1\xe4\xa7cG"
127.0.0.1:6379> restore key "\x00\x01v\x0b\x00\xf8\xd7s\xc1\xe4\xa7cG" 0
(error) BUSYKEY Target key name already exists.
```

src/bloom/command_handler.rs
```diff
@@ -441,3 +441,98 @@ pub fn bloom_filter_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
...
+            let hex = value.to_vec();
+            let new_bf = match BloomFilterType::decoder_bloom_filter(&hex) {
```
KarthikSubbarao: Looks like the main validation in current approach here is to rely on deserializing to succeed and if it fails, we return an error. 

Is there any version check logic we should add? Should we have any versioning of the `BloomFilterType` structure?

src/wrapper/bloom_callback.rs
```diff
@@ -48,6 +49,29 @@ pub unsafe extern "C" fn bloom_rdb_load(
...
+    aof: *mut raw::RedisModuleIO,
+    key: *mut raw::RedisModuleString,
+    value: *mut c_void,
+) {
+    let filter = &*value.cast::<BloomFilterType>();
+    let hex = match filter.encoder_bloom_filter() {
+        Ok(val) => val,
+        Err(err) => panic!("encode bloom filter failed. {err}"),
+    };
+    let cmd = CString::new("BF.LOAD").unwrap();
```
KarthikSubbarao, line 73: The current PR, does not include TTL information. Do we want to add this? We have Module APIs such as `ValkeyModule_GetAbsExpire` which return the unix timestamp of expiration. Using this, we can include it as an optional arg in the BF.LOAD command syntax (seems more optimal) OR if we want to keep it separate, we can write a `EXPIREAT ` command into the AOF (will result in larger AOF to include extra command).

tests/test_aofrewrite.py
```diff
@@ -0,0 +1,41 @@
...
+        time.sleep(1)
+        uptime_in_sec_1 = self.client.info_obj().uptime_in_secs()
+        self.server.restart(remove_rdb=False, remove_nodes_conf=False, connect_client=True)
+        uptime_in_sec_2 = self.client.info_obj().uptime_in_secs()
+        assert self.server.is_alive()
+        assert uptime_in_sec_1 > uptime_in_sec_2
+
+        # verify restore results
+        curr_item_count_2 = client.info_obj().num_keys()
+        assert curr_item_count_2 == curr_item_count_1
```
KarthikSubbarao, line 34: These checks are good. We can consider adding more explicit checks here to look into the command stats and `calls` count for `BF.LOAD`. For example: https://github.com/valkey-io/valkey-bloom/blob/unstable/tests/test_replication.py#L126

KarthikSubbarao (commented):
Thank you @wuranxx. I reviewed this change at a high level and added comments

src/lib.rs
wuranxx: My thought is that in both redis and redisbloom, load and dump are a pair of commands. If the dump command is not implemented, there is no method for obtaining the data parameter of the load command. As a result, the test and migration scenarios are difficult.

src/bloom/command_handler.rs
wuranxx: if we decide load command not override origin key, this is simpler  implementation. I will fix it.

wuranxx: I previously thought that RedisBloom supported the LOAD command to overwrite keys, but after testing, I found it doesn't. I will update this implementation that overwriting is not supported.
```sh
127.0.0.1:6379> bf.reserve key1 0.01 10
OK
127.0.0.1:6379> bf.reserve key2 0.01 15
OK
127.0.0.1:6379> bf.SCANDUMP key1 0
1) (integer) 1
2) "\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x05\x00\x00\x00\x02\x00\x00\x00\x10\x00\x00\x00\x00\x00\x00\x00\x80\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00{\x14\xaeG\xe1zt?\xe9\x86/\xb25\x0e&@\b\x00\x00\x00\n\x00\x00\x00\x00\x00\x00\x00\x00"
127.0.0.1:6379> bf.SCANDUMP key1 1
1) (integer) 17
2) "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
127.0.0.1:6379> bf.SCANDUMP key1 17
1) (integer) 0
2) ""
127.0.0.1:6379> BF.LOADCHUNK key2 1 "\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x05\x00\x00\x00\x02\x00\x00\x00\x10\x00\x00\x00\x00\x00\x00\x00\x80\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00{\x14\xaeG\xe1zt?\xe9\x86/\xb25\x0e&@\b\x00\x00\x00\n\x00\x00\x00\x00\x00\x00\x00\x00"
(error) ERR invalid offset - no link found
127.0.0.1:6379> BF.LOADCHUNK key1 1 "\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x05\x00\x00\x00\x02\x00\x00\x00\x10\x00\x00\x00\x00\x00\x00\x00\x80\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00{\x14\xaeG\xe1zt?\xe9\x86/\xb25\x0e&@\b\x00\x00\x00\n\x00\x00\x00\x00\x00\x00\x00\x00"
(error) ERR invalid offset - no link found
127.0.0.1:6379> BF.LOADCHUNK key3 1 "\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x05\x00\x00\x00\x02\x00\x00\x00\x10\x00\x00\x00\x00\x00\x00\x00\x80\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00{\x14\xaeG\xe1zt?\xe9\x86/\xb25\x0e&@\b\x00\x00\x00\n\x00\x00\x00\x00\x00\x00\x00\x00"
OK
127.0.0.1:6379> bf.info key3
 1) Capacity
 2) (integer) 10
 3) Size
 4) (integer) 112
 5) Number of filters
 6) (integer) 1
 7) Number of items inserted
 8) (integer) 0
 9) Expansion rate
10) (integer) 2
```

src/bloom/command_handler.rs
wuranxx: ok, i will try to add version control when serializing and deserializing.

tests/test_aofrewrite.py
wuranxx: I will add more check in this case.

src/wrapper/bloom_callback.rs
wuranxx: I test current implementation, valkey will auto add `PEXPIREAT` to aof file. (RedisBloom also relies on Redis's ability to automatically add expiration commands.)

command:
```sh
127.0.0.1:6379> bf.reserve key1 0.01 10
OK
127.0.0.1:6379> bf.reserve key2 0.01 15
OK
127.0.0.1:6379> expire key1 1000
(integer) 1
127.0.0.1:6379> BGREWRITEAOF
Background append only file rewriting started
```


aof file:
```
*2
$6
SELECT
$1
0
*3
$7
BF.LOAD
$4
key2
$216
<value>
*3
$7
BF.LOAD
$4
key1
$208
<value>
*3
$9
PEXPIREAT
$4
key1
$13
1729582928837
```

src/bloom/command_handler.rs
wuranxx: I check code, we don't implement versioning in rdb. Do we only implement versioning in AOF rewrites or do we need to implement RDB as well?

One possible implementation is to add a `version` field in `BloomFilterType`, not persisting this field during RDB saving, and assigning the current process's `version` value by default when reading from RDB.

src/wrapper/bloom_callback.rs
KarthikSubbarao: Sounds good. Thanks for checking. This looks like standard AOF behavior to include the expiration time by including the `PEXPIREAT` in the AOF.

Since this is the case, we can remove this sentence from the PR description:
```
(I did not process the expiration time. This solution may invalidate the expiration time. )
```

src/bloom/command_handler.rs
KarthikSubbarao: > I check code, we don't implement versioning in rdb. Do we only implement versioning in AOF rewrites or do we need to implement RDB as well?

We currently have a check for the encoding version here when handling RDB Load. If the encoding version is higher than what the running Module can handle, the object is not restored.

https://github.com/valkey-io/valkey-bloom/blob/unstable/src/bloom/data_type.rs#L57

tests/test_aofrewrite.py
wuranxx: I check the code, comand stats just update in `call` method. but when loading aof file, the `call` method not invoke, is invoke `cmd->proc`, so aof load will not update commandstats.

KarthikSubbarao: Ok, sounds good. Thanks for checking

wuranxx:
@KarthikSubbarao The review comments have been modified and replied. Please review them again.

src/lib.rs
KarthikSubbarao: If the main purpose of BF.DUMP is for testing, I would say that it is not needed since we will have both RDB and AOF as options for migration of bloom data.

src/bloom/utils.rs
```diff
@@ -18,6 +20,10 @@ pub const ERROR_RATE_RANGE: &str = "ERR (0 < error rate range < 1)";
 pub const CAPACITY_LARGER_THAN_0: &str = "ERR (capacity should be larger than 0)";
 pub const MAX_NUM_SCALING_FILTERS: &str = "ERR max number of scaling filters reached";
 pub const UNKNOWN_ARGUMENT: &str = "ERR unknown argument received";
+pub const INVALID_ARGUMENT: &str = "ERR invalid argument received";
+pub const KEY_EXISTS: &str = "-BUSYKEY Target key name already exists.";
+
+pub const BLOOM_TYPE_VERSION: u8 = 1;
```
KarthikSubbarao: Can we add a comment to clarify this is intended to version the `BloomFilterType` structure mainly for the purpose of AOF Rewrite? 

Note: We already have a version `BLOOM_FILTER_TYPE_ENCODING_VERSION` that exists in `bloom/data_type.rs` and it serves the purpose of versioning the RDB data. Since the new constant you are adding is for versioning (for mainly AOF), we can add it and specify it above.

src/bloom/utils.rs
```diff
@@ -150,6 +160,41 @@ impl BloomFilterType {
         }
         Ok(0)
     }
+
+    pub fn encoder_bloom_filter(&self) -> Result<Vec<u8>, Box<bincode::ErrorKind>> {
```
KarthikSubbarao: 1. Can we add documentation here?

2. Also, can we simply use the `BloomError` enum instead of `bincode::Error`? For example, can we return `BloomError:: ERROR`?  If there is a particular case where we want a custom error message that we want to return back to the client, you can extend this enum and add more error cases

src/bloom/utils.rs
```diff
@@ -150,6 +160,41 @@ impl BloomFilterType {
...
+        let vec = bincode::serialize(self)?;
+        Ok(vec)
+    }
+
+    pub fn decoder_bloom_filter(
```
KarthikSubbarao: Can we add documentation here? The most important part here is to mention this function should be backwards compatible. If a new version of `BloomFilterType` is made (when we bump up `BLOOM_TYPE_VERSION`), we need to update this function to handle **both** the existing and the new versions and deserialize accordingly.

We will need to handle any current or previous version and deserializing the bytes into a bloom object of the running Module's current version `BLOOM_TYPE_VERSION`.

src/bloom/utils.rs
```diff
@@ -150,6 +160,41 @@ impl BloomFilterType {
...
+        }
+        let version = decoded_bytes[0];
+        match version {
+            1 => {
+                // always use new version to init bloomFilterType.
+                // This is to ensure that the new fields can be recognized when the object is serialized and deserialized in the future.
+                let (_, expansion, fp_rate, filters): (u8, u32, f32, Vec<BloomFilter>) =
+                    bincode::deserialize(decoded_bytes)?;
+                let filter = BloomFilterType {
+                    version: BLOOM_TYPE_VERSION,
```
KarthikSubbarao: Question: If the `BloomFilter` itself (Not `BloomFilterType`) structure was updated, we would still bump up the `BLOOM_TYPE_VERSION`, and we would have to decode by keeping the old version of the `BloomFilter` structure existing in the code base.

Is this is the expectation you had in mind?

src/bloom/utils.rs
```diff
@@ -150,6 +160,41 @@ impl BloomFilterType {
...
     }
+
+    pub fn encoder_bloom_filter(&self) -> Result<Vec<u8>, Box<bincode::ErrorKind>> {
+        let vec = bincode::serialize(self)?;
+        Ok(vec)
+    }
+
+    pub fn decoder_bloom_filter(
+        decoded_bytes: &[u8],
+    ) -> Result<BloomFilterType, Box<dyn std::error::Error>> {
```
KarthikSubbarao: Can we simply use the `BloomError` enum instead of `bincode::Error`?

For example, can we return `BloomError:: ERROR`? 

If there is a particular case where we want a custom error message that we want to return back to the client, you can extend this enum and add more error cases

src/bloom/utils.rs
KarthikSubbarao: Can we cover the two error cases from a test (integration or unit)?
1. Empty / Invalid input 
2. Unsupported version

src/bloom/utils.rs
wuranxx: yes, when any structure changes, the version needs to be updated.

wuranxx:
@KarthikSubbarao 
I think another purpose of `BF.DUMP` is for migration scenarios. When tools like `redisshake` perform data migration between two Redis instances, the `valkey-bloom` type keys can be replayed using `BF.LOAD` and `BF.DUMP`.

If you believe this command is ultimately unnecessary, please let me know, and I will remove it. (About `BF.DUMP` review can't reply, so i add a new comment to reply it.)

src/bloom/utils.rs
wuranxx: ok, i will add comment an move this const to `data_type.rs`.

src/bloom/utils.rs
wuranxx: 1. of cousrse, I will add it.
2. I'm not sure whether to reference `ValkeyModule` in `utils.rs` because I want any errors to be logged for easier debugging. If we don't log the errors in `utils.rs`, then we can return the errors to the outer layer, where they can be logged and converted to `BloomError` before being returned to the client.

src/bloom/utils.rs
wuranxx: fixed.

src/bloom/utils.rs
wuranxx: added.

wuranxx: added.

src/bloom/utils.rs
wuranxx: fixed.

src/bloom/utils.rs
wuranxx: fixed.

wuranxx:
@KarthikSubbarao  The review comments have been modified and replied. Please review them again, thank you.

src/bloom/utils.rs
```diff
@@ -35,14 +46,19 @@ impl BloomError {
...
             BloomError::MaxNumScalingFilters => MAX_NUM_SCALING_FILTERS,
             BloomError::ExceedsMaxBloomSize => EXCEEDS_MAX_BLOOM_SIZE,
+            BloomError::EncodeBloomFilterFailed => ENCODE_BLOOM_FILTER_FAILED,
+            BloomError::DecodeBloomFilterFailed => DECODE_BLOOM_FILTER_FAILED,
+            BloomError::DecodeNotSupportVersion => DECODE_NOT_SUPPORT_VERSION,
         }
     }
 }
 
 /// The BloomFilterType structure. 32 bytes.
```
zackcam: Could you update the description with the new size of BloomFilterType. Also just a note the current size before adding version is wrong it is 40 bytes after a recent change of fp_rate to f64. I think this should now be 41 bytes

tests/test_aofrewrite.py
zackcam, line 1: Do we want a test with a bloomfilter that has been scaled out? If you want help on how to make a test like this: [scaled example test](https://github.com/valkey-io/valkey-bloom/pull/19/files#diff-2cc2fa9ddbc2d9ae0b6dabfb52eaaf0f0ce54f0796437f683835c69920af5537R82 )

src/bloom/utils.rs
```diff
@@ -177,6 +195,43 @@ impl BloomFilterType {
...
+        match version {
+            1 => {
+                // always use new version to init bloomFilterType.
+                // This is to ensure that the new fields can be recognized when the object is serialized and deserialized in the future.
+                let (_, expansion, fp_rate, filters): (u8, u32, f64, Vec<BloomFilter>) =
+                    match bincode::deserialize(decoded_bytes) {
+                        Ok(_v) => _v,
+                        Err(_) => return Err(BloomError::DecodeBloomFilterFailed),
+                    };
+                let filter = BloomFilterType {
```
KarthikSubbarao: Can we add a memory usage limit check here? 

If any filter in the vector is greater than the allowed memory usage limit, we should reject it 
You can refer to this https://github.com/valkey-io/valkey-bloom/blob/unstable/src/bloom/utils.rs#L167-L170

Ideally, we want to reject the request even before we allocate the BloomFilter object if we estimate it will use more than allowed max limit

src/bloom/utils.rs
```diff
@@ -21,12 +24,20 @@ pub const MAX_NUM_SCALING_FILTERS: &str = "ERR bloom object reached max number o
 pub const UNKNOWN_ARGUMENT: &str = "ERR unknown argument received";
 pub const EXCEEDS_MAX_BLOOM_SIZE: &str =
     "ERR operation results in filter allocation exceeding size limit";
+pub const INVALID_ARGUMENT: &str = "ERR invalid argument received";
+pub const KEY_EXISTS: &str = "-BUSYKEY Target key name already exists.";
+pub const ENCODE_BLOOM_FILTER_FAILED: &str = "ERR encode bloom filter failed.";
+pub const DECODE_BLOOM_FILTER_FAILED: &str = "ERR decode bloom filter failed.";
+pub const DECODE_NOT_SUPPORT_VERSION: &str = "ERR decode bloom filter failed. not support version.";
```
KarthikSubbarao: We can update this to: `pub const DECODE_UNSUPPORTED_VERSION: &str = "ERR decode bloom filter failed.  Unsupported version.";`

src/bloom/utils.rs
```diff
@@ -21,12 +24,20 @@ pub const MAX_NUM_SCALING_FILTERS: &str = "ERR bloom object reached max number o
 pub const UNKNOWN_ARGUMENT: &str = "ERR unknown argument received";
 pub const EXCEEDS_MAX_BLOOM_SIZE: &str =
     "ERR operation results in filter allocation exceeding size limit";
+pub const INVALID_ARGUMENT: &str = "ERR invalid argument received";
```
KarthikSubbarao: If this is not used, can we remove it?

src/bloom/utils.rs
```diff
@@ -35,14 +46,19 @@ impl BloomError {
...
         }
     }
 }
 
 /// The BloomFilterType structure. 32 bytes.
 /// Can contain one or more filters.
 /// This is a generic top level structure which is not coupled to any bloom crate.
+#[derive(Serialize, Deserialize)]
 pub struct BloomFilterType {
+    pub version: u8,
```
KarthikSubbarao: Nit / Observation: We don't need to have `version` within the structure. It is only used during encode and decode operations. So, we just need to include `version` in the first byte (u8) of the encode operation's generated output.

zackcam:
Quick note: When you get the latest changes your changes will cause the tests in `test_bloom_metrics.py` to fail, due to the new size of BloomFilterType, to fix this you can update this [file and line](https://github.com/valkey-io/valkey-bloom/blob/unstable/tests/test_bloom_metrics.py#L7) to be `DEFAULT_BLOOM_FILTER_SIZE = 179961`.

tests/test_aofrewrite.py
wuranxx: Thank you! After adding tests, it was found that encode did not include the statistical values, which have now been added.

src/bloom/utils.rs
wuranxx: already remove version filed. fixed.

src/bloom/utils.rs
wuranxx: I have added limit checks for  `expansion`, `fp_rate` and `filters.len()` . If there are any other checks to be added, please let me know.

src/bloom/utils.rs
wuranxx: fixed.

src/bloom/utils.rs
wuranxx: removed.

src/bloom/utils.rs
wuranxx: Remove version field and just add in encode method.

wuranxx:
> Quick note: When you get the latest changes your changes will cause the tests in `test_bloom_metrics.py` to fail, due to the new size of BloomFilterType, to fix this you can update this [file and line](https://github.com/valkey-io/valkey-bloom/blob/unstable/tests/test_bloom_metrics.py#L7) to be `DEFAULT_BLOOM_FILTER_SIZE = 179961`.

I've removed the version field from `BloomFitlerType`.

zackcam (approved):
LGTM

src/bloom/utils.rs
```diff
@@ -188,6 +208,77 @@ impl BloomFilterType {
...
+        match version {
+            1 => {
+                // always use new version to init bloomFilterType.
+                // This is to ensure that the new fields can be recognized when the object is serialized and deserialized in the future.
+                let (expansion, fp_rate, filters): (u32, f64, Vec<BloomFilter>) =
+                    match bincode::deserialize::<(u32, f64, Vec<BloomFilter>)>(&decoded_bytes[1..])
+                    {
+                        Ok(values) => {
+                            if values.0 == 0 {
+                                return Err(BloomError::NonScalingFilterFull);
```
KarthikSubbarao: Since an expansion with 0 is not possible and invalid, can we return this error instead? :  `BloomError::BAD_EXPANSION`

src/bloom/utils.rs
```diff
@@ -188,6 +208,77 @@ impl BloomFilterType {
...
+                            }
+                            if values.1 < BLOOM_FP_RATE_MIN || values.1 > BLOOM_FP_RATE_MAX {
```
KarthikSubbarao: The correct validation is this:
 ```
if !(values.1 > BLOOM_FP_RATE_MIN && values.1 < BLOOM_FP_RATE_MAX) {
    return Err(BloomError::ErrorRateRange);
}
 ```
 
 This is because `BLOOM_FP_RATE_MIN` is 0 and `BLOOM_FP_RATE_MAX` is 1.

src/bloom/utils.rs
```diff
@@ -188,6 +208,77 @@ impl BloomFilterType {
...
+                                return Err(BloomError::ErrorRateRange);
+                            }
+                            if values.2.len() >= configs::MAX_FILTERS_PER_OBJ as usize {
+                                return Err(BloomError::MaxNumScalingFilters);
+                            }
+                            values
```
KarthikSubbarao, line 267: Can we add a check for the memory usage of the bloom filters?

`validate_size_limit` will need to be passed in from `command_handler::bloom_filter_load`. See: https://github.com/valkey-io/valkey-bloom/blob/unstable/src/bloom/command_handler.rs#L386
```
    let validate_size_limit = !ctx.get_flags().contains(ContextFlags::REPLICATED);
```

We will need to loop through the vector of filters and check that all of them are within the memory limit.

```
            // Reject the request, if the operation will result in creation of a filter of size greater than what is allowed.
            if validate_size_limit && !filter.number_of_bytes() > configs::BLOOM_MEMORY_LIMIT_PER_FILTER.load(Ordering::Relaxed) as u64 {
                return Err(BloomError::ExceedsMaxBloomSize);
            }
```

src/bloom/utils.rs
```diff
@@ -21,12 +27,21 @@ pub const MAX_NUM_SCALING_FILTERS: &str = "ERR bloom object reached max number o
...
+    "ERR decode bloom filter failed.  Unsupported version.";
 
 #[derive(Debug, PartialEq)]
 pub enum BloomError {
     NonScalingFilterFull,
     MaxNumScalingFilters,
     ExceedsMaxBloomSize,
+    EncodeBloomFilterFailed,
+    DecodeBloomFilterFailed,
+    DecodeNotSupportVersion,
```
KarthikSubbarao: We can rename to `DecodeUnsupportedVersion `

KarthikSubbarao:
The PR is almost ready to be merged in. Just have few comments on fixing the validation logic of bloom filters restored from BF.LOAD. We can merge it in after this

src/bloom/command_handler.rs
```diff
@@ -490,3 +490,47 @@ pub fn bloom_filter_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
...
+        }
+        None => {
+            // if filter not exists, create it.
+            let hex = value.to_vec();
+            let bf = match BloomFilterType::decode_bloom_filter(&hex) {
+                Ok(v) => v,
+                Err(_) => {
+                    return Err(ValkeyError::Str(utils::ERROR));
+                }
+            };
```
KarthikSubbarao: Currently, we are not returning back the errors from the `decode_bloom_filter` operation. 

Can we update the error handling here to use the result and return it back to the client?
```
Err(err) => return Err(ValkeyError::Str(err.as_str())),
```

src/bloom/utils.rs
```diff
@@ -21,12 +27,21 @@ pub const MAX_NUM_SCALING_FILTERS: &str = "ERR bloom object reached max number o
 pub const UNKNOWN_ARGUMENT: &str = "ERR unknown argument received";
 pub const EXCEEDS_MAX_BLOOM_SIZE: &str =
     "ERR operation results in filter allocation exceeding size limit";
+pub const KEY_EXISTS: &str = "-BUSYKEY Target key name already exists.";
```
KarthikSubbarao: Just tested out the code with this PR.

We will currently get back the following:
```
(error) -BUSYKEY Target key name already exists.
```

Can we remove the `-`?

src/bloom/data_type.rs
```diff
@@ -22,7 +26,7 @@ pub static BLOOM_FILTER_TYPE: ValkeyType = ValkeyType::new(
         rdb_load: Some(bloom_callback::bloom_rdb_load),
         rdb_save: Some(bloom_callback::bloom_rdb_save),
         // TODO
```
KarthikSubbarao: We can remove the TODO comment :)

src/bloom/data_type.rs
```diff
@@ -12,6 +12,10 @@ use std::os::raw::c_int;
 use valkey_module::native_types::ValkeyType;
 use valkey_module::{logging, raw};
 
+/// Use for decode an encode `BloomFilterType`.
```
KarthikSubbarao: Nit: We can update the comment to: ``Used for decoding and encoding `BloomFilterType`. Currently used in AOF Rewrite.``

src/lib.rs
```diff
@@ -68,6 +68,12 @@ fn info_handler(ctx: &InfoContext, _for_crash_report: bool) -> ValkeyResult<()>
     bloom_info_handler(ctx)
 }
 
+/// Command handler for:
+/// BF.LOAD <key> data
+fn bloom_load_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
+    command_handler::bloom_filter_load(ctx, &args)
+}
+
```
KarthikSubbarao: Nit: We can move this to be right after `bloom_insert_command`

src/bloom/command_handler.rs
wuranxx: fixed

src/bloom/utils.rs
wuranxx: Of course.  I used this error because the error that occurs when the key exists in the `restore` command of `valkey` is this one.

src/lib.rs
wuranxx: fixed.

src/bloom/data_type.rs
wuranxx: fixed.

src/bloom/utils.rs
wuranxx: fixed.

src/bloom/utils.rs
wuranxx: fixed.

src/bloom/utils.rs
wuranxx: fixed.

src/bloom/utils.rs
wuranxx: The judgment condition here seems to be reversed. Maybe correct is:
```
if validate_size_limit && filter.number_of_bytes() > configs::BLOOM_MEMORY_LIMIT_PER_FILTER.load(Ordering::Relaxed) as usize {
                return Err(BloomError::ExceedsMaxBloomSize);
            }
```

KarthikSubbarao: Yes, sorry, that was a typo in my comment - I did not check. This looks good

src/bloom/utils.rs
```diff
@@ -188,6 +210,91 @@ impl BloomFilterType {
...
+            1 => {
+                // always use new version to init bloomFilterType.
+                // This is to ensure that the new fields can be recognized when the object is serialized and deserialized in the future.
+                let (expansion, fp_rate, filters): (u32, f64, Vec<BloomFilter>) =
+                    match bincode::deserialize::<(u32, f64, Vec<BloomFilter>)>(&decoded_bytes[1..])
+                    {
+                        Ok(values) => {
+                            if values.0 == 0 {
+                                return Err(BloomError::BadExpansion);
+                            }
```
KarthikSubbarao: We will need a validation here since we have a limit on max expansion. See how we do this in other places: https://github.com/valkey-io/valkey-bloom/blob/unstable/src/bloom/command_handler.rs#L357C32-L357C87

KarthikSubbarao (approved):

src/bloom/utils.rs
```diff
@@ -621,4 +729,122 @@ mod tests {
...
+        let vec = bf.encode_bloom_filter().unwrap();
+        // should return error
+        assert_eq!(
+            BloomFilterType::decode_bloom_filter(&vec, true).err(),
+            Some(BloomError::ErrorRateRange)
+        );
+        bf.fp_rate = origin_fp_rate;
+
+        // 3. unsupport filter size:
+        configs::BLOOM_MEMORY_LIMIT_PER_FILTER.store(1, Ordering::SeqCst);
```
KarthikSubbarao: Changing this config from a unit test will cause other unit tests to fail. Unit tests run in parallel and reference the same static config variable.

We see this happen in this GitHub Approval Workflow test run: https://github.com/valkey-io/valkey-bloom/actions/runs/11816522355/job/32956192712?pr=17

Would not suggest changing this configuration from the unit test. We can do this safely from the python integration tests

In order to merge in this PR, can we fix this, @wuranxx ?

wuranxx: Of course， i will fix it.

src/bloom/utils.rs
wuranxx: fixed.

KarthikSubbarao (approved):

KarthikSubbarao:
Thank you, @wuranxx . Can you update the PR Description? It currently references BF.DUMP. We can update it to reference BF.LOAD and I will merge the PR after that

wuranxx:
> Thank you, @wuranxx . Can you update the PR Description? It currently references BF.DUMP. We can update it to reference BF.LOAD and I will merge the PR after that

updated.

KarthikSubbarao:
Thank you for helping with AOF Rewrite support, @wuranxx !

PR #18, opened 2024-10-25
Handle bloom object max allowed size limit, switch fp rate to f64

KarthikSubbarao:
* Implement a memory / size based limit per filter within a bloom object. This is a new module config: `bf.bloom-memory-limit-per-filter`.
  * This means every write operation (bloom object creation AND scale outs) that results in a new filter requires the new filter to be within allowed limits. If not, the write request is rejected with an error.
  * Similarly, any RDB Load operation also performs the same memory / size based limit check. RDB Load will fail if any filter within a bloom object has a size larger than what is allowed.
* Switch fp_rate handling through out all occurrences from f32 to f64 to have a higher precision. The main reason for this is during scaling out, every new filter will have a smaller fp_rate (due to the TIGHTENING_RATIO) to maintain the overall fp_rate. We need to handle higher precision fp_rates with every new scale out and f64 handles this better.
* defrag callback was updated to use the new config (`bf.bloom-memory-limit-per-filter`) to exempt defrag operations when over this limit.


YueTang-Vanessa (approved):

src/bloom/utils.rs
```diff
@@ -16,20 +17,24 @@ pub const BAD_CAPACITY: &str = "ERR bad capacity";
 pub const BAD_ERROR_RATE: &str = "ERR bad error rate";
 pub const ERROR_RATE_RANGE: &str = "ERR (0 < error rate range < 1)";
 pub const CAPACITY_LARGER_THAN_0: &str = "ERR (capacity should be larger than 0)";
-pub const MAX_NUM_SCALING_FILTERS: &str = "ERR max number of scaling filters reached";
+pub const MAX_NUM_SCALING_FILTERS: &str = "ERR bloom object reached max number of filters";
 pub const UNKNOWN_ARGUMENT: &str = "ERR unknown argument received";
+pub const EXCEEDS_MAX_BLOOM_SIZE: &str =
+    "ERR operation results in filter allocation exceeding limit";
```
zackcam: nit: Would we want to specify that it failed due to size limit and not some other limit like time? I know we don't have an error like time taking too long but could be nice to be explicit

zackcam (approved):
Just one small nit/question but lgtm!

src/lib.rs
```diff
@@ -82,6 +82,7 @@ valkey_module! {
         i64: [
             ["bloom-max-item-size", &*configs::BLOOM_CAPACITY, configs::BLOOM_CAPACITY_DEFAULT, configs::BLOOM_CAPACITY_MIN as i64, configs::BLOOM_CAPACITY_MAX as i64, ConfigurationFlags::DEFAULT, None],
             ["bloom-expansion-rate", &*configs::BLOOM_EXPANSION, configs::BLOOM_EXPANSION_DEFAULT, configs::BLOOM_EXPANSION_MIN as i64, configs::BLOOM_EXPANSION_MAX as i64, ConfigurationFlags::DEFAULT, None],
+            ["bloom-max-memory-usage", &*configs::BLOOM_MAX_MEMORY_USAGE, configs::BLOOM_MAX_MEMORY_USAGE_DEFAULT, configs::BLOOM_MAX_MEMORY_USAGE_MIN, configs::BLOOM_MAX_MEMORY_USAGE_MAX, ConfigurationFlags::DEFAULT, None],
```
KarthikSubbarao: Will look into renaming this. Reason: The name should explain what it is more accurately, which is the limit on the filter, not the entire bloom object.

Considering: `bloom-memory-limit-per-filter`

PR #19, opened 2024-10-31
Adding info handler, that contains two fields, num objects and memory…

zackcam:
Added an info handler which required updating Cargo.toml to include valkey-module-macros and linkme. The info handler contains two fields bloom_total_memory_bytes and bloom_num_objects. I then updated the creation and copy methods of bloom objects to increment these values where necessary. In order to deincrement added impl Drop for bloom in order to decrease the info fields as necessary.

Testing:
Created a new test file specifically for info. This test file contains for different tests:
test_basic_command_metrics: This tests all create commands and non create commands to check that create commands increase the info metrics while non create dont have an influence.
test_scaled_bloomfilter_metrics: Tests scaling a bloom filter increases the metrics as desired
test_copy_metrics: Tests copying a bloom filter will double the metrics as desired
test_save_and_restore_metrics: Tests saving and restoring a scaled and non scaled bloom filter will increase metrics as desired

### DIFF
--- a/src/bloom/data_type.rs
+++ b/src/bloom/data_type.rs
@@ -3,8 +3,11 @@ use crate::bloom::utils::BloomFilterType;
 use crate::configs::{
     FIXED_SIP_KEY_ONE_A, FIXED_SIP_KEY_ONE_B, FIXED_SIP_KEY_TWO_A, FIXED_SIP_KEY_TWO_B,
 };
+use crate::metrics::BLOOM_NUM_OBJECTS;
+use crate::metrics::BLOOM_OBJECT_TOTAL_MEMORY_BYTES;
 use crate::wrapper::bloom_callback;
 use crate::MODULE_NAME;
+use std::mem;
 use std::os::raw::c_int;
 use valkey_module::native_types::ValkeyType;
 use valkey_module::{logging, raw};
@@ -108,6 +111,11 @@ impl ValkeyDataType for BloomFilterType {
             );
             filters.push(filter);
         }
+        BLOOM_OBJECT_TOTAL_MEMORY_BYTES.fetch_add(
+            mem::size_of::<BloomFilterType>(),
+            std::sync::atomic::Ordering::Relaxed,
+        );
+        BLOOM_NUM_OBJECTS.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
         let item = BloomFilterType {
             expansion: expansion as u32,
             fp_rate,
--- a/src/bloom/utils.rs
+++ b/src/bloom/utils.rs
@@ -1,6 +1,6 @@
-use crate::configs;
+use crate::{configs, metrics};
 use bloomfilter;
-use std::sync::atomic::Ordering;
+use std::{mem, sync::atomic::Ordering};
 
 /// KeySpace Notification Events
 pub const ADD_EVENT: &str = "bloom.add";
@@ -39,7 +39,7 @@ impl BloomError {
     }
 }
 
-/// The BloomFilterType structure. 32 bytes.
+/// The BloomFilterType structure. 40 bytes.
 /// Can contain one or more filters.
 /// This is a generic top level structure which is not coupled to any bloom crate.
 pub struct BloomFilterType {
@@ -61,6 +61,12 @@ impl BloomFilterType {
         if validate_size_limit && !BloomFilter::validate_size(capacity, fp_rate) {
             return Err(BloomError::ExceedsMaxBloomSize);
         }
+        metrics::BLOOM_NUM_OBJECTS.fetch_add(1, Ordering::Relaxed);
+        metrics::BLOOM_OBJECT_TOTAL_MEMORY_BYTES.fetch_add(
+            mem::size_of::<BloomFilterType>(),
+            std::sync::atomic::Ordering::Relaxed,
+        );
+
         // Create the bloom filter and add to the main BloomFilter object.
         let bloom = BloomFilter::new(fp_rate, capacity);
         let filters = vec![bloom];
@@ -75,6 +81,11 @@ impl BloomFilterType {
     /// Create a new BloomFilterType object from an existing one.
     pub fn create_copy_from(from_bf: &BloomFilterType) -> BloomFilterType {
         let mut filters = Vec::new();
+        metrics::BLOOM_NUM_OBJECTS.fetch_add(1, Ordering::Relaxed);
+        metrics::BLOOM_OBJECT_TOTAL_MEMORY_BYTES.fetch_add(
+            mem::size_of::<BloomFilterType>(),
+            std::sync::atomic::Ordering::Relaxed,
+        );
         for filter in &from_bf.filters {
             let new_filter = BloomFilter::create_copy_from(filter);
             filters.push(new_filter);
@@ -199,11 +210,16 @@ impl BloomFilter {
             fp_rate,
             &configs::FIXED_SEED,
         );
-        BloomFilter {
+        let fltr = BloomFilter {
             bloom,
             num_items: 0,
             capacity,
-        }
+        };
+        metrics::BLOOM_NUM_FILTERS_ACROSS_OBJECTS
+            .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+        metrics::BLOOM_OBJECT_TOTAL_MEMORY_BYTES
+            .fetch_add(fltr.number_of_bytes(), std::sync::atomic::Ordering::Relaxed);
+        fltr
     }
 
     /// Create a new BloomFilter from dumped information (RDB load).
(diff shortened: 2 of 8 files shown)


src/configs.rs
```diff
@@ -26,6 +26,8 @@ lazy_static! {
     pub static ref BLOOM_EXPANSION: AtomicI64 = AtomicI64::new(BLOOM_EXPANSION_DEFAULT);
     pub static ref BLOOM_MEMORY_LIMIT_PER_FILTER: AtomicI64 =
         AtomicI64::new(BLOOM_MEMORY_LIMIT_PER_FILTER_DEFAULT);
+    pub static ref BLOOM_NUM_OBJECTS: AtomicUsize = AtomicUsize::new(0);
+    pub static ref BLOOM_OBJECT_TOTAL_MEMORY_BYTES: AtomicUsize = AtomicUsize::new(0);
```
KarthikSubbarao: This works. But for clarity, can we create a `src/metrics.rs` file? We can add both the lazy static and the info handler macro there

src/bloom/utils.rs
```diff
@@ -199,11 +207,15 @@ impl BloomFilter {
...
             &configs::FIXED_SEED,
         );
-        BloomFilter {
+        let fltr = BloomFilter {
             bloom,
             num_items: 0,
             capacity,
-        }
+        };
+        configs::BLOOM_NUM_OBJECTS.fetch_add(1, Ordering::Relaxed);
```
KarthikSubbarao: BLOOM_NUM_OBJECTS is the number of bloom objects. 

We can add a different metric, `BLOOM_NUM_FILTERS`, to track the number of filters created as a result of existing bloom objects

src/bloom/utils.rs
```diff
@@ -56,6 +56,10 @@ impl BloomFilterType {
         expansion: u32,
         validate_size_limit: bool,
     ) -> Result<BloomFilterType, BloomError> {
+        configs::BLOOM_OBJECT_TOTAL_MEMORY_BYTES.fetch_add(
```
KarthikSubbarao: We should increment `BLOOM_NUM_OBJECTS ` here, right?

src/bloom/utils.rs
```diff
@@ -221,11 +233,15 @@ impl BloomFilter {
...
             sip_keys,
         );
-        BloomFilter {
+        let fltr = BloomFilter {
             bloom,
             num_items,
             capacity,
-        }
+        };
+        configs::BLOOM_NUM_OBJECTS.fetch_add(1, Ordering::Relaxed);
```
KarthikSubbarao: We should not increment `BLOOM_NUM_OBJECTS` here. But, instead increment `BLOOM_NUM_FILTERS`.

You can first update all the right places for these two metrics. I will review the next commit where these are fixed

KarthikSubbarao (changes requested):

src/bloom/utils.rs
YueTang-Vanessa: Why increase this metric here? What if there has error? Maybe consider to move this few lines below?

src/bloom/data_type.rs
```diff
@@ -108,6 +110,10 @@ impl ValkeyDataType for BloomFilterType {
             );
             filters.push(filter);
         }
+        BLOOM_OBJECT_TOTAL_MEMORY_BYTES.fetch_add(
+            mem::size_of::<BloomFilterType>(),
```
YueTang-Vanessa, line 115: Will there has duplicate memory add here as few lines above BloomFilter::from_existing will also increase same metric `BLOOM_OBJECT_TOTAL_MEMORY_BYTES`?

zackcam: I don't think so this never has from_exisitng or new called for BloomFilterType as far as I could see. This should also be caught in my rdb test if it would double up. I think its confusing because there is a from_existing in both BloomFilter and BloomFilterType. Does this make more sense now?

zackcam:
Accidently closed the pr, reopening with the following changes:

1. Moved the info handler to a separate file as well as the info metrics.
2. Created a new metric for number of filters (Changed the other metric to actually be number of bloom objects)
3. Updated tests to include this new metric
4. Moved an increment of metrics to be after a possible thrown error

tests/test_bloom_metrics.py
```diff
@@ -0,0 +1,154 @@
...
+            if line.startswith('bf_bloom_total_memory_bytes:'):
+                total_memory_bites = int(line.split(':')[1])
+            elif line.startswith('bf_bloom_num_objects:'):
+                num_objects = int(line.split(':')[1])
+            elif line.startswith('bf_bloom_num_filters'):
+                num_filters = int(line.split(':')[1])
+
+
+        assert total_memory_bites == expected_memory 
+        assert num_objects == expected_num_objects
```
YueTang-Vanessa: Missing `assert num_filters == expected_num_filters`

tests/test_bloom_metrics.py
```diff
@@ -0,0 +1,154 @@
...
+        num_filters = -1
+        for line in lines:
+            if line.startswith('bf_bloom_total_memory_bytes:'):
+                total_memory_bites = int(line.split(':')[1])
+            elif line.startswith('bf_bloom_num_objects:'):
+                num_objects = int(line.split(':')[1])
+            elif line.startswith('bf_bloom_num_filters'):
+                num_filters = int(line.split(':')[1])
+
+
```
YueTang-Vanessa: Nit pick: may just keep one newline

tests/test_bloom_metrics.py
```diff
@@ -0,0 +1,154 @@
...
+from valkeytests.conftest import resource_port_tracker
+from util.waiters import *
+
+DEFAULT_BLOOM_FILTER_SIZE = 179960
+class TestBloommetrics(ValkeyBloomTestCaseBase):
+
+    '''
+    Test data set unit basic metrics
+    '''
+    def test_basic_command_metrics(self):
```
YueTang-Vanessa, line 10: Can you also add one test case of key expiration? Use `EXPIRE` for a bloom filter key, and see if all metrics works as expected.

zackcam: Added this case in test_basic_command_metrics

YueTang-Vanessa (approved):
LGTM

src/metrics.rs
```diff
@@ -0,0 +1,32 @@
...
+            BLOOM_OBJECT_TOTAL_MEMORY_BYTES
+                .load(Ordering::Relaxed)
+                .to_string(),
+        )?
+        .field(
+            "bloom_num_objects",
+            BLOOM_NUM_OBJECTS.load(Ordering::Relaxed).to_string(),
+        )?
+        .field(
+            "bloom_num_filters",
```
KarthikSubbarao: `bloom_num_filters_across_objects`

src/metrics.rs
```diff
@@ -0,0 +1,32 @@
+use lazy_static::lazy_static;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use valkey_module::{InfoContext, ValkeyResult};
+
+lazy_static! {
+    pub static ref BLOOM_NUM_OBJECTS: AtomicUsize = AtomicUsize::new(0);
```
KarthikSubbarao: Can we have this as a u64?

src/metrics.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub static ref BLOOM_OBJECT_TOTAL_MEMORY_BYTES: AtomicUsize = AtomicUsize::new(0);
+    pub static ref BLOOM_NUM_FILTERS: AtomicUsize = AtomicUsize::new(0);
```
KarthikSubbarao: Can we have this as a u64?

tests/test_bloom_metrics.py
```diff
@@ -0,0 +1,166 @@
...
+        for i in range(1, len(original_info_obj), 2):
+            assert original_info_obj[i] == restored_info_obj[i]
+
+        self.verify_bloom_metrics(self.client.execute_command("INFO bf"), original_info_obj[3] + DEFAULT_BLOOM_FILTER_SIZE, 2, 3)
+
+    # Verify the metric values are recorded properly, the expected values are as below
+    # expected_memory: the size of the memory used by the objects
+	# expected_num_objects: the number of module objects stored
+    # expected_num_filters: the number of filters currently created
+    def verify_bloom_metrics(self, info_response, expected_memory, expected_num_objects, expected_num_filters):
```
KarthikSubbarao: Can we move this into the main valkey bloom test class?

zackcam: Moved and comment is updated

src/lib.rs
```diff
@@ -57,6 +60,14 @@ fn bloom_insert_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult
     command_handler::bloom_filter_insert(ctx, &args)
 }
 
+///
+/// Module Info
+///
+#[info_command_handler]
+fn add_info(ctx: &InfoContext, _for_crash_report: bool) -> ValkeyResult<()> {
+    create_info_handler(ctx)
```
KarthikSubbarao: Nit: A better name could be: `bloom_info_handler`

This fits well as if we add other data types, we can add more x_info_handlers to this function

src/lib.rs
```diff
@@ -57,6 +60,14 @@ fn bloom_insert_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult
     command_handler::bloom_filter_insert(ctx, &args)
 }
 
+///
+/// Module Info
+///
+#[info_command_handler]
+fn add_info(ctx: &InfoContext, _for_crash_report: bool) -> ValkeyResult<()> {
```
KarthikSubbarao: Nit: A better name could be: `info_handler`

PR #20, opened 2024-11-11
Add new metrics to show capacity and items across objects

YueTang-Vanessa:
Add two metrics `bloom_num_items_across_objects` and `bloom_capacity_across_objects` into the info section.

 * bloom_num_items_across_objects : indicate current total number of items across bloom filter objects
 * bloom_capacity_across_objects: indicate current total capacity across bloom filter objects

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -6,6 +6,8 @@ lazy_static! {
     pub static ref BLOOM_NUM_OBJECTS: AtomicU64 = AtomicU64::new(0);
     pub static ref BLOOM_OBJECT_TOTAL_MEMORY_BYTES: AtomicUsize = AtomicUsize::new(0);
     pub static ref BLOOM_NUM_FILTERS_ACROSS_OBJECTS: AtomicU64 = AtomicU64::new(0);
+    pub static ref BLOOM_NUM_ITEMS_ACROSS_OBJECTS: AtomicU64 = AtomicU64::new(0);
+    pub static ref BLOOM_CAPACITY_ACROSS_OBJECTS: AtomicU64 = AtomicU64::new(0);
 }
 
 pub fn bloom_info_handler(ctx: &InfoContext) -> ValkeyResult<()> {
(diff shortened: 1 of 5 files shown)


zackcam (commented):
I think we also need to update [this](https://github.com/valkey-io/valkey-bloom/blob/unstable/src/bloom/data_type.rs#L114). test_save_and_restore_metrics should be failing but isnt because of [self.client](https://github.com/valkey-io/valkey-bloom/blob/unstable/tests/test_bloom_metrics.py#L141)  being used and not new_client

YueTang-Vanessa:
> I think we also need to update [this](https://github.com/valkey-io/valkey-bloom/blob/unstable/src/bloom/data_type.rs#L114). test_save_and_restore_metrics should be failing but isnt because of [self.client](https://github.com/valkey-io/valkey-bloom/blob/unstable/tests/test_bloom_metrics.py#L141) being used and not new_client

The `BloomFilter::from_existing()` will change the metric

src/metrics.rs
```diff
@@ -27,6 +29,18 @@ pub fn bloom_info_handler(ctx: &InfoContext) -> ValkeyResult<()> {
...
+            BLOOM_NUM_ITEMS_ACROSS_OBJECTS
+                .load(Ordering::Relaxed)
+                .to_string(),
+        )?
+        .field(
+            "bloom_capacity_across_objects",
+            BLOOM_CAPACITY_ACROSS_OBJECTS
+                .load(Ordering::Relaxed)
+                .to_string(),
+        )?
```
KarthikSubbarao, line 43: These metrics will also be incremented during RDB Load. Can we test this out in (1) RDB Load restore test (2) COPY command test?

YueTang-Vanessa: RDB load will use from_existing function where the metrics get increased. The tests been covered in the test_bloom_metrics.py test_save_and_restore_metrics and test_copy_metrics

KarthikSubbarao (approved):

PR #23, opened 2024-11-19
Updating how we create BloomFilter from rdb loads and upgrading bloomfilter dependency to version 3.0

zackcam:
Before we were initializing a vector of bloomfilters with Vec::New() which could cause the size to be higher when we pushed an item into the vector during RDB Load and Copy. This is due to the raw alloc vector growth. From testing we saw that it could be four times as large. Which makes sense from [this](https://github.com/rust-lang/rust/issues/111307). In order to make this size what we expect we can just set the capacity to the number of filters we know we will be adding

This PR also upgrades the dependency of bloomfilter to 3.0.
Breaking changes included:
1. from_existing no longer exists, it is now from_slice which has different arguments
2. We no longer have an exposed sip_keys but have the seed. Updated tests to check for this.
3. Bitmap is no longer exposed we now have to use the as_slice method to get it
4. len() is now being used instead of number_of_bits()
5. Updated serialize and deserialize to now use the module from bloom

Other changes with this upgrade include removing unneeded fields from rdb load and save as they are now included in bitmap. Due to this we also changed the size validation in rdb load to now calcualte this off of fprate and capacity. This included creating a new helper method in BloomFilterType to claculate fp_rate based on number of filters


KarthikSubbarao (approved):

zackcam:
Breaking changes included: 
1. from_existing no longer exists, it is now from_slice which has different arguments
2. We no longer have an exposed sip_keys but have the seed. Updated tests to check for this.
3. Bitmap is no longer exposed we now have to use the as_slice method to get it
4. len() is now being used instead of number_of_bits()
5. Updated serialize and deserialize to now use the module from bloom

Cargo.toml
```diff
@@ -15,7 +15,7 @@ homepage = "https://github.com/valkey-io/valkey-bloom"
 valkey-module = "0.1.2"
 valkey-module-macros = "0"
 linkme = "0"
-bloomfilter = { version = "1.0.13", features = ["serde"] }
+bloomfilter = { version = "3", features = ["serde"] }
```
YueTang-Vanessa: Please change the version format `3` -> `3.0.0`

src/bloom/utils.rs
```diff
@@ -104,7 +105,7 @@ impl BloomFilterType {
 
     /// Create a new BloomFilterType object from an existing one.
     pub fn create_copy_from(from_bf: &BloomFilterType) -> BloomFilterType {
-        let mut filters = Vec::new();
+        let mut filters: Vec<BloomFilter> = Vec::with_capacity(from_bf.filters.len());
```
YueTang-Vanessa, line 108: Should we worry about over-allocate? ( when `filters.len() = 0`)

KarthikSubbarao: filters.len() will never / should never be 0, so this seems OK

src/bloom/data_type.rs
```diff
@@ -75,6 +74,8 @@ impl ValkeyDataType for BloomFilterType {
         let Ok(fp_rate) = raw::load_double(rdb) else {
             return None;
         };
+        let mut filters: Vec<BloomFilter> = Vec::with_capacity(num_filters as usize);
+
```
KarthikSubbarao, line 75: Few comments on RDB Save and Load.

1. Number of hash functions are included in the bitmap dump as a result of `to_bytes`. So we can remove this from the save/load logic. Right?
2. Do we need number of bits for any purpose? Is it needed for bloom filter restoration (external bloom)? If not, can we remove it? If it helps with any purpose (or if it makes size validation impossible), feel free to retain it

zackcam: We don't need both fields. I removed them both and instead moved the logic of checking the size until we have converted the bloom object into a type where we can get the number of bits

src/bloom/data_type.rs
```diff
@@ -134,11 +123,7 @@ impl ValkeyDataType for BloomFilterType {
         dig.add_long_long(self.expansion.into());
         dig.add_string_buffer(&self.fp_rate.to_le_bytes());
         for filter in &self.filters {
-            dig.add_string_buffer(&filter.bloom.bitmap());
-            for &(key1, key2) in &filter.sip_keys() {
-                dig.add_long_long(key1 as i64);
-                dig.add_long_long(key2 as i64);
-            }
+            dig.add_string_buffer(filter.bloom.as_slice());
```
KarthikSubbarao, line 127: Just Confirming. Does the slice contain the seed as well?

zackcam: Yes it is contained in the header portion of the bitmap. ` BitMap::set_seed(header, &seed);
` it is set like this in sync

src/bloom/data_type.rs
```diff
@@ -75,24 +71,27 @@ impl ValkeyDataType for BloomFilterType {
...
                 return None;
             };
-            let Ok(number_of_bits) = raw::load_unsigned(rdb) else {
+            let Ok(capacity) = raw::load_unsigned(rdb) else {
                 return None;
             };
-            // Reject RDB Load if any bloom filter within a bloom object of a size greater than what is allowed.
-            if !BloomFilter::validate_size_with_bits(number_of_bits) {
+            let new_fp_rate =
+                match Self::calculate_fp_rate(fp_rate, num_filters.try_into().unwrap()) {
```
KarthikSubbarao: Let's not unwrap

src/bloom/data_type.rs
```diff
@@ -75,24 +71,27 @@ impl ValkeyDataType for BloomFilterType {
...
+                    Ok(rate) => rate,
+                    Err(_) => {
+                        logging::log_warning("ERR bloom object reached max number of filters");
+                        return None;
+                    }
+                };
+            if !BloomFilter::validate_size(capacity.try_into().unwrap(), new_fp_rate) {
```
KarthikSubbarao: Let's not unwrap

src/bloom/utils.rs
```diff
@@ -330,7 +337,8 @@ impl BloomFilter {
             capacity as usize,
             fp_rate,
             &configs::FIXED_SEED,
-        );
+        )
+        .unwrap();
```
KarthikSubbarao: Let's not unwrap

src/bloom/utils.rs
```diff
@@ -346,20 +354,8 @@ impl BloomFilter {
...
-        capacity: u32,
-    ) -> BloomFilter {
-        let bloom = bloomfilter::Bloom::from_existing(
-            bitmap,
-            number_of_bits,
-            number_of_hash_functions,
-            sip_keys,
-        );
+    pub fn from_existing(bitmap: &[u8], num_items: u32, capacity: u32) -> BloomFilter {
+        let bloom = bloomfilter::Bloom::from_slice(bitmap).unwrap();
```
KarthikSubbarao: Let's not unwrap

src/bloom/data_type.rs
KarthikSubbarao: You can cast the u64 to a u32 using `as`

YueTang-Vanessa (approved):

src/bloom/data_type.rs
```diff
@@ -75,24 +71,26 @@ impl ValkeyDataType for BloomFilterType {
...
-            let Ok(number_of_bits) = raw::load_unsigned(rdb) else {
+            let Ok(capacity) = raw::load_unsigned(rdb) else {
                 return None;
             };
-            // Reject RDB Load if any bloom filter within a bloom object of a size greater than what is allowed.
-            if !BloomFilter::validate_size_with_bits(number_of_bits) {
+            let new_fp_rate = match Self::calculate_fp_rate(fp_rate, num_filters as i32) {
+                Ok(rate) => rate,
+                Err(_) => {
+                    logging::log_warning("ERR bloom object reached max number of filters");
```
KarthikSubbarao:
```suggestion
                    logging::log_warning("Failed to restore bloom object: Reached max number of filters");
```

src/bloom/utils.rs
```diff
@@ -231,6 +230,13 @@ impl BloomFilterType {
         }
     }
 
+    pub fn calculate_fp_rate(fp_rate: f64, num_filters: i32) -> Result<f64, BloomError> {
```
KarthikSubbarao, line 234:
```suggestion
    /// Calculate the false positive rate for the Nth filter using tightening ratio.
    pub fn calculate_fp_rate(fp_rate: f64, num_filters: i32) -> Result<f64, BloomError> {
```

src/bloom/data_type.rs
```diff
@@ -75,24 +71,26 @@ impl ValkeyDataType for BloomFilterType {
...
-            let Ok(number_of_bits) = raw::load_unsigned(rdb) else {
+            let Ok(capacity) = raw::load_unsigned(rdb) else {
                 return None;
             };
-            // Reject RDB Load if any bloom filter within a bloom object of a size greater than what is allowed.
-            if !BloomFilter::validate_size_with_bits(number_of_bits) {
+            let new_fp_rate = match Self::calculate_fp_rate(fp_rate, num_filters as i32) {
+                Ok(rate) => rate,
+                Err(_) => {
+                    logging::log_warning("Failed to restore bloom object: Reached max number of filters");
```
KarthikSubbarao:
```suggestion
                    logging::log_warning(
                        "Failed to restore bloom object: Reached max number of filters",
                    );
```

src/bloom/data_type.rs
```diff
@@ -75,24 +71,28 @@ impl ValkeyDataType for BloomFilterType {
         let Ok(fp_rate) = raw::load_double(rdb) else {
             return None;
         };
+        let mut filters: Vec<BloomFilter> = Vec::with_capacity(num_filters as usize);
```
KarthikSubbarao, line 74: Note: This will be fixed in a follow up PR. We want to make the vector expansion consistent for (1) bloom object creation and scaling (2) COPY (3) RDB Load (4) BF.LOAD.

We can also add the vector capacity into the DEBUG DIGEST. This will be very useful for correctness testing 

You can keep this in mind when working on the fix @zackcam

KarthikSubbarao (approved):

PR #24, opened 2024-11-22
Updating defragmentation to defrag both bloomfiltertype and bloomfiler structs

zackcam:
# Overview 
Updated the defragmentation callback. Now it defrags the inner BloomFilter by looping over the filters and calling defrag method on each filter. Afterwards we call the defrag on the BloomFilterType and update that value as well. 

In order to allow defragmentation on the inner BloomFilter from the BloomFilterType we needed to update the vec in type to hold a box of a BloomFilter this way we could get ownership of the BloomFilter object and update it as necessary. Sometimes due to the nature of rust the just unallocated addresses could be used by a following defrag call and reallocation 

# Example output of showing the addresses change when defragged

Command Ran: `bf.insert scale CAPACITY 10 ITEMS 1 2 3 4 5 6 7 8 9 11 12 13 14 15 16 17`
``` Before Address: 0x7fab07c510e0 
DefragResult Address: 0x7fab07c56780
new_filter Address: 0x7fab07c56780

Before Address: 0x7fab07c511c0
DefragResult Address: 0x7fab07c510e0
new_filter Address: 0x7fab07c510e0

Before Address of Top level. 0x7fab07c1a618
After Address of Top level. 0x7fab07c1afa0 


Before Address: 0x7fab07c56780
DefragResult Address: 0x7fab07c511c0
new_filter Address: 0x7fab07c511c0

Before Address: 0x7fab07c510e0
DefragResult Address: 0x7fab07c56780
new_filter Address: 0x7fab07c56780

Before Address of Top level. 0x7fab07c1afa0
After Address of Top level. 0x7fab07c1a618 


Before Address: 0x7fab07c511c0
DefragResult Address: 0x7fab07c510e0
new_filter Address: 0x7fab07c510e0

Before Address: 0x7fab07c56780
DefragResult Address: 0x7fab07c511c0
new_filter Address: 0x7fab07c511c0

Before Address of Top level. 0x7fab07c1a618
After Address of Top level. 0x7fab07c1afa0

### CODE REVIEW

**Comment by zackcam** [comment]
Added Defragging for Bloom as well. This involved also setting it in BloomFilter to be inside a box in order to gain ownership

# Example output of showing the addresses of bloom change when defragged
```
Before bloom filter 1 Address: 0x7f8393249540
After bloom filter 1 Address: 0x7f83932493c0

Before bloom filter 2 Address: 0x7f8393249300
After bloom filter 2 Address: 0x7f8393249540

Before bloom filter 1 Address: 0x7f83932493c0
After bloom filter 1 Address: 0x7f8393249300

Before bloom filter 2 Address: 0x7f8393249540
After bloom filter 2 Address: 0x7f83932493c0

src/wrapper/bloom_callback.rs
```diff
@@ -135,15 +136,35 @@ pub unsafe extern "C" fn bloom_defrag(
...
-        return 0;
+    let bloom_filter_type: &mut BloomFilterType = &mut *(*value).cast::<BloomFilterType>();
+
+    let num_filts = bloom_filter_type.filters.len();
+
+    for _ in 0..num_filts {
+        let bloom_filter_box = bloom_filter_type.filters.remove(0);
+        let bloom_filter = Box::into_raw(bloom_filter_box);
+        let defrag_result = unsafe {
+            raw::RedisModule_DefragAlloc.unwrap()(
```
KarthikSubbarao: For all the places we use this API, we should handle the case where it returns `null` re-using the same allocation / structure

You can use `is_null()` on a pointer that you have

https://doc.rust-lang.org/std/ptr/fn.null.html

src/wrapper/bloom_callback.rs
```diff
@@ -135,15 +136,35 @@ pub unsafe extern "C" fn bloom_defrag(
...
+                core::ptr::null_mut(),
+                (bloom_filter as *const BloomFilter as *mut BloomFilter) as *mut c_void,
+            )
+        };
+        let mut defragged_filter = Box::from_raw(defrag_result as *mut BloomFilter);
+
+        let inner_bloom = mem::replace(
+            &mut defragged_filter.bloom,
+            Box::new(bloomfilter::Bloom::new(1, 1)),
```
KarthikSubbarao: We want to avoid additional allocations and frees from the defrag code path.

Is there an alternative approach to get ownership of the inner bloom? any other `std::mem` function that can be used?

Would an `Option` on the inner bloom work instead of a `Box`? This allows us to use `take` and obtain ownership safely

src/wrapper/bloom_callback.rs
```diff
@@ -135,15 +136,35 @@ pub unsafe extern "C" fn bloom_defrag(
...
-    let curr_item = &*(*value).cast::<BloomFilterType>();
-    if curr_item.memory_usage()
-        > configs::BLOOM_MEMORY_LIMIT_PER_FILTER.load(Ordering::Relaxed) as usize
-    {
-        return 0;
+    let bloom_filter_type: &mut BloomFilterType = &mut *(*value).cast::<BloomFilterType>();
+
+    let num_filts = bloom_filter_type.filters.len();
+
+    for _ in 0..num_filts {
```
KarthikSubbarao: We can add Defrag cursors to iterate and maintain the index. It will be more optimal to perform defragmentation in iterations rather than locking up the server for the entire duration of defrag on a bloom object (which can consist of multiple bloom filters)

tests/test_bloom_metrics.py
```diff
@@ -4,7 +4,7 @@
 from valkeytests.conftest import resource_port_tracker
 from util.waiters import *
 
-DEFAULT_BLOOM_FILTER_SIZE = 179960
```
KarthikSubbarao, line 7: Can we add a python integration test for the defrag scenario?

src/wrapper/bloom_callback.rs
```diff
@@ -135,15 +136,35 @@ pub unsafe extern "C" fn bloom_defrag(
     _from_key: *mut RedisModuleString,
     value: *mut *mut c_void,
 ) -> i32 {
-    let curr_item = &*(*value).cast::<BloomFilterType>();
-    if curr_item.memory_usage()
-        > configs::BLOOM_MEMORY_LIMIT_PER_FILTER.load(Ordering::Relaxed) as usize
-    {
-        return 0;
+    let bloom_filter_type: &mut BloomFilterType = &mut *(*value).cast::<BloomFilterType>();
```
KarthikSubbarao, line 232: Can we add a new bool config to decide whether bloom objects should be defragmented?

It can be enabled by default - but should be dynamic so that we allow disabling at any time

src/wrapper/bloom_callback.rs
KarthikSubbarao: Other option is using a LazyStatic that contains a mutex of an Option of a Box of a Bloom :D 

```
lazy_static! {
    static ref DEFRAG_BLOOM_FILTER: Mutex<Option<Box<Bloom<[u8]>>>> = Mutex::new(Some(Box::new(Bloom::<[u8]>::new(1, 1))));
}
```

With this, we can re-use the same allocation for every bloom operation - this is not allocated / freed and instead we swap it back and forth to help obtain ownership.

src/bloom/utils.rs
```diff
@@ -377,7 +378,9 @@ impl BloomFilter {
     }
 
     pub fn number_of_bytes(&self) -> usize {
-        std::mem::size_of::<BloomFilter>() + (self.bloom.number_of_bits() / 8) as usize
+        std::mem::size_of::<BloomFilter>()
```
KarthikSubbarao, line 455: Let's fix the memory usage fn for the BloomFilterType to correctly report the bytes used by including the vector capacity * size of Box<BloomFilter>

src/wrapper/bloom_callback.rs
KarthikSubbarao: Additionally, we can rely on the ValkeyModule_ShouldStopDefrag to know when to yield back to the server for it to process other operations. We can use this to define one iteration and set the cursors accordingly.

src/wrapper/bloom_callback.rs
```diff
@@ -135,15 +136,35 @@ pub unsafe extern "C" fn bloom_defrag(
...
+
+    let num_filts = bloom_filter_type.filters.len();
```
KarthikSubbarao: Let's defrag the Vec (filters) as well

Cargo.toml
```diff
@@ -15,8 +15,9 @@ homepage = "https://github.com/valkey-io/valkey-bloom"
 valkey-module = "0.1.2"
 valkey-module-macros = "0"
 linkme = "0"
-bloomfilter = { version = "1.0.13", features = ["serde"] }
+bloomfilter = { path = "../rust-bloom-filter", features = ["serde"] }
```
parthpatel: Remove this before checking in.

src/bloom/data_type.rs
```diff
@@ -112,7 +112,7 @@ impl ValkeyDataType for BloomFilterType {
                 num_items as u32,
                 capacity as u32,
             );
-            filters.push(filter);
+            filters.push(Box::new(filter));
```
parthpatel, line 118: Is this to avoid re-assigning the ownership of filter variable? I don't understand the reason for it.

src/configs.rs
```diff
@@ -14,6 +14,7 @@ pub const BLOOM_FP_RATE_DEFAULT: f64 = 0.001;
 pub const BLOOM_FP_RATE_MIN: f64 = 0.0;
 pub const BLOOM_FP_RATE_MAX: f64 = 1.0;
 
+pub const BLOOM_DEFRAG_DEAFULT: bool = true;
```
parthpatel, line 19: How will a user/operator change this value?

src/wrapper/bloom_callback.rs
```diff
@@ -128,22 +139,113 @@ pub unsafe extern "C" fn bloom_free_effort(
     curr_item.free_effort()
 }
 
+lazy_static! {
+    static ref DEFRAG_BLOOM_FILTER: Mutex<Option<Box<Bloom<[u8]>>>> =
```
parthpatel, line 155: Can you document the intention here? I am not able to reverse engineer it from code.

parthpatel: Why ref?

src/bloom/data_type.rs
zackcam: If we put the filter in the box this allows us to take it out later and have ownership which means we can update the pointers address to actually change it. Without putting it into the box we couldn't actually change the addresses associated with it. [Documentation for box](https://doc.rust-lang.org/std/boxed/struct.Box.html)

Cargo.toml
zackcam: Removed was used as a temporary way to get the bloomfilter to provide a callback. The new version has been released so have updated the version to 3.0

src/configs.rs
zackcam: The user can add this to the start up arguments: 
--bf.bloom-defrag-enabled no
This will skip all the code related to bloom defragmenting and instantly just return 0

src/wrapper/bloom_callback.rs
zackcam: This is used as a safe way to take out the Bloom object without dropping it. This is a temporary bloom object which we put in while we attempt to defrag the original Bloom object. The ref is used to show it is a reference to the temp bloom object instead of the data of the temp object

tests/test_save_and_restore.py
```diff
@@ -14,12 +15,11 @@ def test_basic_save_and_restore(self):
         bf_info_result_1 = client.execute_command('BF.INFO testSave')
         assert(len(bf_info_result_1)) != 0
         curr_item_count_1 = client.info_obj().num_keys()
-        
+        # Keep the server running for 1 second more to have a larger uptime.
+        time.sleep(1)
```
KarthikSubbarao: We can remove this

tests/test_save_and_restore.py
```diff
@@ -14,12 +15,11 @@ def test_basic_save_and_restore(self):
...
         # save rdb, restart sever
         client.bgsave()
         self.server.wait_for_save_done()
-        # Keep the server running for 1 second more to have a larger uptime.
-        time.sleep(1)
         uptime_in_sec_1 = self.client.info_obj().uptime_in_secs()
         self.server.restart(remove_rdb=False, remove_nodes_conf=False, connect_client=True)
         uptime_in_sec_2 = self.client.info_obj().uptime_in_secs()
```
KarthikSubbarao: We can remove the checks below.

Just do the following
```
assert self.server.is_alive()
wait_for_equal(lambda: self.server.is_rdb_done_loading(), True)
# TODO: Check for message in engine logs - "RDB Loading ....." - You can add the right message here
```

tests/test_save_and_restore.py
```diff
@@ -35,6 +35,33 @@ def test_basic_save_and_restore(self):
...
+        assert curr_item_count_1 == count
+        # save rdb, restart sever
+        time.sleep(10)
+        client.bgsave()
+        self.server.wait_for_save_done()
+
+        self.server.restart(remove_rdb=False, remove_nodes_conf=False, connect_client=True)
+        assert self.server.is_alive()
+        wait_for_equal(lambda: self.server.is_rdb_done_loading(), True)
+
```
KarthikSubbarao, line 65: Can we add a check?
```
# TODO: Check for message in engine logs - "RDB Loading ....." - You can add the right message here
```

src/wrapper/defrag.rs
```diff
@@ -0,0 +1,41 @@
+use std::os::raw::{c_ulong, c_void};
+
+use valkey_module::raw;
+
+pub struct Defrag {
```
KarthikSubbarao, line 4: Not a blocker. When you have time, let's contribute this to valkeymodule-rs. Once it is released, we can use it from there

src/wrapper/bloom_callback.rs
```diff
@@ -128,22 +136,166 @@ pub unsafe extern "C" fn bloom_free_effort(
...
+        // Convert the inner_bloom into the correct type and then try to defragment it
+        let inner_bloom_ptr = Box::into_raw(inner_bloom);
+        let defragged_inner_bloom = defrag.alloc(inner_bloom_ptr as *mut c_void);
+        // Defragment the Vec within the Bloom object using the external callback
+        if !defragged_inner_bloom.is_null() {
+            let mut inner_bloom =
+                unsafe { Box::from_raw(defragged_inner_bloom as *mut bloomfilter::Bloom<[u8]>) };
+            let external_bloom =
+                inner_bloom.realloc_large_heap_allocated_objects(external_vec_defrag);
+            let old_bloom = mem::replace(&mut defragged_filter.bloom, Box::new(external_bloom));
```
KarthikSubbarao: `let placeholder_bloom = ....`

src/wrapper/bloom_callback.rs
```diff
@@ -128,22 +136,166 @@ pub unsafe extern "C" fn bloom_free_effort(
...
+            *temporary_bloom = Some(old_bloom); // Reset the original static
+        } else {
+            let mut inner_bloom = unsafe { Box::from_raw(inner_bloom_ptr) };
+            let external_bloom =
+                inner_bloom.realloc_large_heap_allocated_objects(external_vec_defrag);
+            let old_bloom = mem::replace(&mut defragged_filter.bloom, Box::new(external_bloom));
```
KarthikSubbarao: `let placeholder_bloom = ....`

src/wrapper/bloom_callback.rs
```diff
@@ -128,22 +136,166 @@ pub unsafe extern "C" fn bloom_free_effort(
...
+        let defrag_result = defrag.alloc(bloom_filter as *mut c_void);
+        let mut defragged_filter = {
+            if !defrag_result.is_null() {
+                Box::from_raw(defrag_result as *mut BloomFilter)
+            } else {
+                Box::from_raw(bloom_filter)
+            }
+        };
+        // Swap the Bloom object with a temporary one for defragmentation
+        let mut temporary_bloom = DEFRAG_BLOOM_FILTER.lock().unwrap();
```
KarthikSubbarao, line 237: `let mut placeholder_bloom = ....`

src/wrapper/bloom_callback.rs
```diff
@@ -128,22 +136,166 @@ pub unsafe extern "C" fn bloom_free_effort(
...
+    // Defragment the Vec of filters itself
+    let filters_vec = mem::take(&mut bloom_filter_type.filters);
+    let filters_ptr = Box::into_raw(filters_vec.into_boxed_slice()) as *mut c_void;
+    let defragged_filters_ptr = defrag.alloc(filters_ptr);
+    if !defragged_filters_ptr.is_null() {
+        bloom_filter_type.filters = unsafe {
+            Vec::from_raw_parts(
+                defragged_filters_ptr as *mut Box<BloomFilter>,
+                num_filters,
+                num_filters,
```
KarthikSubbarao: Are both supposed to be `num_filters`? Should one be capacity?

src/wrapper/bloom_callback.rs
```diff
@@ -128,22 +136,166 @@ pub unsafe extern "C" fn bloom_free_effort(
...
+/// * `defrag_ctx` - A raw pointer to the defragmentation context.
+/// * `_from_key` - A raw pointer to the Redis module string (unused in this function).
+/// * `value` - A mutable raw pointer to a raw pointer representing the BloomFilterType to be defragmented.
+///
+/// # Returns
+///
+/// Returns an `i32` where:
+/// * 0 indicates successful complete defragmentation.
+/// * 1 indicates incomplete defragmentation (not all filters were defragged).
 pub unsafe extern "C" fn bloom_defrag(
```
KarthikSubbarao, line 217: This function is looking good functionally (and from testing). Thank you!

Next, we can think about monitoring:

For monitoring, we can consider adding hit and miss metrics for the different layers of allocations

tests/test_bloom_defrag.py
```diff
@@ -0,0 +1,177 @@
+import time
+from valkeytests.valkey_test_case import ValkeyAction
+from valkey_bloom_test_case import ValkeyBloomTestCaseBase
+from valkeytests.conftest import resource_port_tracker
+from util.waiters import *
+import pytest
+
+@pytest.mark.skip_for_asan(reason="These tests are skipped due to not being able to set activedefrag to yes when valkey server is an ASAN build")
```
KarthikSubbarao, line 8: Nice!

tests/test_bloom_defrag.py
```diff
@@ -0,0 +1,177 @@
...
+class TestBloomDefrag(ValkeyBloomTestCaseBase):
```
KarthikSubbarao, line 9: I still need to review the python tests. Will do shortly

src/wrapper/bloom_callback.rs
zackcam: Will add the metrics for hits and misses in a seperate pr

src/wrapper/bloom_callback.rs
zackcam: Fixed and changed to capacity now

src/wrapper/bloom_callback.rs
zackcam: Updated all old_bloom to be placeholder_bloom

tests/test_save_and_restore.py
zackcam: Updated test to have two different checks for it loading and loaded from engine logs

tests/test_save_and_restore.py
zackcam: Updated test to have two different checks for it loading and loaded from engine logs

src/bloom/data_type.rs
```diff
@@ -71,8 +71,11 @@ impl ValkeyDataType for BloomFilterType {
         let Ok(fp_rate) = raw::load_double(rdb) else {
             return None;
         };
-        let mut filters: Vec<BloomFilter> = Vec::with_capacity(num_filters as usize);
-
+        let mut filters = if num_filters == 1 {
```
KarthikSubbarao: Let's add a comment to explain that this is to keep a consistent vector expansion WRT to how we scale out bloom objects (as a result of BF commands)

src/bloom/utils.rs
```diff
@@ -87,14 +86,15 @@ impl BloomFilterType {
...
 
         // Create the bloom filter and add to the main BloomFilter object.
-        let bloom = BloomFilter::new(fp_rate, capacity);
+        let bloom: Box<BloomFilter> = Box::new(BloomFilter::new(fp_rate, capacity));
         let filters = vec![bloom];
+        metrics::BLOOM_OBJECT_TOTAL_MEMORY_BYTES.fetch_add(
+            mem::size_of::<BloomFilterType>()
+                + (filters.capacity() * std::mem::size_of::<Box<BloomFilter>>()),
+            std::sync::atomic::Ordering::Relaxed,
+        );
```
KarthikSubbarao: If this is common / exactly the same as RDB load, we can move it into a common fn and use it in both places

src/bloom/utils.rs
```diff
@@ -105,16 +105,17 @@ impl BloomFilterType {
...
         for filter in &from_bf.filters {
-            let new_filter = BloomFilter::create_copy_from(filter);
+            let new_filter = Box::new(BloomFilter::create_copy_from(filter));
             filters.push(new_filter);
         }
+        metrics::BLOOM_OBJECT_TOTAL_MEMORY_BYTES.fetch_add(
+            mem::size_of::<BloomFilterType>()
+                + (filters.capacity() * std::mem::size_of::<Box<BloomFilter>>()),
+            std::sync::atomic::Ordering::Relaxed,
+        );
```
KarthikSubbarao: If this is common / exactly the same as RDB load, we can move it into a common fn and use it in both places

src/bloom/utils.rs
```diff
@@ -290,7 +302,8 @@ impl BloomFilterType {
                 // add bloom filter type metrics.
                 metrics::BLOOM_NUM_OBJECTS.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
                 metrics::BLOOM_OBJECT_TOTAL_MEMORY_BYTES.fetch_add(
-                    mem::size_of::<BloomFilterType>(),
+                    mem::size_of::<BloomFilterType>()
+                        + (item.filters.capacity() * std::mem::size_of::<Box<BloomFilter>>()),
```
KarthikSubbarao: Same comment

zackcam: fixed

src/bloom/data_type.rs
zackcam: Added

tests/test_save_and_restore.py
```diff
@@ -43,6 +42,35 @@ def test_basic_save_and_restore(self):
...
+        for i in range(0, count):
+            name = str(i) + "key"
+
+            bf_add_result_1 = client.execute_command('BF.ADD ' + name + ' item')
+            assert bf_add_result_1 == 1
+
+        curr_item_count_1 = client.info_obj().num_keys()
+        assert curr_item_count_1 == count
+        # save rdb, restart sever
+        time.sleep(10)
```
KarthikSubbarao: We can remove this

tests/test_save_and_restore.py
```diff
@@ -43,6 +42,35 @@ def test_basic_save_and_restore(self):
...
+        client.bgsave()
+        self.server.wait_for_save_done()
+
+        self.server.restart(remove_rdb=False, remove_nodes_conf=False, connect_client=True)
+        assert self.server.is_alive()
+        self.server.verify_string_in_logfile("Loading RDB produced by Valkey")
```
KarthikSubbarao: Maybe move this below to allow some time for the server to start up
```
wait_for_equal(lambda: self.server.is_rdb_done_loading(), True)
```

tests/test_bloom_metrics.py
```diff
@@ -132,8 +132,4 @@ def test_save_and_restore_metrics(self):
 
         # Compare original and loaded scaled bloomfilter infos
         new_client = self.server.get_new_client()
-        restored_info_obj = new_client.execute_command('BF.INFO key1')
-        for i in range(1, len(original_info_obj), 2):
-            assert original_info_obj[i] == restored_info_obj[i]
```
KarthikSubbarao: Let's add this back to validate correctness across both APIs

tests/test_bloom_defrag.py
```diff
@@ -0,0 +1,176 @@
...
+            num_items = num_items_inseted_per_object[index]
+            items = list(range(1, num_items + 1))
+            
+            command = f'bf.mexists {scale} ' + ' '.join(map(str, items))
+            results = self.client.execute_command(command)
+            
+            expected_results = [1] * num_items
+            assert results == expected_results, f"Unexpected results for scale {scale}: {results}"
+
+    def wait_for_defrag(self, initial_hits, initial_misses):
```
KarthikSubbarao: We can move this into valkeytests and make it accept a duration

tests/test_bloom_defrag.py
```diff
@@ -0,0 +1,176 @@
...
+        max_wait_time = 20
+        start_time = time.time()
+        while time.time() - start_time < max_wait_time:
+            time.sleep(5)
```
KarthikSubbarao: Let's check if `current_active_defrag_time` is closely proportional (almost 1:1) to time.sleep(5), if so, let's use this instead to ensure we ran defrag for X seconds

tests/test_bloom_defrag.py
```diff
@@ -0,0 +1,176 @@
...
+        # Initial stats
+        stats = self.parse_valkey_info("STATS")
+        initial_defrag_hits = int(stats.get('active_defrag_hits', 0))
+        initial_defrag_misses = int(stats.get('active_defrag_misses', 0))
+
+
+        # Create list of key names we will operate with
+        scale_names = [f'scale_{i}' for i in range(1, 2000)]
+        # A list of the number of items we inserted into each Bloom object
+        num_items_inseted_per_object = []
```
KarthikSubbarao: `num_items_inserted_per_object `

tests/test_bloom_defrag.py
```diff
@@ -0,0 +1,176 @@
...
+        first_defrag_stats = self.parse_valkey_info("STATS")
+        first_defrag_hits = int(first_defrag_stats.get('active_defrag_hits', 0))
+        first_defrag_misses = int(first_defrag_stats.get('active_defrag_misses', 0))
+
+        # Assertion we got hits and misses when defragging
+        assert first_defrag_hits > initial_defrag_hits and first_defrag_misses > initial_defrag_misses
+        assert float(memory_info_after_defrag.get('allocator_frag_ratio', 0)) < float(memory_info_non_defragged.get('allocator_frag_ratio', 0))
+        # Check that items we added still exist in the respective bloom objects
+        for index, scale in enumerate(scale_names[1::2]):
+            num_items = num_items_inseted_per_object[index]
```
KarthikSubbarao: Let's add a comment to say that this is the range of items from 0 to X which should all indicate that they exist. Also, we can clarify this during item population above

tests/test_bloom_defrag.py
```diff
@@ -0,0 +1,176 @@
...
+        final_defrag_hits = int(final_stats.get('active_defrag_hits', 0))
+        final_defrag_misses = int(final_stats.get('active_defrag_misses', 0))
+        assert  final_defrag_hits > initial_defrag_hits or final_defrag_misses > initial_defrag_misses, "No defragmentation occurred after RDB load"
+        # Check that items we added still exist in the respective bloom objects
+        for index, scale in enumerate(scale_names[1::2]):
+            num_items = num_items_inseted_per_object[index]
+            items = list(range(1, num_items + 1))
+            
+            command = f'bf.mexists {scale} ' + ' '.join(map(str, items))
+            results = self.client.execute_command(command)
```
KarthikSubbarao: We can use a common function

src/wrapper/defrag.rs
```diff
@@ -0,0 +1,41 @@
...
+    ///
+    /// This function is temporary and will be removed once implemented in valkeymodule-rs .
+    pub unsafe fn alloc(&self, ptr: *mut c_void) -> *mut c_void {
+        unsafe { raw::RedisModule_DefragAlloc.unwrap()(self.defrag_ctx, ptr) }
+    }
+
+    /// # Safety
+    ///
+    /// This function sis temporary and will be removed once implemented in valkeymodule-rs .
+    pub unsafe fn curserset(&self, cursor: u64) -> i32 {
```
KarthikSubbarao: naming: `cursorset`

src/wrapper/defrag.rs
```diff
@@ -0,0 +1,41 @@
...
+        unsafe { raw::RedisModule_DefragCursorSet.unwrap()(self.defrag_ctx, cursor) }
+    }
+
+    /// # Safety
+    ///
+    /// This function sis temporary and will be removed once implemented in valkeymodule-rs .
+    pub unsafe fn curserget(&self, cursor: *mut u64) -> i32 {
```
KarthikSubbarao: `cursorget`

src/wrapper/defrag.rs
```diff
@@ -0,0 +1,41 @@
...
+        unsafe { raw::RedisModule_DefragCursorGet.unwrap()(self.defrag_ctx, cursor) }
+    }
+
+    /// # Safety
+    ///
+    /// This function sis temporary and will be removed once implemented in valkeymodule-rs .
+    pub unsafe fn should_stop_defrag(&self) -> i32 {
```
KarthikSubbarao: Can this be a boolean type?

src/wrapper/bloom_callback.rs
```diff
@@ -142,22 +149,169 @@ pub unsafe extern "C" fn bloom_free_effort(
...
+    defrag.curserget(&mut cursor);
+
+    // Convert pointer to BloomFilterType so we can operate on it.
+    let bloom_filter_type: &mut BloomFilterType = &mut *(*value).cast::<BloomFilterType>();
+
+    let num_filters = bloom_filter_type.filters.len();
+    let filters_capacity = bloom_filter_type.filters.capacity();
+
+    // While we are within a timeframe decided from should_stop_defrag and not over the number of filters defrag the next filter
+    while defrag.should_stop_defrag() == 0 && cursor < num_filters.try_into().unwrap() {
```
KarthikSubbarao: For here and in 2 lines below
```
num_filters.try_into().unwrap(
```

If you are converting to larger/equal type, you can just use `as`

src/wrapper/bloom_callback.rs
```diff
@@ -142,22 +149,169 @@ pub unsafe extern "C" fn bloom_free_effort(
...
+        let defrag_result = defrag.alloc(bloom_filter as *mut c_void);
+        let mut defragged_filter = {
+            if !defrag_result.is_null() {
+                Box::from_raw(defrag_result as *mut BloomFilter)
+            } else {
+                Box::from_raw(bloom_filter)
+            }
+        };
+        // Swap the Bloom object with a temporary one for defragmentation
+        let mut temporary_bloom = DEFRAG_BLOOM_FILTER.lock().unwrap();
```
KarthikSubbarao: `expect`

src/wrapper/bloom_callback.rs
```diff
@@ -142,22 +149,169 @@ pub unsafe extern "C" fn bloom_free_effort(
...
+        // Reinsert the defragmented filter and increment the cursor
+        bloom_filter_type
+            .filters
+            .insert(cursor.try_into().unwrap(), defragged_filter);
+        cursor += 1;
+    }
+    // Save the cursor for where we will start defragmenting from next time
+    defrag.curserset(cursor);
+    // If not all filters were looked at, return 1 to indicate incomplete defragmentation
+    if cursor < (num_filters).try_into().unwrap() {
```
KarthikSubbarao: Use `as`

src/bloom/data_type.rs
```diff
@@ -71,11 +71,16 @@ impl ValkeyDataType for BloomFilterType {
...
         let Ok(is_seed_random_u64) = raw::load_unsigned(rdb) else {
             return None;
         };
         let is_seed_random = is_seed_random_u64 == 1;
+        // This is to keep consistent vector expansion with how we scale bloom objects.
+        let mut filters = if num_filters == 1 {
+            Vec::with_capacity(1)
+        } else {
+            Vec::new()
+        };
```
KarthikSubbarao: We can always use ` Vec::with_capacity(1)` and have it consistent with how we create objects

src/bloom/utils.rs
```diff
@@ -217,18 +216,35 @@ impl BloomFilterType {
...
+            );
             metrics::BLOOM_NUM_ITEMS_ACROSS_OBJECTS
                 .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
             return Ok(1);
         }
         Ok(0)
     }
 
+    /// Calculates the memory usage of a BloomFilterType object
+    fn calculate_bloom_filter_type_memory_usage(&self) -> usize {
```
KarthikSubbarao: `bloom_filter_type_memory_usage`

src/bloom/data_type.rs
```diff
@@ -111,12 +116,14 @@ impl ValkeyDataType for BloomFilterType {
...
             }
-            filters.push(filter);
+            filters.push(Box::new(filter));
         }
         BLOOM_OBJECT_TOTAL_MEMORY_BYTES.fetch_add(
-            mem::size_of::<BloomFilterType>(),
+            mem::size_of::<BloomFilterType>()
+                + (filters.capacity() * std::mem::size_of::<Box<BloomFilter>>()),
             std::sync::atomic::Ordering::Relaxed,
         );
```
KarthikSubbarao: We can call `bloom_filter_type_incr_metrics_on_new_create` instead

src/bloom/utils.rs
```diff
@@ -127,7 +125,8 @@ impl BloomFilterType {
 
     /// Return the total memory usage of the BloomFilterType object.
     pub fn memory_usage(&self) -> usize {
-        let mut mem: usize = std::mem::size_of::<BloomFilterType>();
+        let mut mem: usize = std::mem::size_of::<BloomFilterType>()
+            + (self.filters.capacity() * std::mem::size_of::<Box<BloomFilter>>());
```
KarthikSubbarao: We can use `bloom_filter_type_memory_usage` for the initial `mem` variable assignment

src/bloom/utils.rs
```diff
@@ -127,7 +125,8 @@ impl BloomFilterType {
 
     /// Return the total memory usage of the BloomFilterType object.
```
KarthikSubbarao: `Return the total memory usage of the BloomFilterType object and every allocation it contains.`

src/bloom/utils.rs
```diff
@@ -250,6 +266,13 @@ impl BloomFilterType {
         }
     }
 
+    fn bloom_filter_type_incr_metrics_on_new_create(&self) {
```
KarthikSubbarao: Lets add documentation

tests/test_bloom_defrag.py
zackcam: I removed this method and made it a one liner using active_defrag_time. As active_defrag time was close to sleep we can just use a one liner instead

src/bloom/data_type.rs
```diff
@@ -79,6 +80,9 @@ impl ValkeyDataType for BloomFilterType {
             return None;
         };
         let is_seed_random = is_seed_random_u64 == 1;
+
+        let mut filters = Vec::with_capacity(1);
```
KarthikSubbarao, line 81: Can we add a short documentation explaining why we use this - `Vec::with_capacity(1)` here? I think I had made this comment earlier

src/bloom/data_type.rs
```diff
@@ -114,12 +118,9 @@ impl ValkeyDataType for BloomFilterType {
...
             }
-            filters.push(filter);
+            filters.push(Box::new(filter));
         }
-        BLOOM_OBJECT_TOTAL_MEMORY_BYTES.fetch_add(
-            mem::size_of::<BloomFilterType>(),
-            std::sync::atomic::Ordering::Relaxed,
-        );
+
         BLOOM_NUM_OBJECTS.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
```
KarthikSubbarao: Can we move this line into the function `bloom_filter_type_incr_metrics_on_new_create `? It is used in every caller site of the `bloom_filter_type_incr_metrics_on_new_create` and can help condense metrics code
```
BLOOM_NUM_OBJECTS.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
```

You can update the following caller sites:
* new_reserved
* load_from_rdb
* create_copy_from
* decode_bloom_filter

src/wrapper/defrag.rs
```diff
@@ -0,0 +1,41 @@
...
+    ///
+    /// This function is temporary and will be removed once implemented in valkeymodule-rs .
+    pub unsafe fn alloc(&self, ptr: *mut c_void) -> *mut c_void {
+        unsafe { raw::RedisModule_DefragAlloc.unwrap()(self.defrag_ctx, ptr) }
+    }
+
+    /// # Safety
+    ///
+    /// This function is temporary and will be removed once implemented in valkeymodule-rs .
+    pub unsafe fn cursorset(&self, cursor: u64) -> i32 {
```
KarthikSubbarao: 1. Naming: `set_cursor` might make this more readable / Rust like
2. Question: What is the return code here? What should the caller site do with this i32? if this is meant to be a void, can we remove the return code?

src/wrapper/defrag.rs
```diff
@@ -0,0 +1,41 @@
...
+        unsafe { raw::RedisModule_DefragCursorSet.unwrap()(self.defrag_ctx, cursor) }
+    }
+
+    /// # Safety
+    ///
+    /// This function is temporary and will be removed once implemented in valkeymodule-rs .
+    pub unsafe fn cursorget(&self, cursor: *mut u64) -> i32 {
```
KarthikSubbarao: 1. Naming: `get_cursor` might make this more readable / Rust like
2. Can this simply return an u64 instead of accepting a mutable pointer?  It will a simpler interface to use

tests/valkey_bloom_test_case.py
```diff
@@ -193,8 +193,20 @@ def verify_bloom_metrics(self, info_response, expected_memory, expected_num_obje
             elif line.startswith('bf_bloom_capacity_across_objects'):
                 sum_capacity = int(line.split(':')[1])
 
+        print(total_memory_bites)
+        print(expected_memory)
```
KarthikSubbarao: We can remove this

tests/valkey_bloom_test_case.py
```diff
@@ -193,8 +193,20 @@ def verify_bloom_metrics(self, info_response, expected_memory, expected_num_obje
...
         assert total_memory_bites == expected_memory 
         assert num_objects == expected_num_objects
         assert num_filters == expected_num_filters
         assert num_items == expected_num_items
         assert sum_capacity == expected_sum_capacity
+
+    def parse_valkey_info(self, section):
```
KarthikSubbarao, line 209: Let's add a comment for documentation here

src/wrapper/defrag.rs
KarthikSubbarao: The return code is useful. But we can update it from i32 to use https://docs.rs/valkey-module/latest/valkey_module/raw/enum.Status.html#variant.Ok to follow the convention of this SDK

src/wrapper/defrag.rs
KarthikSubbarao: Maybe have this return `Result<u64, ()>` or `Option<u64>`

KarthikSubbarao (approved):

PR #25, opened 2024-11-27
Support random seed per bloom object by default (configurable)

KarthikSubbarao:
Support random seed per bloom object by default (configurable) where all filters within the bloom object will use the same sip_keys (for their hash functions) as the first filter within the object which will be randomly generated.

When the new config (bf.bloom-use-random-seed) is disabled, new bloom object creates will use a fixed seed. In this case, the RDB data for that particular object will be slightly more compact as it does not contain the sip_keys - 32 bytes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ homepage = "https://github.com/valkey-io/valkey-bloom"
 valkey-module = "0.1.2"
 valkey-module-macros = "0"
 linkme = "0"
-bloomfilter = { version = "1.0.13", features = ["serde"] }
+bloomfilter = { version = "3.0.1", features = ["serde"] }
 lazy_static = "1.4.0"
 libc = "0.2"
 serde = { version = "1.0", features = ["derive"] }
--- a/src/bloom/utils.rs
+++ b/src/bloom/utils.rs
@@ -5,10 +5,11 @@ use crate::{
     metrics,
 };
 use bloomfilter;
+use bloomfilter::{deserialize, serialize};
 use serde::{Deserialize, Serialize};
-use std::{mem, sync::atomic::Ordering};
 
 use super::data_type::BLOOM_TYPE_VERSION;
+use std::{mem, sync::atomic::Ordering};
 
 /// KeySpace Notification Events
 pub const ADD_EVENT: &str = "bloom.add";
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -15,7 +15,7 @@ homepage = "https://github.com/valkey-io/valkey-bloom"
 valkey-module = "0.1.2"
 valkey-module-macros = "0"
 linkme = "0"
-bloomfilter = { version = "1.0.13", features = ["serde"] }
+bloomfilter = { version = "2.0.0", features = ["serde"] }
```
YueTang-Vanessa: Can you align with changes in https://github.com/valkey-io/valkey-bloom/pull/23/files?

src/bloom/utils.rs
```diff
@@ -324,24 +342,49 @@ pub struct BloomFilter {
...
 impl BloomFilter {
-    /// Instantiate empty BloomFilter object.
-    pub fn new(fp_rate: f64, capacity: u32) -> BloomFilter {
-        let bloom = bloomfilter::Bloom::new_for_fp_rate_with_seed(
-            capacity as usize,
-            fp_rate,
-            &configs::FIXED_SEED,
-        );
+    /// Instantiate empty BloomFilter object with a fixed seed used to create sip keys.
+    pub fn with_fixed_seed(fp_rate: f64, capacity: u32, seed: &[u8; 32]) -> BloomFilter {
```
YueTang-Vanessa: Can we rename it so that to show this function will create a new bloom filter? Like `new_with_fixed_seed`?

src/bloom/utils.rs
```diff
@@ -324,24 +342,49 @@ pub struct BloomFilter {
...
-        metrics::BLOOM_OBJECT_TOTAL_MEMORY_BYTES
-            .fetch_add(fltr.number_of_bytes(), std::sync::atomic::Ordering::Relaxed);
-        metrics::BLOOM_CAPACITY_ACROSS_OBJECTS
-            .fetch_add(capacity.into(), std::sync::atomic::Ordering::Relaxed);
+        fltr.incr_metrics_on_new_create();
+        fltr
+    }
+
+    /// Instantiate empty BloomFilter object with a randomly generated seed used to create sip keys.
+    pub fn with_random_seed(fp_rate: f64, capacity: u32) -> BloomFilter {
```
YueTang-Vanessa: Same as above, maybe `new_with_random_seed`?

src/bloom/utils.rs
```diff
@@ -324,24 +342,49 @@ pub struct BloomFilter {
...
+            bloom,
+            num_items: 0,
+            capacity,
+        };
+        fltr.incr_metrics_on_new_create();
+        fltr
+    }
+
+    /// Instantiate empty BloomFilter object using fixed sip keys.
+    pub fn with_fixed_sip_keys(fp_rate: f64, capacity: u32, keys: [(u64, u64); 2]) -> BloomFilter {
```
YueTang-Vanessa: Same as above.

PR #26, opened 2024-11-28
Add tightening_ratio support

nnmehta:
Add tightening_ratio support for tuning of the `TIGHENTING_RATIO` to allow scalable bloom objects to have a higher capacity at cost of higher false positive rate.

We allow the `TIGHTENING_RATIO` argument only if the command is replicated.

### DIFF
--- a/src/bloom/command_handler.rs
+++ b/src/bloom/command_handler.rs
@@ -4,7 +4,7 @@ use crate::bloom::utils::BloomFilterType;
 use crate::configs;
 use crate::configs::{
     BLOOM_CAPACITY_MAX, BLOOM_CAPACITY_MIN, BLOOM_EXPANSION_MAX, BLOOM_EXPANSION_MIN,
-    BLOOM_FP_RATE_MAX, BLOOM_FP_RATE_MIN,
+    BLOOM_FP_RATE_MAX, BLOOM_FP_RATE_MIN, TIGHTENING_RATIO_MAX, TIGHTENING_RATIO_MIN,
 };
 use std::sync::atomic::Ordering;
 use valkey_module::ContextFlags;
@@ -110,11 +110,13 @@ pub fn bloom_filter_add_value(
         None => {
             // Instantiate empty bloom filter.
             let fp_rate = configs::BLOOM_FP_RATE_DEFAULT;
+            let tightening_ratio = configs::TIGHTENING_RATIO;
             let capacity = configs::BLOOM_CAPACITY.load(Ordering::Relaxed) as u32;
             let expansion = configs::BLOOM_EXPANSION.load(Ordering::Relaxed) as u32;
             let use_random_seed = configs::BLOOM_USE_RANDOM_SEED.load(Ordering::Relaxed);
             let mut bloom = match BloomFilterType::new_reserved(
                 fp_rate,
+                tightening_ratio,
                 capacity,
                 expansion,
                 use_random_seed,
@@ -279,8 +281,10 @@ pub fn bloom_filter_reserve(ctx: &Context, input_args: &[ValkeyString]) -> Valke
             let use_random_seed = configs::BLOOM_USE_RANDOM_SEED.load(Ordering::Relaxed);
             // Skip bloom filter size validation on replicated cmds.
             let validate_size_limit = !ctx.get_flags().contains(ContextFlags::REPLICATED);
+            let tightening_ratio = configs::TIGHTENING_RATIO;
             let bloom = match BloomFilterType::new_reserved(
                 fp_rate,
+                tightening_ratio,
                 capacity,
                 expansion,
                 use_random_seed,
@@ -302,6 +306,7 @@ pub fn bloom_filter_reserve(ctx: &Context, input_args: &[ValkeyString]) -> Valke
 
 pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
     let argc = input_args.len();
+    let replicated_cmd = ctx.get_flags().contains(ContextFlags::REPLICATED);
     // At the very least, we need: BF.INSERT <key> ITEMS <item>
     if argc < 4 {
         return Err(ValkeyError::WrongArity);
@@ -311,6 +316,7 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
     let filter_name = &input_args[idx];
     idx += 1;
     let mut fp_rate = configs::BLOOM_FP_RATE_DEFAULT;
+    let mut tightening_ratio = configs::TIGHTENING_RATIO;
     let mut capacity = configs::BLOOM_CAPACITY.load(Ordering::Relaxed) as u32;
     let mut expansion = configs::BLOOM_EXPANSION.load(Ordering::Relaxed) as u32;
     let mut nocreate = false;
@@ -328,6 +334,21 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
                     }
                     _ => {
                         return Err(ValkeyError::Str(utils::BAD_ERROR_RATE));
+                    }
+                };
+            }
+            "TIGHTENING" if replicated_cmd => {
+                if idx >= (argc - 1) {
+                    return Err(ValkeyError::WrongArity);
+                }
+                idx += 1;
+                tightening_ratio = match input_args[idx].to_string_lossy().parse::<f64>() {
+                    Ok(num) if num > TIGHTENING_RATIO_MIN && num < TIGHTENING_RATIO_MAX => num,
+                    Ok(num) if !(num > TIGHTENING_RATIO_MIN && num < TIGHTENING_RATIO_MAX) => {
+                        return Err(ValkeyError::Str(utils::ERROR_RATIO_RANGE));
+                    }
+                    _ => {
+                        return Err(ValkeyError::Str(utils::BAD_ERROR_RATIO));
                     }
                 };
             }
@@ -387,7 +408,6 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
         }
     };
     // Skip bloom filter size validation on replicated cmds.
-    let validate_size_limit = !ctx.get_flags().contains(ContextFlags::REPLICATED);
     let mut add_succeeded = false;
     match value {
         Some(bloom) => {
@@ -398,7 +418,7 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
                 bloom,
                 true,
                 &mut add_succeeded,
-                validate_size_limit,
+                !replicated_cmd,
             );
             replicate_and_notify_events(ctx, filter_name, add_succeeded, false);
             response
@@ -410,10 +430,11 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
             let use_random_seed = configs::BLOOM_USE_RANDOM_SEED.load(Ordering::Relaxed);
             let mut bloom = match BloomFilterType::new_reserved(
                 fp_rate,
+                tightening_ratio,
                 capacity,
                 expansion,
                 use_random_seed,
-                validate_size_limit,
+                !replicated_cmd,
             ) {
                 Ok(bf) => bf,
                 Err(err) => return Err(ValkeyError::Str(err.as_str())),
@@ -425,7 +446,7 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
                 &mut bloom,
                 true,
                 &mut add_succeeded,
-                validate_size_limit,
+                !replicated_cmd,
             );
             match filter_key.set_value(&BLOOM_FILTER_TYPE, bloom) {
                 Ok(()) => {
--- a/src/bloom/data_type.rs
+++ b/src/bloom/data_type.rs
@@ -71,6 +71,9 @@ impl ValkeyDataType for BloomFilterType {
         let Ok(fp_rate) = raw::load_double(rdb) else {
             return None;
         };
+        let Ok(tightening_ratio) = raw::load_double(rdb) else {
+            return None;
+        };
         let mut filters: Vec<BloomFilter> = Vec::with_capacity(num_filters as usize);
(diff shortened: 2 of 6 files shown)


KarthikSubbarao:
Can this PR be rebased? The digest change is merged in and looks like this PR contains commits from digest (which should be removed after rebasing)

src/bloom/command_handler.rs
```diff
@@ -228,6 +230,16 @@ pub fn bloom_filter_reserve(ctx: &Context, input_args: &[ValkeyString]) -> Valke
...
+    let tightening_ratio = match input_args[curr_cmd_idx].to_string_lossy().parse::<f64>() {
+        Ok(num) if num > TIGHTENING_RATIO_MIN && num < TIGHTENING_RATIO_MAX => num,
+        Ok(num) if !(num > TIGHTENING_RATIO_MIN && num < TIGHTENING_RATIO_MAX) => {
+            return Err(ValkeyError::Str(utils::ERROR_RATIO_RANGE));
+        }
+        _ => {
+            return Err(ValkeyError::Str(utils::BAD_ERROR_RATIO));
+        }
+    };
+    curr_cmd_idx += 1;
```
KarthikSubbarao: This will not be provided as a command argument for the BF.RESERVE command. We can remove this

src/bloom/command_handler.rs
```diff
@@ -327,6 +341,18 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
                     }
                 };
             }
+            "RATIO" => {
```
KarthikSubbarao: 1. Renaming to `TIGHTENING` makes this clearer
2. Let's add a check to only allow this argument if the command is replicated. If it is not replicated, we should not accept it as command args. This means, if the command is on a primary, it will use the constant from the configs.rs  file

src/bloom/command_handler.rs
```diff
@@ -228,6 +230,16 @@ pub fn bloom_filter_reserve(ctx: &Context, input_args: &[ValkeyString]) -> Valke
         }
```
KarthikSubbarao: We will need to update the arity check at the beginning of the fn to handle the extra TIGHTENING argument

src/configs.rs
```diff
@@ -32,6 +32,8 @@ lazy_static! {
 // Tightening ratio used during scale out for the calculation of fp_rate of every new filter within a bloom object to
 // maintain the bloom object's overall fp_rate to the configured value.
 pub const TIGHTENING_RATIO: f64 = 0.5;
+pub const TIGHTENING_RATIO_MIN: f64 = 0.001;
```
KarthikSubbarao: This should be 0, right?

src/wrapper/bloom_callback.rs
```diff
@@ -2,6 +2,7 @@ use crate::bloom;
 use crate::bloom::data_type::ValkeyDataType;
 use crate::bloom::utils::BloomFilterType;
 use crate::configs;
+use crate::wrapper::digest::Digest;
```
KarthikSubbarao: Since we are updating RDB Load, we need to update RDB Save to add the tightening ratio at the right place.

src/bloom/data_type.rs
```diff
@@ -122,10 +127,28 @@ impl ValkeyDataType for BloomFilterType {
...
             filters,
         };
         Some(item)
     }
+
+    /// Function that is used to generate a digest on the Bloom Object.
+    fn debug_digest(&self, mut dig: Digest) {
+        dig.add_long_long(self.expansion.into());
+        dig.add_string_buffer(&self.fp_rate.to_le_bytes());
+        // dig.add_string_buffer(&self.tightening_ratio.to_le_bytes());
```
KarthikSubbarao: We should add the tightening ratio to the digest. Why is this commented out?

src/bloom/command_handler.rs
nnmehta: Sure will remove this

src/configs.rs
nnmehta: Will update

src/wrapper/bloom_callback.rs
nnmehta: Will update [here](https://github.com/valkey-io/valkey-bloom/blob/7c254680374a22794a7676da5797a9f84b2e9087/src/wrapper/bloom_callback.rs#L23) in `rdb_save`

src/bloom/data_type.rs
nnmehta: I was testing out something else and commented this briefly, will uncomment in future revision

src/bloom/command_handler.rs
nnmehta: Will update this

src/bloom/command_handler.rs
```diff
@@ -299,14 +298,15 @@ pub fn bloom_filter_reserve(ctx: &Context, input_args: &[ValkeyString]) -> Valke
 pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
     let argc = input_args.len();
     // At the very least, we need: BF.INSERT <key> ITEMS <item>
-    if argc < 4 {
+    if argc < 5 {
```
KarthikSubbarao: This depends on `ctx.get_flags().contains(ContextFlags::REPLICATED)`, right?

If replica, < 5. If primary, < 4. Can we update it?

src/bloom/utils.rs
```diff
@@ -94,9 +98,11 @@ impl BloomFilterType {
         // Create the bloom filter and add to the main BloomFilter object.
         let bloom = BloomFilter::new(fp_rate, capacity);
         let filters = vec![bloom];
+        let tightening_ratio = 0.5;
```
KarthikSubbarao: This will need to be passed in as arguments from `new_reserved`. This should not be hard coded

nnmehta: I had added it in this [change](https://github.com/nnmehta/ValkeyBloom/commit/cc18784ba7706352bb74c918128057feeb740bdd#diff-706cf6a8f58683c30caf89a9e9b6db3af4a4ed8df5c2af8cfe1bc8d6ce56d74aR118) but since you mentioned above: 
```
 not be provided as a command argument for the BF.RESERVE command
```

I had to remove it wherever we use `new_reserved`

src/bloom/command_handler.rs
```diff
@@ -227,7 +227,6 @@ pub fn bloom_filter_reserve(ctx: &Context, input_args: &[ValkeyString]) -> Valke
             return Err(ValkeyError::Str(utils::BAD_ERROR_RATE));
         }
     };
-    curr_cmd_idx += 1;
```
KarthikSubbarao: Let's add this back. It might have been accidentally removed in your PR

src/bloom/utils.rs
```diff
@@ -779,6 +794,7 @@ mod tests {
 
         // verify new_bf and bf
         assert_eq!(bf.fp_rate, new_bf.fp_rate);
+        assert_eq!(bf.tightening_ratio, new_bf.tightening_ratio);
```
KarthikSubbarao, line 833: In the verify_restored_items function, can we assert on this?
You can also do this assert for the false positive rate
```
        assert_eq!(
            restored_bloom_filter_type.tightenting_ratio,
            original_bloom_filter_type.tightenting_ratio
        );
```

src/bloom/command_handler.rs
```diff
@@ -298,15 +302,17 @@ pub fn bloom_filter_reserve(ctx: &Context, input_args: &[ValkeyString]) -> Valke
 
 pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
     let argc = input_args.len();
+    let validate_size_limit = ctx.get_flags().contains(ContextFlags::REPLICATED);
```
KarthikSubbarao:
```suggestion
    let replicated_cmd = ctx.get_flags().contains(ContextFlags::REPLICATED);
```

src/bloom/command_handler.rs
```diff
@@ -327,6 +333,21 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
                     }
                 };
             }
+            "TIGHTENING" => {
+                if !validate_size_limit {
+                    return Err(ValkeyError::Str(utils::ERROR));
+                }
```
KarthikSubbarao:
```suggestion
            "TIGHTENING" if replicated_cmd => {
                if idx >= (argc - 1) {
                    return Err(ValkeyError::WrongArity);
                }
```

src/bloom/command_handler.rs
```diff
@@ -298,15 +302,17 @@ pub fn bloom_filter_reserve(ctx: &Context, input_args: &[ValkeyString]) -> Valke
 
 pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
     let argc = input_args.len();
+    let validate_size_limit = ctx.get_flags().contains(ContextFlags::REPLICATED);
     // At the very least, we need: BF.INSERT <key> ITEMS <item>
-    if argc < 4 {
+    if !validate_size_limit && argc < 4 || validate_size_limit && argc < 5 {
```
KarthikSubbarao:
```suggestion
    if argc < 4 {
```

This should handle it already. It is not required to contain tightening ratio on the replicated command - it is optional and we should handle cases where it provided and not provided.

The check above is sufficient

src/bloom/command_handler.rs
```diff
@@ -405,6 +426,7 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
             }
             let mut bloom = match BloomFilterType::new_reserved(
                 fp_rate,
+                tightening_ratio,
```
KarthikSubbarao, line 433: Lets remove the following code at line 406
```
    let validate_size_limit = !ctx.get_flags().contains(ContextFlags::REPLICATED);
```

src/bloom/utils.rs
```diff
@@ -95,9 +100,11 @@ impl BloomFilterType {
         // Create the bloom filter and add to the main BloomFilter object.
         let bloom = BloomFilter::new(fp_rate, capacity);
         let filters = vec![bloom];
+        let tightening_ratio = 0.5;
```
KarthikSubbarao: Why is this needed? Removing it
```suggestion
```

tests/test_replication.py
```diff
@@ -34,7 +34,7 @@ def test_replication_behavior(self):
             ('BF.ADD', 'BF.ADD key item', 'BF.ADD key item1', 2),
             ('BF.MADD', 'BF.MADD key item', 'BF.MADD key item1', 2),
             ('BF.RESERVE', 'BF.RESERVE key 0.001 100000', 'BF.ADD key item1', 1),
-            ('BF.INSERT', 'BF.INSERT key items item', 'BF.INSERT key items item1', 2),
+            ('BF.INSERT', 'BF.INSERT key items item tightening 0.5', 'BF.INSERT key items item1 tightening 0.5', 2),
```
KarthikSubbarao, line 37: This is an invalid test. The tightening ratio is coming after `ITEM` so it is treated as an item. It is not treated as a property. Also, we cannot use this command on the primary, since we only allow this arg during replication :D 

Let's just use a normal BF.INSERT command.

Once we support the tightening ratio config, we can then (1) update the config on the primary node, (2) Use BF.INSERT (3) Validate Digest matches on the replica even though the tightening ratio config is different

KarthikSubbarao:
```suggestion
            ('BF.INSERT', 'BF.INSERT key items item', 'BF.INSERT key items item1', 2),
```

This is an invalid test. The tightening ratio is coming after `ITEM` so it is treated as an item. It is not treated as a property. Also, we cannot use this command on the primary, since we only allow this arg during replication :D 

Let's just use a normal BF.INSERT command.

Once we support the tightening ratio config, we can then (1) update the config on the primary node, (2) Use BF.INSERT (3) Validate Digest matches on the replica even though the tightening ratio config is different

For now, let us leave it as it is because it is essentially untested code (unless we use a monitor on replica during bloom creation on the primary node)

src/bloom/utils.rs
```diff
@@ -642,12 +672,18 @@ mod tests {
         let rand_prefix = random_prefix(7);
         // 1 in every 1000 operations is expected to be a false positive.
         let expected_fp_rate: f64 = 0.001;
+        let expected_tightening_ratio: f64 = 0.5;
```
KarthikSubbarao:
```suggestion
        let tightening_ratio: f64 = 0.5;
```

src/bloom/utils.rs
```diff
@@ -584,13 +608,19 @@ mod tests {
...
```
KarthikSubbarao:
```suggestion
        let tightening_ratio: f64 = 0.5;
```

We are not asserting against this. so there is no need to call it "expected"

src/bloom/utils.rs
nnmehta: Missed to remove this, it is not required

tests/test_replication.py
nnmehta: Ah got it, will revert this change accordingly

src/bloom/command_handler.rs
nnmehta: Will update

src/bloom/utils.rs
nnmehta: Will change accordingly

src/bloom/command_handler.rs
nnmehta: Why would we want to remove this?
Can we discuss this?

KarthikSubbarao (approved):

src/bloom/utils.rs
```diff
@@ -22,6 +25,8 @@ pub const BAD_EXPANSION: &str = "ERR bad expansion";
 pub const BAD_CAPACITY: &str = "ERR bad capacity";
 pub const BAD_ERROR_RATE: &str = "ERR bad error rate";
 pub const ERROR_RATE_RANGE: &str = "ERR (0 < error rate range < 1)";
+pub const BAD_ERROR_RATIO: &str = "ERR bad error ratio";
+pub const ERROR_RATIO_RANGE: &str = "ERR (0 < error ratio range < 1)";
```
zackcam, line 29: Would having a better error message i.e. something like `ratio must be in the range (0, 1)` or is following convention for error rate range better?

zackcam (approved):

PR #27, opened 2024-12-04
Support random seed per bloom object by default (configurable)

KarthikSubbarao:
Support random seed per bloom object by default (configurable) where all filters within the bloom object will use the same sip_keys (for their hash functions) as the first filter within the object which will be randomly generated.

When the new config (bf.bloom-use-random-seed) is disabled, new bloom object creates will use a fixed seed. In this case, the false positives on one bloom object can be reproduced on another bloom object with the same properties due to fixed seed making hashing results deterministic.

### DIFF
--- a/src/bloom/utils.rs
+++ b/src/bloom/utils.rs
@@ -1,14 +1,11 @@
+use super::data_type::BLOOM_TYPE_VERSION;
 use crate::{
-    configs::{
-        self, BLOOM_EXPANSION_MAX, BLOOM_EXPANSION_MIN, BLOOM_FP_RATE_MAX, BLOOM_FP_RATE_MIN,
-    },
+    configs::{self, BLOOM_EXPANSION_MAX, BLOOM_FP_RATE_MAX, BLOOM_FP_RATE_MIN},
     metrics,
 };
 use bloomfilter;
 use bloomfilter::{deserialize, serialize};
 use serde::{Deserialize, Serialize};
-
-use super::data_type::BLOOM_TYPE_VERSION;
 use std::{mem, sync::atomic::Ordering};
 
 /// KeySpace Notification Events
@@ -70,6 +67,7 @@ impl BloomError {
 pub struct BloomFilterType {
     pub expansion: u32,
     pub fp_rate: f64,
+    pub is_seed_random: bool,
     pub filters: Vec<BloomFilter>,
 }
 
@@ -79,6 +77,7 @@ impl BloomFilterType {
         fp_rate: f64,
         capacity: u32,
         expansion: u32,
+        use_random_seed: bool,
         validate_size_limit: bool,
     ) -> Result<BloomFilterType, BloomError> {
         // Reject the request, if the operation will result in creation of a bloom object containing a filter
@@ -91,14 +90,17 @@ impl BloomFilterType {
             mem::size_of::<BloomFilterType>(),
             std::sync::atomic::Ordering::Relaxed,
         );
-
         // Create the bloom filter and add to the main BloomFilter object.
-        let bloom = BloomFilter::new(fp_rate, capacity);
+        let bloom = match use_random_seed {
+            true => BloomFilter::with_random_seed(fp_rate, capacity),
+            false => BloomFilter::with_fixed_seed(fp_rate, capacity, &configs::FIXED_SEED),
+        };
         let filters = vec![bloom];
         let bloom = BloomFilterType {
             expansion,
             fp_rate,
             filters,
+            is_seed_random: use_random_seed,
         };
         Ok(bloom)
     }
@@ -118,6 +120,7 @@ impl BloomFilterType {
         BloomFilterType {
             expansion: from_bf.expansion,
             fp_rate: from_bf.fp_rate,
+            is_seed_random: from_bf.is_seed_random,
             filters,
         }
     }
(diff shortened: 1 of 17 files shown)


YueTang-Vanessa:
Please fix the failed checks

src/bloom/utils.rs
```diff
@@ -332,25 +349,30 @@ pub struct BloomFilter {
...
-    /// Instantiate empty BloomFilter object.
-    pub fn new(fp_rate: f64, capacity: u32) -> BloomFilter {
-        let bloom = bloomfilter::Bloom::new_for_fp_rate_with_seed(
-            capacity as usize,
-            fp_rate,
-            &configs::FIXED_SEED,
-        )
-        .expect("We expect bloomfilter::Bloom<[u8]> creation to succeed");
+    /// Instantiate empty BloomFilter object with a fixed seed used to create sip keys.
+    pub fn with_fixed_seed(fp_rate: f64, capacity: u32, fixed_seed: &[u8; 32]) -> BloomFilter {
```
YueTang-Vanessa, line 353: nitpick: consider renaming to `new_with_fixed_seed`

src/bloom/utils.rs
```diff
@@ -332,25 +349,30 @@ pub struct BloomFilter {
...
-        metrics::BLOOM_OBJECT_TOTAL_MEMORY_BYTES
-            .fetch_add(fltr.number_of_bytes(), std::sync::atomic::Ordering::Relaxed);
-        metrics::BLOOM_CAPACITY_ACROSS_OBJECTS
-            .fetch_add(capacity.into(), std::sync::atomic::Ordering::Relaxed);
+        fltr.incr_metrics_on_new_create();
+        fltr
+    }
+
+    /// Instantiate empty BloomFilter object with a randomly generated seed used to create sip keys.
+    pub fn with_random_seed(fp_rate: f64, capacity: u32) -> BloomFilter {
```
YueTang-Vanessa, line 367: nitpick: considering renaming to `new_with_random_seed`

src/bloom/utils.rs
```diff
@@ -581,183 +629,199 @@ mod tests {
...
+                &restore_bf,
+                add_operation_idx,
+                expected_fp_rate,
+                fp_margin,
+                &rand_prefix,
+            );
+        }
     }
 
     #[test]
```
zackcam: Instead of having  `let is_seed_random_vec = [true, false];` and doing a for loop would doing something like this work for the test function `#[parameterize(is_seed_random = [true, false])]`?

src/bloom/utils.rs
```diff
@@ -581,183 +629,199 @@ mod tests {
...
```
zackcam: Instead of having let is_seed_random_vec = [true, false]; and doing a for loop would doing something like this work for the test function #[parameterize(is_seed_random = [true, false])]?

src/bloom/utils.rs
```diff
@@ -581,183 +629,199 @@ mod tests {
...
-        let new_bf = new_bf_result.unwrap();
-
-        // verify new_bf and bf
-        assert_eq!(bf.fp_rate, new_bf.fp_rate);
-        assert_eq!(bf.expansion, new_bf.expansion);
-        assert_eq!(bf.capacity(), new_bf.capacity());
-
-        // contains key
-        assert!(new_bf.item_exists(key.as_bytes()));
+        // Validate with non scaling (0) and scaling (2).
```
zackcam: Instead of having `let test_expansions = [0, 2]` and doing a for loop would doing something like this work for the test function `#[parameterize(expansion_rate = [0, 2])]`? Would make it so we dont have a loop and might be easier to follow

KarthikSubbarao: We will need a testing dependency for this. Maybe: https://docs.rs/rstest/latest/rstest/

src/bloom/utils.rs
KarthikSubbarao: With Rust naming conventions, both  `with_fixed_seed` or `new` are valid.
`with_fixed_seed` implies that it is new and it is creating a new instance

src/bloom/utils.rs
KarthikSubbarao: With Rust naming conventions, both  `new_with_random_seed` or `new` are valid.
`new_with_random_seed` implies that it is new and it is creating a new instance

src/bloom/utils.rs
KarthikSubbarao: We can consider adding this later

src/bloom/utils.rs
KarthikSubbarao: Replied below

src/bloom/utils.rs
KarthikSubbarao: Replied below

src/configs.rs
```diff
@@ -44,3 +48,7 @@ pub const FIXED_SIP_KEY_ONE_A: u64 = 15713473521876537177;
 pub const FIXED_SIP_KEY_ONE_B: u64 = 15671187751654921383;
 pub const FIXED_SIP_KEY_TWO_A: u64 = 9766223185946773789;
 pub const FIXED_SIP_KEY_TWO_B: u64 = 9453907914610147120;
+pub const SIP_KEYS: [(u64, u64); 2] = [
+    (FIXED_SIP_KEY_ONE_A, FIXED_SIP_KEY_ONE_B),
+    (FIXED_SIP_KEY_TWO_A, FIXED_SIP_KEY_TWO_B),
+];
```
KarthikSubbarao:
```suggestion
```

src/bloom/utils.rs
KarthikSubbarao: I added the parameterization with the rstest dependency

PR #28, opened 2024-12-04
Update default configuration and related test.

YueTang-Vanessa:
In this PR:
* Update default bloom filter configuration
    * capacity: data type: u32 -> i64, default value: 100
    * num_items: data type: u32 -> i64, default value: 0.01
    * max limit of capacity, expansion has been updated.


src/configs.rs
```diff
@@ -2,15 +2,15 @@ use lazy_static::lazy_static;
 use std::sync::atomic::AtomicI64;
 
 /// Configurations
-pub const BLOOM_CAPACITY_DEFAULT: i64 = 100000;
-pub const BLOOM_CAPACITY_MIN: u32 = 1;
-pub const BLOOM_CAPACITY_MAX: u32 = u32::MAX;
+pub const BLOOM_CAPACITY_DEFAULT: i64 = 100;
+pub const BLOOM_CAPACITY_MIN: i64 = 1;
+pub const BLOOM_CAPACITY_MAX: i64 = i32::MAX as i64 * 2;
```
KarthikSubbarao: Why not `i64::MAX`?

src/configs.rs
```diff
@@ -2,15 +2,15 @@ use lazy_static::lazy_static;
...
 
 pub const BLOOM_EXPANSION_DEFAULT: i64 = 2;
-pub const BLOOM_EXPANSION_MIN: u32 = 1;
-pub const BLOOM_EXPANSION_MAX: u32 = 10;
+pub const BLOOM_EXPANSION_MIN: i64 = 1;
+pub const BLOOM_EXPANSION_MAX: i64 = i32::MAX as i64 * 2;
```
KarthikSubbarao: expansion on the BloomFilterDataType is a u32. this should be `u32::MAX`, right?

tests/test_save_and_restore.py
```diff
@@ -22,8 +22,8 @@ def test_basic_save_and_restore(self):
         # save rdb, restart sever
         client.bgsave()
         self.server.wait_for_save_done()
-        # Keep the server running for 1 second more to have a larger uptime.
-        time.sleep(1)
+        # Keep the server running for 2 seconds more to have a larger uptime.
+        time.sleep(2)
         uptime_in_sec_1 = self.client.info_obj().uptime_in_secs()
```
KarthikSubbarao: Let's remove this. Time makes this flaky.

tests/test_save_and_restore.py
```diff
@@ -22,8 +22,8 @@ def test_basic_save_and_restore(self):
...
         self.server.restart(remove_rdb=False, remove_nodes_conf=False, connect_client=True)
         uptime_in_sec_2 = self.client.info_obj().uptime_in_secs()
```
KarthikSubbarao: 1. We can remove this and just use a waiter until `self.server.is_rdb_done_loading()` is true
2. We can validate that we see RDB Loading is done in the engine logs

src/bloom/utils.rs
```diff
@@ -68,7 +68,7 @@ impl BloomError {
 /// This is a generic top level structure which is not coupled to any bloom crate.
 #[derive(Serialize, Deserialize)]
 pub struct BloomFilterType {
-    pub expansion: u32,
+    pub expansion: i64,
```
KarthikSubbarao: This can remain an i64, we don't expect expansion is be >4Billion (u32::MAX)

src/configs.rs
YueTang-Vanessa: During our sync up offline we discussed about this, at that time we thought`i64::MAX` is too big.

src/configs.rs
YueTang-Vanessa: This PR was raised before the `BloomFilterDataType` merged. Will update it.

tests/test_save_and_restore.py
YueTang-Vanessa: From previous failed check log, this test failed due to `uptime_in_sec_1 = uptime_in_sec_2 (= 1)` , which very likely due to restore operation completed within ms. Both checking the `RDB Loading log` and the `self.server.is_rdb_done_loading()` won't help with this situation.

src/bloom/utils.rs
```diff
@@ -210,7 +210,7 @@ impl BloomFilterType {
                 Ok(rate) => rate,
                 Err(e) => return Err(e),
             };
-            let new_capacity = match filter.capacity.checked_mul(self.expansion) {
+            let new_capacity = match filter.capacity.checked_mul(self.expansion.into()) {
```
KarthikSubbarao, line 213: why do we need into()?

YueTang-Vanessa: Because capacity is i64 and expansion is u32 now

Issue #11, opened 2024-10-07
Support for ACL Categories for BF.* commands


KarthikSubbarao:
BF.* commands should have a new "@bloom" ACL Category such that clients authenticated as a user with "@bloom" permissions can execute these commands.

This can be implemented using the valkey-module-rs SDK changes implemented from this issue: https://github.com/valkey-io/valkeymodule-rs/issues/96

zackcam:
Hi - I'd like to start working on this in conjunction with [valkey-io/valkeymodule-rs#96](https://github.com/valkey-io/valkeymodule-rs/issues/96), can this be assigned to me as well.

PR #29, opened 2024-12-06
Adding bloom acl category to commands. Added tests to check that the commands contain the correct acl categories

zackcam:
# Changes
Addressing issue: #11 
~~Currently the dependency for valkey_module doesn't have the change required to create acl categories so the cargo.toml file will be updated once that change is released.~~
Updated dependency of valkey-modules to v0.1.3
Changes in this pr include having to add an 8.0 and 7.2 feature flag as the create_acl_category and set_acl_category methods are behind these flags in the dependency. 
For each bloom command we have added a bloom category as well as created the bloom category.
# Testing
Testing: There are two tests. We have test that creates two users one with bloom acl category access and one without. We then check that we get the expected behavior when running bloom commands. The other test, checks that acl categories are set correctly for each command. We check that both bloom and other categories are present.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ homepage = "https://github.com/valkey-io/valkey-bloom"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-valkey-module = "0.1.2"
+valkey-module = { version = "0.1.3", features = ["min-valkey-compatibility-version-8-0", "min-redis-compatibility-version-7-2"]}
 valkey-module-macros = "0"
 linkme = "0"
 bloomfilter = { version = "3.0.1", features = ["serde"] }
@@ -35,4 +35,6 @@ debug = 2
 debug-assertions = true
 
 [features]
+default = ["min-valkey-compatibility-version-8-0"]
 enable-system-alloc = ["valkey-module/enable-system-alloc"]
+min-valkey-compatibility-version-8-0 = []
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,6 +7,7 @@ pub mod metrics;
 pub mod wrapper;
 use crate::bloom::command_handler;
 use crate::bloom::data_type::BLOOM_FILTER_TYPE;
+use valkey_module::logging;
 use valkey_module_macros::info_command_handler;
 
 pub const MODULE_NAME: &str = "bf";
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -12,7 +12,7 @@ homepage = "https://github.com/valkey-io/valkey-bloom"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-valkey-module = "0.1.2"
+valkey-module = { path = "../valkeymodule-rs", features = ["min-valkey-compatibility-version-8-0", "min-redis-compatibility-version-7-2"]}
```
YueTang-Vanessa: This change seems break the test. https://github.com/valkey-io/valkey-bloom/actions/runs/12190295376/job/34007129431

src/lib.rs
```diff
@@ -85,16 +86,19 @@ valkey_module! {
...
-        ["BF.ADD", bloom_add_command, "write fast deny-oom", 1, 1, 1],
-        ["BF.MADD", bloom_madd_command, "write fast deny-oom", 1, 1, 1],
-        ["BF.EXISTS", bloom_exists_command, "readonly fast", 1, 1, 1],
-        ["BF.MEXISTS", bloom_mexists_command, "readonly fast", 1, 1, 1],
-        ["BF.CARD", bloom_card_command, "readonly fast", 1, 1, 1],
-        ["BF.RESERVE", bloom_reserve_command, "write fast deny-oom", 1, 1, 1],
-        ["BF.INFO", bloom_info_command, "readonly fast", 1, 1, 1],
-        ["BF.INSERT", bloom_insert_command, "write fast deny-oom", 1, 1, 1],
-        ["BF.LOAD", bloom_load_command, "write fast deny-oom", 1, 1, 1]
+        ["BF.ADD", bloom_add_command, "write fast deny-oom", 1, 1, 1, "bloom"],
```
KarthikSubbarao, line 93: Nice!

tests/test_bloom_acl_category.py
```diff
@@ -0,0 +1,66 @@
...
+            except Exception as e:
+                assert str(e) == f"User nonbloomuser has no permissions to run the '{cmd_name}' command"
+
+        # Switch to the user with bloom command access and check commands are run as expected
+        client.execute_command(f"AUTH bloomuser bloom_pass")
+        for cmd in bloom_commands:
+            cmd_name = cmd[0].split()[0]
+            try:
+                result = client.execute_command(cmd[0])
+                print(result)
```
KarthikSubbarao: can be removed

Cargo.toml
zackcam: This is expected the changes that allow this will be released soon. The pr will be updated when this dependency gets updated and has the necessary changes

tests/test_bloom_acl_category.py
zackcam: Removed

KarthikSubbarao (approved):
Approved and waiting for the new valkeymodule-rs version to be released. Then, the version in cargo.toml can be updated and we can merge this in.

Confirmed that all tests pass once we use the new version.

YueTang-Vanessa (approved):

nnmehta (approved):

PR #30, opened 2024-12-06
Adding a wrong type error message when performing a bloom operation on a non bloom key

zackcam:
Updated the error on the get_value return to be a WRONG_TYPE_ERROR. Added a test as well checking this happens and the error message is correct for each bloom command.

### DIFF
--- a/src/bloom/command_handler.rs
+++ b/src/bloom/command_handler.rs
@@ -87,7 +87,7 @@ pub fn bloom_filter_add_value(
     let value = match filter_key.get_value::<BloomFilterType>(&BLOOM_FILTER_TYPE) {
         Ok(v) => v,
         Err(_) => {
-            return Err(ValkeyError::Str(utils::ERROR));
+            return Err(ValkeyError::WrongType);
         }
     };
     // Skip bloom filter size validation on replicated cmds.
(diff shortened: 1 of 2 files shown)


src/bloom/utils.rs
```diff
@@ -32,7 +32,7 @@ pub const ENCODE_BLOOM_FILTER_FAILED: &str = "ERR encode bloom filter failed.";
 pub const DECODE_BLOOM_FILTER_FAILED: &str = "ERR decode bloom filter failed.";
 pub const DECODE_UNSUPPORTED_VERSION: &str =
     "ERR decode bloom filter failed.  Unsupported version.";
-
+pub const WRONG_TYPE_ERROR: &str = "WRONGTYPE Not a BloomFilter key";
```
KarthikSubbarao: Could you check what message this returns? Maybe we can use this
```
https://docs.rs/valkey-module/latest/src/valkey_module/rediserror.rs.html#11
```

src/bloom/command_handler.rs
```diff
@@ -87,7 +87,7 @@ pub fn bloom_filter_add_value(
     let value = match filter_key.get_value::<BloomFilterType>(&BLOOM_FILTER_TYPE) {
         Ok(v) => v,
         Err(_) => {
```
KarthikSubbarao, line 89: The best practice is to handle every case in a match. I understand that the SDK does not make this easy for us because they return a specific string message when an ENUM would have made matching easier.

In the future, if it is possible to update the SDK, we should be able to update our handling to make it stricter.

https://github.com/valkey-io/valkeymodule-rs/blob/0a1fb4e8e9b5cc3e26d8fcdbab0857c8bfd87f33/src/key.rs#L678

KarthikSubbarao (approved):

PR #31, opened 2024-12-09
Add TIGHTENING Ratio and FP_Rate module string configs and parse to float

nnmehta:
Supporting False Positive config and Tightening Ratio Config.

We have already implemented feature for module configs to support validation & rejection of config sets so that we have the right values for False Positive config and Tightening Ratio Config. https://github.com/valkey-io/valkeymodule-rs/pull/144

With these changes, we can fine tune the configs based on user's workloads and requirements

### DIFF
--- a/src/bloom/command_handler.rs
+++ b/src/bloom/command_handler.rs
@@ -4,7 +4,7 @@ use crate::bloom::utils::BloomFilterType;
 use crate::configs;
 use crate::configs::{
     BLOOM_CAPACITY_MAX, BLOOM_CAPACITY_MIN, BLOOM_EXPANSION_MAX, BLOOM_EXPANSION_MIN,
-    BLOOM_FP_RATE_MAX, BLOOM_FP_RATE_MIN, TIGHTENING_RATIO_MAX, TIGHTENING_RATIO_MIN,
+    BLOOM_FP_RATE_MAX, BLOOM_FP_RATE_MIN, BLOOM_TIGHTENING_RATIO_MAX, BLOOM_TIGHTENING_RATIO_MIN,
 };
 use std::sync::atomic::Ordering;
 use valkey_module::ContextFlags;
@@ -203,8 +203,12 @@ pub fn bloom_filter_add_value(
         }
         None => {
             // Instantiate empty bloom filter.
-            let fp_rate = configs::BLOOM_FP_RATE_DEFAULT;
-            let tightening_ratio = configs::TIGHTENING_RATIO;
+            let fp_rate = *configs::BLOOM_FP_RATE_F64
+                .lock()
+                .expect("Unable to get a lock on fp_rate static");
+            let tightening_ratio = *configs::BLOOM_TIGHTENING_F64
+                .lock()
+                .expect("Unable to get a lock on tightening ratio static");
             let capacity = configs::BLOOM_CAPACITY.load(Ordering::Relaxed);
             let expansion = configs::BLOOM_EXPANSION.load(Ordering::Relaxed) as u32;
             let use_random_seed = configs::BLOOM_USE_RANDOM_SEED.load(Ordering::Relaxed);
@@ -401,7 +405,9 @@ pub fn bloom_filter_reserve(ctx: &Context, input_args: &[ValkeyString]) -> Valke
             };
             // Skip bloom filter size validation on replicated cmds.
             let validate_size_limit = !ctx.get_flags().contains(ContextFlags::REPLICATED);
-            let tightening_ratio = configs::TIGHTENING_RATIO;
+            let tightening_ratio = *configs::BLOOM_TIGHTENING_F64
+                .lock()
+                .expect("Failed to lock tightening ratio");
             let bloom = match BloomFilterType::new_reserved(
                 fp_rate,
                 tightening_ratio,
@@ -444,8 +450,12 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
     let filter_name = &input_args[idx];
     idx += 1;
     let replicated_cmd = ctx.get_flags().contains(ContextFlags::REPLICATED);
-    let mut fp_rate = configs::BLOOM_FP_RATE_DEFAULT;
-    let mut tightening_ratio = configs::TIGHTENING_RATIO;
+    let mut fp_rate = *configs::BLOOM_FP_RATE_F64
+        .lock()
+        .expect("Unable to get a lock on fp_rate static");
+    let mut tightening_ratio = *configs::BLOOM_TIGHTENING_F64
+        .lock()
+        .expect("Unable to get a lock on tightening ratio static");
     let mut capacity = configs::BLOOM_CAPACITY.load(Ordering::Relaxed);
     let mut expansion = configs::BLOOM_EXPANSION.load(Ordering::Relaxed) as u32;
     let use_random_seed = configs::BLOOM_USE_RANDOM_SEED.load(Ordering::Relaxed);
@@ -479,8 +489,15 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
                 }
                 idx += 1;
                 tightening_ratio = match input_args[idx].to_string_lossy().parse::<f64>() {
-                    Ok(num) if num > TIGHTENING_RATIO_MIN && num < TIGHTENING_RATIO_MAX => num,
-                    Ok(num) if !(num > TIGHTENING_RATIO_MIN && num < TIGHTENING_RATIO_MAX) => {
+                    Ok(num)
+                        if num > BLOOM_TIGHTENING_RATIO_MIN && num < BLOOM_TIGHTENING_RATIO_MAX =>
+                    {
+                        num
+                    }
+                    Ok(num)
+                        if !(num > BLOOM_TIGHTENING_RATIO_MIN
+                            && num < BLOOM_TIGHTENING_RATIO_MAX) =>
+                    {
                         return Err(ValkeyError::Str(utils::ERROR_RATIO_RANGE));
                     }
                     _ => {
--- a/src/bloom/utils.rs
+++ b/src/bloom/utils.rs
@@ -1,8 +1,8 @@
 use super::data_type::BLOOM_TYPE_VERSION;
 use crate::{
     configs::{
-        self, BLOOM_EXPANSION_MAX, BLOOM_FP_RATE_MAX, BLOOM_FP_RATE_MIN, TIGHTENING_RATIO_MAX,
-        TIGHTENING_RATIO_MIN,
+        self, BLOOM_EXPANSION_MAX, BLOOM_FP_RATE_MAX, BLOOM_FP_RATE_MIN,
+        BLOOM_TIGHTENING_RATIO_MAX, BLOOM_TIGHTENING_RATIO_MIN,
     },
     metrics,
 };
@@ -265,10 +265,11 @@ impl BloomFilterType {
             }
             // Scale out by adding a new filter with capacity bounded within the u32 range. false positive rate is also
             // bound within the range f64::MIN_POSITIVE <= x < 1.0.
-            let new_fp_rate = match Self::calculate_fp_rate(self.fp_rate, num_filters) {
-                Ok(rate) => rate,
-                Err(e) => return Err(e),
-            };
+            let new_fp_rate =
+                match Self::calculate_fp_rate(self.fp_rate, num_filters, self.tightening_ratio) {
+                    Ok(rate) => rate,
+                    Err(e) => return Err(e),
+                };
             let new_capacity = match filter.capacity.checked_mul(self.expansion.into()) {
                 Some(new_capacity) => new_capacity,
                 None => {
@@ -325,8 +326,12 @@ impl BloomFilterType {
     }
 
     /// Calculate the false positive rate for the Nth filter using tightening ratio.
-    pub fn calculate_fp_rate(fp_rate: f64, num_filters: i32) -> Result<f64, BloomError> {
-        match fp_rate * configs::TIGHTENING_RATIO.powi(num_filters) {
+    pub fn calculate_fp_rate(
+        fp_rate: f64,
+        num_filters: i32,
+        tightening_ratio: f64,
+    ) -> Result<f64, BloomError> {
+        match fp_rate * tightening_ratio.powi(num_filters) {
             x if x > f64::MIN_POSITIVE => Ok(x),
             _ => Err(BloomError::MaxNumScalingFilters),
         }
@@ -374,7 +379,9 @@ impl BloomFilterType {
                         if !(values.1 > BLOOM_FP_RATE_MIN && values.1 < BLOOM_FP_RATE_MAX) {
                             return Err(BloomError::ErrorRateRange);
                         }
-                        if !(values.2 > TIGHTENING_RATIO_MIN && values.2 < TIGHTENING_RATIO_MAX) {
+                        if !(values.2 > BLOOM_TIGHTENING_RATIO_MIN
+                            && values.2 < BLOOM_TIGHTENING_RATIO_MAX)
+                        {
                             return Err(BloomError::ErrorRateRange);
                         }
                         if values.4.len() >= configs::MAX_FILTERS_PER_OBJ as usize {
--- a/src/configs.rs
+++ b/src/configs.rs
@@ -1,5 +1,13 @@
+use crate::bloom::utils;
 use lazy_static::lazy_static;
 use std::sync::atomic::{AtomicBool, AtomicI64};
+use std::sync::Mutex;
+use valkey_module::logging;
+use valkey_module::{
+    configuration::{ConfigurationContext, ConfigurationFlags},
+    valkey_module, ConfigurationValue, Context, InfoContext, Status, ValkeyError, ValkeyGILGuard,
+    ValkeyResult, ValkeyString,
+};
 
 /// Configurations
 pub const BLOOM_CAPACITY_DEFAULT: i64 = 100;
@@ -10,9 +18,15 @@ pub const BLOOM_EXPANSION_DEFAULT: i64 = 2;
 pub const BLOOM_EXPANSION_MIN: u32 = 1;
 pub const BLOOM_EXPANSION_MAX: u32 = u32::MAX;
 
-pub const BLOOM_FP_RATE_DEFAULT: f64 = 0.01;
+pub const BLOOM_FP_RATE_DEFAULT: &str = "0.01";
 pub const BLOOM_FP_RATE_MIN: f64 = 0.0;
 pub const BLOOM_FP_RATE_MAX: f64 = 1.0;
+
+// Tightening ratio used during scale out for the calculation of fp_rate of every new filter within a bloom object to
+// maintain the bloom object's overall fp_rate to the configured value.
+pub const TIGHTENING_RATIO_DEFAULT: &str = "0.5";
+pub const BLOOM_TIGHTENING_RATIO_MIN: f64 = 0.0;
+pub const BLOOM_TIGHTENING_RATIO_MAX: f64 = 1.0;
 
 pub const BLOOM_USE_RANDOM_SEED_DEFAULT: bool = true;
 
@@ -31,14 +45,23 @@ lazy_static! {
         AtomicI64::new(BLOOM_MEMORY_LIMIT_PER_FILTER_DEFAULT);
     pub static ref BLOOM_USE_RANDOM_SEED: AtomicBool = AtomicBool::default();
     pub static ref BLOOM_DEFRAG: AtomicBool = AtomicBool::new(BLOOM_DEFRAG_DEAFULT);
+    pub static ref BLOOM_FP_RATE_F64: Mutex<f64> = Mutex::new(
+        BLOOM_FP_RATE_DEFAULT
+            .parse::<f64>()
+            .expect("Expected valid f64 for fp rate.")
+    );
+    pub static ref BLOOM_FP_RATE: ValkeyGILGuard<ValkeyString> =
+        ValkeyGILGuard::new(ValkeyString::create(None, BLOOM_FP_RATE_DEFAULT));
+    pub static ref BLOOM_TIGHTENING_F64: Mutex<f64> = Mutex::new(
+        TIGHTENING_RATIO_DEFAULT
+            .parse::<f64>()
+            .expect("Expected valid f64 for tightening ratio.")
+    );
+    pub static ref BLOOM_TIGHTENING_RATIO: ValkeyGILGuard<ValkeyString> =
+        ValkeyGILGuard::new(ValkeyString::create(None, TIGHTENING_RATIO_DEFAULT));
 }
 
 /// Constants
-// Tightening ratio used during scale out for the calculation of fp_rate of every new filter within a bloom object to
-// maintain the bloom object's overall fp_rate to the configured value.
-pub const TIGHTENING_RATIO: f64 = 0.5;
-pub const TIGHTENING_RATIO_MIN: f64 = 0.0;
-pub const TIGHTENING_RATIO_MAX: f64 = 1.0;
 // Max number of filters allowed within a bloom object.
 pub const MAX_FILTERS_PER_OBJ: i32 = i32::MAX;
 /// Below constants are fixed seed and sip keys to help create bloom objects using the same seed and to restore the bloom objects with the same hasher which
@@ -47,3 +70,43 @@ pub const FIXED_SEED: [u8; 32] = [
     89, 15, 245, 34, 234, 120, 17, 218, 167, 20, 216, 9, 59, 62, 123, 217, 29, 137, 138, 115, 62,
     152, 136, 135, 48, 127, 151, 205, 40, 7, 51, 131,
 ];
+
+/// This is a config set handler for the False Positive Rate and Tightening Ratio configs.
+pub fn on_string_config_set(
+    config_ctx: &ConfigurationContext,
+    name: &str,
+    val: &'static ValkeyGILGuard<ValkeyString>,
+) -> Result<(), ValkeyError> {
+    let v = val.get(config_ctx);
+    let value_str = v.to_string_lossy();
+    let value = match value_str.parse::<f64>() {
+        Ok(v) => v,
+        Err(_) => {
+            return Err(ValkeyError::Str("Invalid floating-point value"));
+        }
+    };
+
+    match name {
+        "bloom-fp-rate" => {
+            if !(BLOOM_FP_RATE_MIN..BLOOM_FP_RATE_MAX).contains(&value) {
+                return Err(ValkeyError::Str(utils::ERROR_RATE_RANGE));
+            }
+            let mut fp_rate = BLOOM_FP_RATE_F64
+                .lock()
+                .expect("We expect the fp_rate static to exist.");
+            *fp_rate = value;
+            Ok(())
+        }
+        "bloom-tightening-ratio" => {
+            if !(BLOOM_TIGHTENING_RATIO_MIN..BLOOM_TIGHTENING_RATIO_MAX).contains(&value) {
+                return Err(ValkeyError::Str(utils::ERROR_RATIO_RANGE));
+            }
+            let mut tightening = BLOOM_TIGHTENING_F64
+                .lock()
+                .expect("We expect the tightening_ratio static to exist.");
+            *tightening = value;
+            Ok(())
+        }
+        _ => Err(ValkeyError::Str("Unknown configuration parameter")),
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,8 @@
 use metrics::bloom_info_handler;
-use valkey_module::configuration::ConfigurationFlags;
-use valkey_module::{valkey_module, Context, InfoContext, Status, ValkeyResult, ValkeyString};
+use valkey_module::{
+    configuration::ConfigurationFlags, valkey_module, Context, InfoContext, Status, ValkeyGILGuard,
+    ValkeyResult, ValkeyString,
+};
 pub mod bloom;
 pub mod configs;
 pub mod metrics;
@@ -106,6 +108,8 @@ valkey_module! {
             ["bloom-memory-limit-per-filter", &*configs::BLOOM_MEMORY_LIMIT_PER_FILTER, configs::BLOOM_MEMORY_LIMIT_PER_FILTER_DEFAULT, configs::BLOOM_MEMORY_LIMIT_PER_FILTER_MIN, configs::BLOOM_MEMORY_LIMIT_PER_FILTER_MAX, ConfigurationFlags::DEFAULT, None],
         ],
         string: [
+            ["bloom-fp-rate", &*configs::BLOOM_FP_RATE, configs::BLOOM_FP_RATE_DEFAULT, ConfigurationFlags::DEFAULT, None, Some(Box::new(configs::on_string_config_set))],
+            ["bloom-tightening-ratio", &*configs::BLOOM_TIGHTENING_RATIO, configs::TIGHTENING_RATIO_DEFAULT, ConfigurationFlags::DEFAULT, None, Some(Box::new(configs::on_string_config_set))],
         ],
         bool: [
             ["bloom-use-random-seed", &*configs::BLOOM_USE_RANDOM_SEED, configs::BLOOM_USE_RANDOM_SEED_DEFAULT, ConfigurationFlags::DEFAULT, None],
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -299,6 +299,22 @@ class TestBloomBasic(ValkeyBloomTestCaseBase):
         else:
             madd_scenario_object_digest == madd_default_object_digest
 
+        # scenario 6 validates that digest differs on bloom objects after changing the tightening_ratio config
+        client.execute_command('BF.RESERVE tightening_ratio 0.001 1000')
+        assert self.client.execute_command('CONFIG SET bf.bloom-tightening-ratio 0.75') == b'OK'
+        client.execute_command('BF.RESERVE tightening_ratio2 0.001 1000')
+        scenario_tightening_ratio_object_digest = client.execute_command('DEBUG DIGEST-VALUE tightening_ratio')
+        scenario_tightening_ratio2_digest = client.execute_command('DEBUG DIGEST-VALUE tightening_ratio2')
+        assert scenario_tightening_ratio_object_digest != scenario_tightening_ratio2_digest
+
+        # scenario 7 validates that digest differs on bloom objects after changing the fp_rate config
+        client.execute_command('BF.INSERT fp_rate capacity 1000 items 1')
+        assert self.client.execute_command('CONFIG SET bf.bloom-fp-rate 0.5') == b'OK'
+        client.execute_command('BF.INSERT fp_rate2 capacity 1000 items 1')
+        fp_rate_object_digest = client.execute_command('DEBUG DIGEST-VALUE fp_rate')
+        scenario_fp_rate2_digest = client.execute_command('DEBUG DIGEST-VALUE fp_rate2')
+        assert fp_rate_object_digest != scenario_fp_rate2_digest
+
     def test_bloom_wrong_type(self):
         # List of all bloom commands
         bloom_commands = [
@@ -323,3 +339,19 @@ class TestBloomBasic(ValkeyBloomTestCaseBase):
             except Exception as e:
                 
                 assert str(e) == f"WRONGTYPE Operation against a key holding the wrong kind of value"
+    
+    def test_bloom_string_config_set(self):
+        """
+        This is a test that validates the bloom string configuration set logic.
+        """     
+        assert self.client.execute_command('CONFIG SET bf.bloom-fp-rate 0.1') == b'OK'
+        assert self.client.execute_command('CONFIG SET bf.bloom-tightening-ratio 0.75') == b'OK'
+        
+        assert self.client.execute_command('CONFIG GET bf.bloom-fp-rate')[1] == b'0.1'
+        assert self.client.execute_command('CONFIG GET bf.bloom-tightening-ratio')[1] == b'0.75'
+        try:
+            assert self.client.execute_command('CONFIG SET bf.bloom-fp-rate 1.1') == b'ERR (0 < error rate range < 1)'\
+                and self.client.execute_command('CONFIG SET bf.bloom-tightening-ratio 1.75') == b'ERR (0 < error ratio range < 1)'
+        except ResponseError as e:
+            assert str(e) == f"CONFIG SET failed (possibly related to argument 'bf.bloom-fp-rate') - ERR (0 < error rate range < 1)"\
+                and f"CONFIG SET failed (possibly related to argument 'bf.bloom-tightening-ratio') - ERR (0 < error rate range < 1)"
--- a/tests/test_replication.py
+++ b/tests/test_replication.py
@@ -161,6 +161,8 @@ class TestBloomReplication(ReplicationTestCase):
         # replicated with the properties below.
         assert self.client.execute_command('CONFIG SET bf.bloom-capacity 1000') == b'OK'
         assert self.client.execute_command('CONFIG SET bf.bloom-expansion 3') == b'OK'
+        assert self.client.execute_command('CONFIG SET bf.bloom-fp-rate 0.1') == b'OK'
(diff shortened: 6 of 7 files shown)


src/configs.rs
```diff
@@ -47,3 +61,66 @@ pub const FIXED_SEED: [u8; 32] = [
...
+    value: &'static ValkeyGILGuard<ValkeyString>,
+) {
+    let fp_rate_str = value.get(config_ctx).to_string_lossy();
+
+    // Try to parse the string value to f64
+    match fp_rate_str.parse::<f64>() {
+        Ok(fp_rate) => {
+            // Validate the range
+            if !(BLOOM_FP_RATE_MIN..=BLOOM_FP_RATE_MAX).contains(&fp_rate) {
+                logging::log_warning(format!(
```
YueTang-Vanessa: Why we use logging instead of throw `Err` with message?

src/configs.rs
```diff
@@ -47,3 +61,66 @@ pub const FIXED_SEED: [u8; 32] = [
...
+    config_ctx: &ConfigurationContext,
+    _name: &str,
+    value: &'static ValkeyGILGuard<ValkeyString>,
+) {
+    let ratio_str = value.get(config_ctx).to_string_lossy();
+
+    match ratio_str.parse::<f64>() {
+        Ok(ratio) => {
+            if !(TIGHTENING_RATIO_MIN..=TIGHTENING_RATIO_MAX).contains(&ratio) {
+                logging::log_warning(format!(
```
YueTang-Vanessa: Same as above. Why not return Err?

src/configs.rs
```diff
@@ -47,3 +61,66 @@ pub const FIXED_SEED: [u8; 32] = [
     89, 15, 245, 34, 234, 120, 17, 218, 167, 20, 216, 9, 59, 62, 123, 217, 29, 137, 138, 115, 62,
     152, 136, 135, 48, 127, 151, 205, 40, 7, 51, 131,
 ];
+
+pub fn set_bloom_fp_rate(
```
YueTang-Vanessa: Can you add test case for this new function?

src/configs.rs
```diff
@@ -47,3 +61,66 @@ pub const FIXED_SEED: [u8; 32] = [
...
+        Err(e) => {
+            logging::log_warning(format!(
+                "Invalid value for bloom-fp-rate: {}, error: {}",
+                fp_rate_str, e
+            ));
+        }
+    }
+}
+
+pub fn set_tightening_ratio(
```
YueTang-Vanessa: Can you add test case for this new function?

src/lib.rs
```diff
@@ -107,6 +109,8 @@ valkey_module! {
             ["bloom-memory-limit-per-filter", &*configs::BLOOM_MEMORY_LIMIT_PER_FILTER, configs::BLOOM_MEMORY_LIMIT_PER_FILTER_DEFAULT, configs::BLOOM_MEMORY_LIMIT_PER_FILTER_MIN, configs::BLOOM_MEMORY_LIMIT_PER_FILTER_MAX, ConfigurationFlags::DEFAULT, None],
         ],
         string: [
```
KarthikSubbarao, line 110: These new configs need to be tested.
1. Basic config set and get tests for valid and invalid values.
2. Create object 1. Change fp rate config. Create object 2. Digest should be different
3. Create object 1. Change tightening ratio config. Create object 2. Digest should be different
4. For replication testing, ensure that even if a primary node has updated configs and the replica does not, we should have same digest on the replicated bloom object creation. To test this, change the configs here to a non default value: https://github.com/valkey-io/valkey-bloom/blob/unstable/tests/test_replication.py#L163

src/configs.rs
KarthikSubbarao: Can we use one single callback for both configs?

You can have matching based logic to perform both config handling in the same function. So you can match by name.

src/configs.rs
KarthikSubbarao: +1 you can remove the log and reject with an error message

src/configs.rs
```diff
@@ -47,3 +61,66 @@ pub const FIXED_SEED: [u8; 32] = [
...
+            if !(BLOOM_FP_RATE_MIN..=BLOOM_FP_RATE_MAX).contains(&fp_rate) {
+                logging::log_warning(format!(
+                    "FP rate {} out of range [{}, {}]",
+                    fp_rate, BLOOM_FP_RATE_MIN, BLOOM_FP_RATE_MAX
+                ));
+                return;
+            }
+            // If validation passes, update the value
+            BLOOM_FP_RATE.set(config_ctx, value.get(config_ctx).clone());
+            logging::log_notice(format!("Successfully set bloom-fp-rate to {}", fp_rate));
```
KarthikSubbarao: We don't need to log the success path. This is expected to succeed, so we can remove this

src/configs.rs
```diff
@@ -47,3 +61,66 @@ pub const FIXED_SEED: [u8; 32] = [
...
+        }
+        Err(e) => {
+            logging::log_warning(format!(
+                "Invalid value for bloom-fp-rate: {}, error: {}",
+                fp_rate_str, e
+            ));
```
KarthikSubbarao: Same here. Let's return an error message

src/configs.rs
nnmehta: Will do

src/configs.rs
nnmehta: Will do that

nnmehta: Will be adding tests as Karthik suggested

src/configs.rs
nnmehta: I will throw error instead of just logging.

tests/test_replication.py
```diff
@@ -161,6 +161,8 @@ def test_deterministic_replication(self):
         # replicated with the properties below.
         assert self.client.execute_command('CONFIG SET bf.bloom-capacity 1000') == b'OK'
         assert self.client.execute_command('CONFIG SET bf.bloom-expansion 3') == b'OK'
+        assert self.client.execute_command('CONFIG SET bf.bloom-fp-rate 0.1') == b'OK'
+        assert self.client.execute_command('CONFIG SET bf.bloom-tightening-ratio 0.75') == b'OK'
         # Test bloom object creation with every command type.
         bloom_write_cmds = [
```
KarthikSubbarao, line 167: Can we add a new line at 186 and then test config get on the replica? This will validate that despite the replica having a different set of global configs, bloom object creation will use the replicated properties.
```
        assert self.replicas[0].client.execute_command('CONFIG GET bf.bloom-capacity')[1] == b'100'
        assert self.replicas[0].client.execute_command('CONFIG GET bf.bloom-expansion')[1] == b'2'
        assert self.replicas[0].client.execute_command('CONFIG GET bf.bloom-fp-rate')[1] == b'0.01'
        assert self.replicas[0].client.execute_command('CONFIG GET bf.bloom-tightening-ratio)[1] == b'0.5'
```

tests/test_basic.py
```diff
@@ -299,6 +299,22 @@ def test_debug_cmd(self):
         else:
             madd_scenario_object_digest == madd_default_object_digest
 
+        # validates that digest differs on bloom objects after changing the tightening_ratio
+        client.execute_command('BF.INSERT tightening_ratio ERROR 0.001 CAPACITY 1000 ITEMS item1')
```
KarthikSubbarao: Nit: For a simpler example, can we use `BF.RESERVE` without any item? Let's use BF.RESERVE for every bloom object creation from line 302 to 316

src/configs.rs
```diff
@@ -31,6 +39,13 @@ lazy_static! {
         AtomicI64::new(BLOOM_MEMORY_LIMIT_PER_FILTER_DEFAULT);
     pub static ref BLOOM_USE_RANDOM_SEED: AtomicBool = AtomicBool::default();
     pub static ref BLOOM_DEFRAG: AtomicBool = AtomicBool::new(BLOOM_DEFRAG_DEAFULT);
+    pub static ref BLOOM_FP_RATE_DEFAULT_STRING: String = BLOOM_FP_RATE_DEFAULT.to_string();
+    pub static ref BLOOM_FP_RATE: ValkeyGILGuard<ValkeyString> = ValkeyGILGuard::new(
+        ValkeyString::create(None, BLOOM_FP_RATE_DEFAULT.to_string())
+    );
+    pub static ref TIGHTENING_RATIO_DEFAULT_STRING: String = TIGHTENING_RATIO.to_string();
+    pub static ref TIGHTENING_RATIO_CONFIG: ValkeyGILGuard<ValkeyString> =
```
KarthikSubbarao, line 47: Let's rename this to `TIGHTENING_RATIO`

src/configs.rs
```diff
@@ -47,3 +62,41 @@ pub const FIXED_SEED: [u8; 32] = [
     89, 15, 245, 34, 234, 120, 17, 218, 167, 20, 216, 9, 59, 62, 123, 217, 29, 137, 138, 115, 62,
     152, 136, 135, 48, 127, 151, 205, 40, 7, 51, 131,
 ];
+
+pub fn on_string_config_set(
```
KarthikSubbarao, line 66: Can we add documentation above this line using document type comments?
```
/// This is a config set handler for the False Positive Rate and Tightening Ratio configs.
```

src/configs.rs
```diff
@@ -47,3 +62,41 @@ pub const FIXED_SEED: [u8; 32] = [
...
+    config_ctx: &ConfigurationContext,
+    name: &str,
+    val: &'static ValkeyGILGuard<ValkeyString>,
+) {
```
KarthikSubbarao: Change this to return an error... Right now this is not the config set handler. With the current syntax, it is the on config change handler.

```suggestion
) -> Result<(), ValkeyError> {
```

src/configs.rs
KarthikSubbarao, line 75: Can we add documentation above this line using document type comments?


```suggestion
/// This is a config set handler for the False Positive Rate and Tightening Ratio configs.
pub fn on_string_config_set(
```

src/configs.rs
KarthikSubbarao:
```suggestion
    pub static ref BLOOM_TIGHTENING_RATIO: ValkeyGILGuard<ValkeyString> =
```

src/configs.rs
```diff
@@ -47,3 +62,41 @@ pub const FIXED_SEED: [u8; 32] = [
...
+    let v = val.get(config_ctx);
+    let value_str = v.to_string_lossy();
+
+    let value = match value_str.parse::<f64>() {
+        Ok(v) => v,
+        Err(_) => {
+            let _: Result<(), ValkeyError> = Err(ValkeyError::Str("Invalid floating-point value"));
+            return;
```
KarthikSubbarao: We need to return an error...

```suggestion
            return Err(ValkeyError::Str("Invalid value: Unable to parse string as f64."));
```

src/configs.rs
```diff
@@ -47,3 +62,41 @@ pub const FIXED_SEED: [u8; 32] = [
...
+        "bloom-tightening-ratio" => {
+            if !(TIGHTENING_RATIO_MIN..=TIGHTENING_RATIO_MAX).contains(&value) {
+                let _: Result<(), ValkeyError> = Err(ValkeyError::Str(utils::BAD_ERROR_RATIO));
+                return;
+            }
+            TIGHTENING_RATIO_CONFIG.set(config_ctx, v.clone());
+        }
+        _ => {
+            let _: Result<(), ValkeyError> =
+                Err(ValkeyError::Str("Unknown configuration parameter"));
```
KarthikSubbarao: We need to return an error..
```suggestion
            return Err(ValkeyError::Str("Unknown configuration parameter"));
```

src/configs.rs
```diff
@@ -1,5 +1,13 @@
+use crate::bloom::utils;
```
KarthikSubbarao, line 1: We need to use the configs in command_handler.rs 

Can you update every mutative command in the file above to use the new f64 global configs instead of constants?

tests/test_basic.py
```diff
@@ -299,6 +299,22 @@ def test_debug_cmd(self):
...
+        assert self.client.execute_command('CONFIG SET bf.bloom-tightening-ratio 0.75') == b'OK'
+        client.execute_command('BF.INSERT tightening_ratio2 ERROR 0.01 CAPACITY 1000 ITEMS item1')
+        scenario_tightening_ratio_object_digest = client.execute_command('DEBUG DIGEST-VALUE tightening_ratio')
+        scenario_tightening_ratio2_digest = client.execute_command('DEBUG DIGEST-VALUE tightening_ratio2')
+        assert scenario_tightening_ratio_object_digest != scenario_tightening_ratio2_digest
+
+        # validates that digest differs on bloom objects after changing the fp_rate
+        client.execute_command('BF.INSERT fp_rate ITEMS item1')
+        assert self.client.execute_command('CONFIG SET bf.bloom-fp-rate 0.5') == b'OK'
+        client.execute_command('BF.INSERT fp_rate2 ERROR 0.1 ITEMS item1')
```
KarthikSubbarao:
```suggestion
        client.execute_command('BF.RESERVE fp_rate2 0.001 1000')
```

tests/test_basic.py
```diff
@@ -299,6 +299,22 @@ def test_debug_cmd(self):
...
+        # validates that digest differs on bloom objects after changing the tightening_ratio
+        client.execute_command('BF.INSERT tightening_ratio ERROR 0.001 CAPACITY 1000 ITEMS item1')
+        assert self.client.execute_command('CONFIG SET bf.bloom-tightening-ratio 0.75') == b'OK'
+        client.execute_command('BF.INSERT tightening_ratio2 ERROR 0.01 CAPACITY 1000 ITEMS item1')
+        scenario_tightening_ratio_object_digest = client.execute_command('DEBUG DIGEST-VALUE tightening_ratio')
+        scenario_tightening_ratio2_digest = client.execute_command('DEBUG DIGEST-VALUE tightening_ratio2')
+        assert scenario_tightening_ratio_object_digest != scenario_tightening_ratio2_digest
+
+        # validates that digest differs on bloom objects after changing the fp_rate
+        client.execute_command('BF.INSERT fp_rate ITEMS item1')
```
KarthikSubbarao:
```suggestion
        client.execute_command('BF.RESERVE fp_rate 0.001 1000)
```

tests/test_basic.py
```diff
@@ -299,6 +299,22 @@ def test_debug_cmd(self):
         else:
             madd_scenario_object_digest == madd_default_object_digest
 
+        # validates that digest differs on bloom objects after changing the tightening_ratio
+        client.execute_command('BF.INSERT tightening_ratio ERROR 0.001 CAPACITY 1000 ITEMS item1')
+        assert self.client.execute_command('CONFIG SET bf.bloom-tightening-ratio 0.75') == b'OK'
+        client.execute_command('BF.INSERT tightening_ratio2 ERROR 0.01 CAPACITY 1000 ITEMS item1')
```
KarthikSubbarao:
```suggestion
        client.execute_command('BF.RESERVE tightening_ratio2 0.001 1000')
```

tests/test_basic.py
KarthikSubbarao:
```suggestion
        client.execute_command('BF.RESERVE tightening_ratio 0.001 1000')
```

src/lib.rs
```diff
@@ -106,6 +108,8 @@ valkey_module! {
...
+            ["bloom-fp-rate", &*configs::BLOOM_FP_RATE, &*configs::BLOOM_FP_RATE_DEFAULT_STRING, ConfigurationFlags::DEFAULT, Some(Box::new(configs::on_string_config_set))],
```
KarthikSubbarao: This is not implementing a config set handler. It is implementing a config on change handler. This is how you implement a config set handler:
```suggestion
            ["bloom-fp-rate", &*configs::BLOOM_FP_RATE, &*configs::BLOOM_FP_RATE_DEFAULT_STRING, ConfigurationFlags::DEFAULT, None, Some(Box::new(configs::on_string_config_set))],
```

src/lib.rs
```diff
@@ -106,6 +108,8 @@ valkey_module! {
...
+            ["bloom-tightening-ratio", &*configs::TIGHTENING_RATIO_CONFIG, &*configs::TIGHTENING_RATIO_DEFAULT_STRING, ConfigurationFlags::DEFAULT, Some(Box::new(configs::on_string_config_set))],
```
KarthikSubbarao: Currently, this is not implementing a config set handler. It is implementing a config on change handler. This is how you implement a config set handler:
```suggestion
            ["bloom-tightening-ratio", &*configs::TIGHTENING_RATIO_CONFIG, &*configs::TIGHTENING_RATIO_DEFAULT_STRING, ConfigurationFlags::DEFAULT, None, Some(Box::new(configs::on_string_config_set))],
```

src/configs.rs
```diff
@@ -31,6 +39,13 @@ lazy_static! {
         AtomicI64::new(BLOOM_MEMORY_LIMIT_PER_FILTER_DEFAULT);
     pub static ref BLOOM_USE_RANDOM_SEED: AtomicBool = AtomicBool::default();
     pub static ref BLOOM_DEFRAG: AtomicBool = AtomicBool::new(BLOOM_DEFRAG_DEAFULT);
+    pub static ref BLOOM_FP_RATE_DEFAULT_STRING: String = BLOOM_FP_RATE_DEFAULT.to_string();
```
KarthikSubbarao: Can you add an actual f64 Mutex above this? We need to add this and change the f64 mutex.

```suggestion
    pub static ref BLOOM_FP_RATE_F64: Mutex<f64> = Mutex::new(BLOOM_FP_RATE_DEFAULT);
```

You also don't need `BLOOM_FP_RATE_DEFAULT`. We can use to_string in the config macro during creation

src/configs.rs
```diff
@@ -31,6 +39,13 @@ lazy_static! {
...
+    pub static ref BLOOM_FP_RATE: ValkeyGILGuard<ValkeyString> = ValkeyGILGuard::new(
+        ValkeyString::create(None, BLOOM_FP_RATE_DEFAULT.to_string())
+    );
+    pub static ref TIGHTENING_RATIO_DEFAULT_STRING: String = TIGHTENING_RATIO.to_string();
```
KarthikSubbarao: Can you add an actual f64 Mutex above this? We need to add this and change the f64 mutex.

```suggestion
    pub static ref BLOOM_TIGHTENING_F64: Mutex<f64> = Mutex::new(TIGHTENING_RATIO);
```

You also don't need `TIGHTENING_RATIO_DEFAULT_STRING `. We can use to_string in the config macro during creation

src/configs.rs
```diff
@@ -47,3 +62,41 @@ pub const FIXED_SEED: [u8; 32] = [
...
+        }
+    };
+
+    match name {
+        "bloom-fp-rate" => {
+            if !(BLOOM_FP_RATE_MIN..=BLOOM_FP_RATE_MAX).contains(&value) {
+                let _: Result<(), ValkeyError> = Err(ValkeyError::Str(utils::BAD_ERROR_RATE));
+                return;
+            }
+            BLOOM_FP_RATE.set(config_ctx, v.clone());
```
KarthikSubbarao: We need to change the f64 config here...
```suggestion
        let mut fp_rate = BLOOM_FP_RATE_F64.lock().expect("We expect the fp_rate static to exist.");
        *fp_rate = value;
```

src/configs.rs
```diff
@@ -47,3 +62,41 @@ pub const FIXED_SEED: [u8; 32] = [
...
+        }
+        "bloom-tightening-ratio" => {
+            if !(TIGHTENING_RATIO_MIN..=TIGHTENING_RATIO_MAX).contains(&value) {
+                let _: Result<(), ValkeyError> = Err(ValkeyError::Str(utils::BAD_ERROR_RATIO));
+                return;
+            }
+            TIGHTENING_RATIO_CONFIG.set(config_ctx, v.clone());
```
KarthikSubbarao: We need to change the f64 config here...
```suggestion
        let mut tightening = BLOOM_TIGHTENING_RATIO.lock().expect("We expect the tightening_ratio static to exist.");
        *tightening = value;
```

src/lib.rs
KarthikSubbarao: @nnmehta - We are missing test case 1.

Can we add a new test in test_basic.py to ensure where you perform config sets on every config and then config gets to make sure  the configs have been applied? Here, we can also try setting with invalid values and ensure they are rejected

nnmehta: Sure, I will do that and add these tests

src/configs.rs
nnmehta: If I just directly use BLOOM_FP_RATE_DEFAULT rather than using the string, it throws this error

```
81  | / valkey_module! {
82  | |     name: MODULE_NAME,
83  | |     version: 1,
84  | |     allocator: (valkey_module::alloc::ValkeyAlloc, valkey_module::alloc::ValkeyAlloc),
...   |
111 | |             ["bloom-fp-rate", &*configs::BLOOM_FP_RATE, &*configs::BLOOM_FP_RATE_DEFAULT.to_string(), ConfigurationFlags::DEFAULT, None, Some(Box::new(configs::on_string_config_set))],
    | |                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creates a temporary value which is freed while still in use
...   |
121 | |     ]
122 | | }
    | | -
    | | |
    | |_temporary value is freed at the end of this statement
    |   borrow later stored here
```

KarthikSubbarao: Can you change this `&*configs::BLOOM_FP_RATE_DEFAULT.to_string()` to a string slice? The macro requires a string slice type for the default value

src/bloom/utils.rs
```diff
@@ -326,7 +326,8 @@ impl BloomFilterType {
 
     /// Calculate the false positive rate for the Nth filter using tightening ratio.
     pub fn calculate_fp_rate(fp_rate: f64, num_filters: i32) -> Result<f64, BloomError> {
-        match fp_rate * configs::TIGHTENING_RATIO.powi(num_filters) {
+        let tightening_ratio = *configs::BLOOM_TIGHTENING_F64.lock().unwrap();
```
KarthikSubbarao: Can we change this function to accept a tightening_ratio parameter?

We should not be using a constant or config here to find the tightening ratio. We should be using what is in the bloom object itself - because tightening ratio is specific to the object at the time of its creation. 

There are two places where this function is used (1) `add_item` (2) `load_from_rdb`. We need to update how this function is called from both places.

For `add_item`, provide the  tightening ratio from the BloomFilterType structure.

For `load_from_rdb`, provide the  tightening ratio from the f64  loaded from RDB .

src/configs.rs
```diff
@@ -31,12 +39,22 @@ lazy_static! {
         AtomicI64::new(BLOOM_MEMORY_LIMIT_PER_FILTER_DEFAULT);
     pub static ref BLOOM_USE_RANDOM_SEED: AtomicBool = AtomicBool::default();
     pub static ref BLOOM_DEFRAG: AtomicBool = AtomicBool::new(BLOOM_DEFRAG_DEAFULT);
+    pub static ref BLOOM_FP_RATE_F64: Mutex<f64> = Mutex::new(BLOOM_FP_RATE_DEFAULT);
+    pub static ref BLOOM_FP_RATE_DEFAULT_STRING: String = BLOOM_FP_RATE_DEFAULT.to_string();
+    pub static ref BLOOM_FP_RATE: ValkeyGILGuard<ValkeyString> = ValkeyGILGuard::new(
+        ValkeyString::create(None, BLOOM_FP_RATE_DEFAULT.to_string())
+    );
+    pub static ref BLOOM_TIGHTENING_F64: Mutex<f64> = Mutex::new(TIGHTENING_RATIO_DEFAULT);
+    pub static ref TIGHTENING_RATIO_DEFAULT_STRING: String = TIGHTENING_RATIO_DEFAULT.to_string();
```
KarthikSubbarao, line 48: We can remove this

src/bloom/utils.rs
nnmehta: Okay, sure.

I will be changing the implementation as the following diff:

```
--- a/src/bloom/data_type.rs
+++ b/src/bloom/data_type.rs
@@ -87,15 +87,16 @@ impl ValkeyDataType for BloomFilterType {
             let Ok(capacity) = raw::load_unsigned(rdb) else {
                 return None;
             };
-            let new_fp_rate = match Self::calculate_fp_rate(fp_rate, num_filters as i32) {
-                Ok(rate) => rate,
-                Err(_) => {
-                    logging::log_warning(
-                        "Failed to restore bloom object: Reached max number of filters",
-                    );
-                    return None;
-                }
-            };
+            let new_fp_rate =
+                match Self::calculate_fp_rate(fp_rate, num_filters as i32, tightening_ratio) {
+                    Ok(rate) => rate,
+                    Err(_) => {
+                        logging::log_warning(
+                            "Failed to restore bloom object: Reached max number of filters",
+                        );
+                        return None;
+                    }
+                };
             if !BloomFilter::validate_size(capacity as i64, new_fp_rate) {
                 logging::log_warning("Failed to restore bloom object: Contains a filter larger than the max allowed size limit.");
                 return None;
diff --git a/src/bloom/utils.rs b/src/bloom/utils.rs
index d8f2556..d808fdc 100644
--- a/src/bloom/utils.rs
+++ b/src/bloom/utils.rs
@@ -265,10 +265,11 @@ impl BloomFilterType {
             }
             // Scale out by adding a new filter with capacity bounded within the u32 range. false positive rate is also
             // bound within the range f64::MIN_POSITIVE <= x < 1.0.
-            let new_fp_rate = match Self::calculate_fp_rate(self.fp_rate, num_filters) {
-                Ok(rate) => rate,
-                Err(e) => return Err(e),
-            };
+            let new_fp_rate =
+                match Self::calculate_fp_rate(self.fp_rate, num_filters, self.tightening_ratio) {
+                    Ok(rate) => rate,
+                    Err(e) => return Err(e),
+                };
             let new_capacity = match filter.capacity.checked_mul(self.expansion.into()) {
                 Some(new_capacity) => new_capacity,
                 None => {
@@ -325,8 +326,12 @@ impl BloomFilterType {
     }
 
     /// Calculate the false positive rate for the Nth filter using tightening ratio.
-    pub fn calculate_fp_rate(fp_rate: f64, num_filters: i32) -> Result<f64, BloomError> {
-        let tightening_ratio = *configs::BLOOM_TIGHTENING_F64.lock().unwrap();
+    pub fn calculate_fp_rate(
+        fp_rate: f64,
+        num_filters: i32,
+        tightening_ratio: f64,
+    ) -> Result<f64, BloomError> {
         match fp_rate * tightening_ratio.powi(num_filters) {
             x if x > f64::MIN_POSITIVE => Ok(x),
             _ => Err(BloomError::MaxNumScalingFilters),
```

src/configs.rs
nnmehta: So do you want me to use this directly:
```
["bloom-fp-rate", &*configs::BLOOM_FP_RATE, "0.01", ConfigurationFlags::DEFAULT, None, Some(Box::new(configs::on_string_config_set))],
["bloom-tightening-ratio", &*configs::BLOOM_TIGHTENING_RATIO, "0.5", ConfigurationFlags::DEFAULT, None, Some(Box::new(configs::on_string_config_set))],
```

and avoid using to_string and the string constant we have in configs?

tests/test_basic.py
```diff
@@ -299,6 +299,25 @@ def test_debug_cmd(self):
         else:
             madd_scenario_object_digest == madd_default_object_digest
 
+        # validates that digest differs on bloom objects after changing the tightening_ratio
```
KarthikSubbarao:
```suggestion
        # scenario 6 validates that digest differs on bloom objects after changing the tightening_ratio config
```

tests/test_basic.py
```diff
@@ -299,6 +299,25 @@ def test_debug_cmd(self):
...
+        client.execute_command('BF.RESERVE tightening_ratio 0.001 1000')
+        assert self.client.execute_command('CONFIG SET bf.bloom-tightening-ratio 0.75') == b'OK'
+        client.execute_command('BF.RESERVE tightening_ratio2 0.001 1000')
+        scenario_tightening_ratio_object_digest = client.execute_command('DEBUG DIGEST-VALUE tightening_ratio')
+        scenario_tightening_ratio2_digest = client.execute_command('DEBUG DIGEST-VALUE tightening_ratio2')
+        assert scenario_tightening_ratio_object_digest != scenario_tightening_ratio2_digest
+
+        # validates that digest differs on bloom objects after changing the fp_rate
```
KarthikSubbarao:
```suggestion
        # scenario 7 validates that digest differs on bloom objects after changing the fp_rate config
```

tests/test_basic.py
```diff
@@ -299,6 +299,25 @@ def test_debug_cmd(self):
...
+        client.execute_command('BF.RESERVE fp_rate 0.001 1000')
+        assert self.client.execute_command('CONFIG SET bf.bloom-fp-rate 0.5') == b'OK'
+        client.execute_command('BF.RESERVE fp_rate2 0.001 1000')
```
KarthikSubbarao: We can use a BF.INSERT here without specifying the error rate in both cases

tests/test_basic.py
```diff
@@ -323,3 +342,25 @@ def test_bloom_wrong_type(self):
...
+            assert self.client.execute_command('CONFIG SET bf.bloom-tightening-ratio 0.75') == b'OK'
+            
+            assert self.client.execute_command('CONFIG GET bf.bloom-fp-rate')[1] == b'0.1'
+            assert self.client.execute_command('CONFIG GET bf.bloom-tightening-ratio')[1] == b'0.75'
+            try:
+                self.client.execute_command('CONFIG SET bf.bloom-fp-rate 1.1')
+                self.client.execute_command('CONFIG SET bf.bloom-tightening-ratio 1.75')
+                assert False
+            except ResponseError as e:
+                pass
```
KarthikSubbarao: Can we validate the error here?

src/configs.rs
```diff
@@ -31,12 +39,20 @@ lazy_static! {
...
+    pub static ref BLOOM_TIGHTENING_F64: Mutex<f64> = Mutex::new(TIGHTENING_RATIO_DEFAULT);
+    pub static ref BLOOM_TIGHTENING_RATIO: ValkeyGILGuard<ValkeyString> = ValkeyGILGuard::new(
+        ValkeyString::create(None, TIGHTENING_RATIO_DEFAULT.to_string())
+    );
 }
 
 /// Constants
 // Tightening ratio used during scale out for the calculation of fp_rate of every new filter within a bloom object to
 // maintain the bloom object's overall fp_rate to the configured value.
-pub const TIGHTENING_RATIO: f64 = 0.5;
```
KarthikSubbarao, line 39: We can change this to a string if it is not used anywhere. It will help with config creation

src/configs.rs
```diff
@@ -31,12 +39,20 @@ lazy_static! {
...
+pub const TIGHTENING_RATIO_DEFAULT: f64 = 0.5;
```
KarthikSubbarao: We can change this to a string if it is not used anywhere. It will help with config creation

src/bloom/command_handler.rs
```diff
@@ -444,8 +450,12 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
     let filter_name = &input_args[idx];
     idx += 1;
     let replicated_cmd = ctx.get_flags().contains(ContextFlags::REPLICATED);
-    let mut fp_rate = configs::BLOOM_FP_RATE_DEFAULT;
-    let mut tightening_ratio = configs::TIGHTENING_RATIO;
+    let mut fp_rate = *configs::BLOOM_FP_RATE_F64
+        .lock()
+        .expect("Failed to lock fp_rate");
```
KarthikSubbarao:
```suggestion
        .expect("Unable to get a lock on fp_rate static");
```

src/configs.rs
```diff
@@ -31,12 +39,20 @@ lazy_static! {
         AtomicI64::new(BLOOM_MEMORY_LIMIT_PER_FILTER_DEFAULT);
     pub static ref BLOOM_USE_RANDOM_SEED: AtomicBool = AtomicBool::default();
     pub static ref BLOOM_DEFRAG: AtomicBool = AtomicBool::new(BLOOM_DEFRAG_DEAFULT);
+    pub static ref BLOOM_FP_RATE_F64: Mutex<f64> =
+        Mutex::new(BLOOM_FP_RATE_DEFAULT.parse::<f64>().unwrap());
```
KarthikSubbarao: Let's use expect instead of unwrap
```suggestion
        Mutex::new(BLOOM_FP_RATE_DEFAULT.parse::<f64>().expect("Expected valid f64 for fp rate."));
```

src/configs.rs
```diff
@@ -31,12 +39,20 @@ lazy_static! {
...
+    pub static ref BLOOM_FP_RATE: ValkeyGILGuard<ValkeyString> =
+        ValkeyGILGuard::new(ValkeyString::create(None, BLOOM_FP_RATE_DEFAULT));
+    pub static ref BLOOM_TIGHTENING_F64: Mutex<f64> =
+        Mutex::new(TIGHTENING_RATIO_DEFAULT.parse::<f64>().unwrap());
```
KarthikSubbarao:
```suggestion
        Mutex::new(TIGHTENING_RATIO_DEFAULT.parse::<f64>().expect("Expected valid f64 for tightening ratio."));
```

src/configs.rs
```diff
@@ -47,3 +63,44 @@ pub const FIXED_SEED: [u8; 32] = [
...
+    let value = match value_str.parse::<f64>() {
+        Ok(v) => v,
+        Err(_) => {
+            return Err(ValkeyError::Str("Invalid floating-point value"));
+        }
+    };
+
+    match name {
+        "bloom-fp-rate" => {
+            if !(BLOOM_FP_RATE_MIN..=BLOOM_FP_RATE_MAX).contains(&value) {
```
KarthikSubbarao: This will allow invalid values for fp rate (e.g 0.0 and 1.0) which should be rejected instead.

Can you change this to something like this?
```suggestion
            if !(BLOOM_FP_RATE_MIN > value  && value  < BLOOM_FP_RATE_MAX) {
```

src/configs.rs
```diff
@@ -47,3 +63,44 @@ pub const FIXED_SEED: [u8; 32] = [
...
+                return Err(ValkeyError::Str(utils::ERROR_RATE_RANGE));
+            }
+            let mut fp_rate = BLOOM_FP_RATE_F64
+                .lock()
+                .expect("We expect the fp_rate static to exist.");
+            *fp_rate = value;
+            Ok(())
+        }
+        "bloom-tightening-ratio" => {
+            if !(TIGHTENING_RATIO_MIN..=TIGHTENING_RATIO_MAX).contains(&value) {
```
KarthikSubbarao: This will allow invalid values for tightening ratio (e.g 0.0 and 1.0) which should be rejected instead.

Can you change this to something like this?
```suggestion
            if !(TIGHTENING_RATIO_MIN > value  && value  < TIGHTENING_RATIO_MAX) {
```

src/configs.rs
```diff
@@ -31,12 +39,20 @@ lazy_static! {
...
+    pub static ref BLOOM_TIGHTENING_RATIO: ValkeyGILGuard<ValkeyString> =
+        ValkeyGILGuard::new(ValkeyString::create(None, TIGHTENING_RATIO_DEFAULT));
 }
 
 /// Constants
 // Tightening ratio used during scale out for the calculation of fp_rate of every new filter within a bloom object to
 // maintain the bloom object's overall fp_rate to the configured value.
-pub const TIGHTENING_RATIO: f64 = 0.5;
+pub const TIGHTENING_RATIO_DEFAULT: &str = "0.5";
 pub const TIGHTENING_RATIO_MIN: f64 = 0.0;
```
KarthikSubbarao:
```suggestion
pub const BLOOM_TIGHTENING_RATIO_MIN: f64 = 0.0;
```

src/configs.rs
```diff
@@ -31,12 +39,20 @@ lazy_static! {
...
 pub const TIGHTENING_RATIO_MAX: f64 = 1.0;
```
KarthikSubbarao:
```suggestion
pub const BLOOM_TIGHTENING_RATIO_MAX: f64 = 1.0;
```

src/configs.rs
```diff
@@ -47,3 +63,44 @@ pub const FIXED_SEED: [u8; 32] = [
...
+    let v = val.get(config_ctx);
+    let value_str = v.to_string_lossy();
+
+    let value = match value_str.parse::<f64>() {
+        Ok(v) => v,
+        Err(_) => {
+            return Err(ValkeyError::Str("Invalid floating-point value"));
+        }
+    };
+
```
KarthikSubbarao, line 82: We can remove the extra new lines here to help with readability

tests/test_basic.py
```diff
@@ -323,3 +339,17 @@ def test_bloom_wrong_type(self):
             except Exception as e:
                 
                 assert str(e) == f"WRONGTYPE Operation against a key holding the wrong kind of value"
+    
+    #validates config set correctly and invalid values are rejected
+    def test_on_string_config_set(self):
```
KarthikSubbarao:
```suggestion
    def test_bloom_string_config_set(self):
        """
            This is a test that validates the bloom string configuration set logic.
        """
```

src/configs.rs
KarthikSubbarao: We can remove the extra new lines here to help with readability
```suggestion
    let value = match value_str.parse::<f64>() {
        Ok(v) => v,
        Err(_) => {
            return Err(ValkeyError::Str("Invalid floating-point value"));
        }
    };
```

src/configs.rs
KarthikSubbarao: Can we move this block of code up to line 24? This is no longer a constant - it is a configuration, so we can include it in the section of configurations

src/configs.rs
nnmehta: I will be removing = to for the range i.e. use
```
if !(BLOOM_TIGHTENING_RATIO_MIN..BLOOM_TIGHTENING_RATIO_MAX).contains(&value) {
```

When I use this
```
 if !(BLOOM_TIGHTENING_RATIO_MIN > value && value < BLOOM_TIGHTENING_RATIO_MAX) {
```
it throws the following error
```
error: right-hand side of `&&` operator has no effect
   --> src/configs.rs:101:17
    |
101 |             if !(BLOOM_TIGHTENING_RATIO_MIN > value && value < BLOOM_TIGHTENING_RATIO_MAX) {
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: `if `BLOOM_TIGHTENING_RATIO_MIN > value` evaluates to true, value < BLOOM_TIGHTENING_RATIO_MAX` will always evaluate to true as well
   --> src/configs.rs:101:53
    |
101 |             if !(BLOOM_TIGHTENING_RATIO_MIN > value && value < BLOOM_TIGHTENING_RATIO_MAX) {
```

src/configs.rs
nnmehta: I will be removing = to for the range i.e. use
```
if !(BLOOM_FP_RATE_MIN..BLOOM_FP_RATE_MAX).contains(&value) {
```

When I use this
```
 if !(BLOOM_FP_RATE_MIN > value && value < BLOOM_FP_RATE_MAX) {
```
it throws the following error
```
error: right-hand side of `&&` operator has no effect
  --> src/configs.rs:91:17
   |
91 |             if !(BLOOM_FP_RATE_MIN > value && value < BLOOM_FP_RATE_MAX) {
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: `if `BLOOM_FP_RATE_MIN > value` evaluates to true, value < BLOOM_FP_RATE_MAX` will always evaluate to true as well
  --> src/configs.rs:91:44
```

src/configs.rs
nnmehta: Changed

src/configs.rs
nnmehta: Changed

src/configs.rs
nnmehta: Removed

src/configs.rs
nnmehta: Will move above

YueTang-Vanessa (approved):

src/configs.rs
KarthikSubbarao: This check is incorrect as it allows 0.0 which is an invalid fp rate. 


Correct check:
```
if !(value > BLOOM_TIGHTENING_RATIO_MIN  && value < BLOOM_TIGHTENING_RATIO_MAX) {
 ```

PR #32, opened 2024-12-10
Deterministic replication of bloom object creations

KarthikSubbarao:
With this PR, the following changes are made:
1. Every bloom object creation is replicated with a custom created BF.INSERT command that includes every bloom property that can be a result of global default configs on the primary. With this, the exact same bloom object created on the primary will be created on the replica. Replicated properties include: Capacity, False Positive Rate, Tightening Ratio, Seed (used by hash functions), Items (incase of a object creation + item add operation).
2. Refactored BloomFilterType and BloomFilter structures to not have public exposed structure members. This is accessible from getter methods now.
3. Updated the DIGEST logic to include the `is_seed_random` bool property which indicates whether the bloom object is using a fixed seed or random seed. 
4. Added python integration testing for validation of deterministic replication creation behavior wise. Also, when the primary node and replica nodes have different global configs, tests were added to validate the digest of the bloom object created from replicated command matches the object on the primary node

### DIFF
--- a/src/bloom/command_handler.rs
+++ b/src/bloom/command_handler.rs
@@ -11,6 +11,9 @@ use valkey_module::ContextFlags;
 use valkey_module::NotifyEvent;
 use valkey_module::{Context, ValkeyError, ValkeyResult, ValkeyString, ValkeyValue, VALKEY_OK};
 
+/// Helper function used to add items to a bloom object. It handles both multi item and single item add operations.
+/// It is used by any command that allows adding of items: BF.ADD, BF.MADD, and BF.INSERT.
+/// Returns the result of the item add operation on success as a ValkeyValue and a ValkeyError on failure.
 fn handle_bloom_add(
     args: &[ValkeyString],
     argc: usize,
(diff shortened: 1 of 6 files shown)


src/wrapper/bloom_callback.rs
```diff
@@ -26,28 +26,28 @@ use super::defrag::Defrag;
...
     let v = &*value.cast::<BloomFilterType>();
-    raw::save_unsigned(rdb, v.filters.len() as u64);
-    raw::save_unsigned(rdb, v.expansion as u64);
-    raw::save_double(rdb, v.fp_rate);
-    raw::save_double(rdb, v.tightening_ratio);
+    raw::save_unsigned(rdb, v.num_filters() as u64);
+    raw::save_unsigned(rdb, v.expansion() as u64);
+    raw::save_double(rdb, v.fp_rate());
+    raw::save_double(rdb, v.tightening_ratio());
     let mut is_seed_random = 0;
```
zackcam: nit
`let is_seed_random = if is_seed_random() { 1 } else { 0 };`
Maybe something like this would be better?

src/bloom/command_handler.rs
```diff
@@ -304,9 +432,13 @@ pub fn bloom_filter_reserve(ctx: &Context, input_args: &[ValkeyString]) -> Valke
     }
 }
 
+/// Function that implements logic to handle the BF.INSERT command.
 pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
+    use valkey_module::logging;
```
zackcam: Do we want to log on insert?

src/bloom/data_type.rs
```diff
@@ -117,27 +117,27 @@ impl ValkeyDataType for BloomFilterType {
...
-        let item = BloomFilterType {
-            expansion: expansion as u32,
+        let item = BloomFilterType::from_existing(
+            expansion as u32,
             fp_rate,
             tightening_ratio,
             is_seed_random,
             filters,
-        };
-        item.bloom_filter_type_incr_metrics_on_new_create();
```
zackcam, line 128: As we now don't call `bloom_filter_type_incr_metrics_on_new_create` from here we can remove the pub part of `pub fn bloom_filter_type_incr_metrics_on_new_create`

src/wrapper/bloom_callback.rs
KarthikSubbarao: Updated

src/bloom/command_handler.rs
KarthikSubbarao: No, this was just added for testing. I forgot to remove it. Removed

src/bloom/data_type.rs
KarthikSubbarao: Updated

tests/test_replication.py
```diff
@@ -25,60 +25,74 @@ def use_random_seed_fixture(self, bloom_config_parameterization):
...
             self.waitForReplicaToSyncUp(self.replicas[0])
             assert self.replicas[0].client.execute_command('BF.EXISTS key item1') == 1
-
-            # Validate that the bloom object creation command and item add command was replicated.
-            expected_calls = test_case[3]
-            primary_cmd_stats = self.client.info("Commandstats")['cmdstat_' + prefix]
-            replica_cmd_stats = self.replicas[0].client.info("Commandstats")['cmdstat_' + prefix]
-            assert primary_cmd_stats["calls"] == expected_calls and replica_cmd_stats["calls"] == expected_calls
+            # Validate that item addition (not bloom creation) is using the original command
+            if prefix != 'BF.RESERVE':
```
zackcam, line 72: Would this be better as a function to reduce repeated code and make test easier to read?
Could pass in calls for both replica and primary I think.

KarthikSubbarao: Yes, I can do that.

zackcam (approved):
Looks good to me!

PR #33, opened 2024-12-11
Adding bloom defrag hits and misses to info, added check these metrics get incremented in defrag test

zackcam:
Added a new section in info specifically for defrag. Added two new fields in this section of defrag hits and misses that are incremented in bloom_callback.
Added a check in defrag test to make sure these metrics are actually incremented

### DIFF
--- a/tests/test_bloom_defrag.py
+++ b/tests/test_bloom_defrag.py
@@ -65,7 +65,8 @@ class TestBloomDefrag(ValkeyBloomTestCaseBase):
         assert float(memory_info_after_defrag.get('allocator_frag_ratio', 0)) < float(memory_info_non_defragged.get('allocator_frag_ratio', 0))
         # Check that items we added still exist in the respective bloom objects
         self.check_values_present(scale_names, num_items_inserted_per_object)
-
+        info_results = self.client.info_section("bf")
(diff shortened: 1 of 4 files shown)


tests/test_bloom_defrag.py
```diff
@@ -101,3 +103,17 @@ def check_values_present(self, scale_names, num_items_inserted_per_object):
             # All items should be present so we compare with an array of length num items where all items are 1
             expected_results = [1] * num_items
             assert results == expected_results, f"Unexpected results for scale {scale}: {results}"
+
+    def verfiy_bloom_defrag_metrics(self):
```
KarthikSubbarao: One suggestion here.

KarthikSubbarao: Actually, can we simply use this instead? 

```
    def info_obj(self):
        """Return a ValkeyInfo object for the current client info"""
        return ValkeyInfo(self.info('all'))
```

You can add a custom function to accept a custom section name (MODULE, BF etc) and return the result. The caller site can access what they want from
```
self.info[<custom-metric>]
```

KarthikSubbarao (approved):

YueTang-Vanessa (approved):
LGTM

YueTang-Vanessa (approved):

PR #34, opened 2024-12-14
Deflaking tests to remove chance of false positives causing tests to fail

zackcam:
Updated unit tests and integration tests to remove the chance of false positives causing the tests to fail. Before we would assert for each item of insert madd and mexists but due to false positive sometimes the output can be unexpected. For these outputs we now assert if they are equal to 1 or 0 and the length of the return is correct instead of the exact number we expect.
I also updated areas where we are checking bf.card as again false positives could cause this number to be lower than expected. In order to remedy this I changed where we test this command or I called a method that will add up until that number of items taking into account fp's.  
For the unit tests we could have a false positive in the add where that item thinks its already been added. To remedy this we now perform a loop until we try and add an item we don't think exists. This will still then check that we get the correct error while not having a small chance to fail on a fp.

Also removed some creations and commands that weren't used anywhere

### DIFF
--- a/src/bloom/utils.rs
+++ b/src/bloom/utils.rs
@@ -598,12 +598,14 @@ mod tests {
     /// Loops until the capacity of the provided bloom filter is reached and adds a new item to it in every iteration.
     /// The item name is rand_prefix + the index (starting from starting_item_idx).
     /// With every add operation, fp_count is tracked as we expect the add operation to return 1, since it is a new item.
+    /// There is an option to pass in an expected error and assert that we throw that error
     /// Returns the number of errors (false positives) and the final item index.
     fn add_items_till_capacity(
         bf: &mut BloomFilterType,
         capacity_needed: i64,
         starting_item_idx: i64,
         rand_prefix: &String,
+        expected_error: Option<BloomError>,
     ) -> (i64, i64) {
         let mut new_item_idx = starting_item_idx;
         let mut fp_count = 0;
@@ -616,14 +618,26 @@ mod tests {
                     fp_count += 1;
                 }
                 Ok(1) => {
+                    if let Some(err) = expected_error {
+                        panic!(
+                            "Expected error on the bloom object during during item add: {:?}",
+                            err
+                        );
+                    }
                     cardinality += 1;
                 }
                 Ok(i64::MIN..=-1_i64) | Ok(2_i64..=i64::MAX) => {
                     panic!("We do not expect add_item to return any Integer other than 0 or 1.")
                 }
-                Err(e) => {
-                    panic!("We do not expect add_item to throw errors on this scalable filter test, {:?}", e);
-                }
+                Err(e) => match &expected_error {
+                    Some(expected) => {
+                        assert_eq!(&e, expected, "Error doesn't match the expected error");
+                        break;
+                    }
+                    None => {
+                        panic!("Unexpected error when adding items: {:?}", e);
+                    }
+                },
             };
             new_item_idx += 1;
         }
@@ -781,10 +795,14 @@ mod tests {
         )
         .expect("Expect bloom creation to succeed");
         let (error_count, add_operation_idx) =
-            add_items_till_capacity(&mut bf, initial_capacity, 1, &rand_prefix);
-        assert_eq!(
-            bf.add_item(b"new_item", true),
-            Err(BloomError::NonScalingFilterFull)
+            add_items_till_capacity(&mut bf, initial_capacity, 1, &rand_prefix, None);
+        // Check adding to a full non scaling filter will throw an error
+        add_items_till_capacity(
+            &mut bf,
+            initial_capacity + 1,
+            1,
+            &rand_prefix,
+            Some(BloomError::NonScalingFilterFull),
         );
         assert_eq!(bf.capacity(), initial_capacity);
         assert_eq!(bf.cardinality(), initial_capacity);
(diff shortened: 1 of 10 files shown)


src/bloom/utils.rs
```diff
@@ -782,10 +818,11 @@ mod tests {
         .expect("Expect bloom creation to succeed");
         let (error_count, add_operation_idx) =
             add_items_till_capacity(&mut bf, initial_capacity, 1, &rand_prefix);
-        assert_eq!(
-            bf.add_item(b"new_item", true),
-            Err(BloomError::NonScalingFilterFull)
-        );
+        assert!(add_items_to_get_scaling_failure(
```
KarthikSubbarao: Can we instead re-use the same `add_items_till_capacity` by setting a capacity to be equal to `initial_capacity  + 1` and starting index to be equal to `add_operation_idx`?

src/bloom/utils.rs
```diff
@@ -595,6 +595,42 @@ mod tests {
...
 
+    fn add_items_to_get_scaling_failure(
+        bf: &mut BloomFilterType,
+        starting_item_idx: i64,
+        rand_prefix: &String,
+    ) -> bool {
+        let mut new_item_idx = starting_item_idx;
+        let mut fp_count = 0;
+        let mut found_new_key: bool = false;
+        // This loop could be bad do we want to have a time limit on it by counting number of fp_count?
```
KarthikSubbarao: Replied on the caller site

src/bloom/utils.rs
zackcam: The code for that panics when we get any error however we want an error in this return, the other problem i could see is if we get a return value of 1 we would want to panic as this means we added an item to a nonscaling filter that was full. My thinking is this could be fixed by passing in another variable called `expect_failure` and then if that was true add ifs based in this in those two relevant places. Does this seem like a better idea?

src/bloom/utils.rs
```diff
@@ -616,14 +617,27 @@ mod tests {
...
+                            // We expect an error to be thrown when trying to add to a full non scaling filter
+                            break;
+                        } else {
+                            panic!("We do not expect add_item to throw errors on a scalable filter, {:?}", e);
+                        }
+                    }
+                    _ => {
+                        panic!("Unexpected error when adding items: {:?}", e);
+                    }
+                },
```
KarthikSubbarao: An easier way might be to have to remove the bool `expect_scaling_failure` here and have this function accept an Option<BloomError>. If None, panic on any error. If Some(error), assert that the real error matches the provided (expected) error

This will make this test function more extendable to test other failure scenarios

tests/test_aofrewrite.py
```diff
@@ -78,4 +77,5 @@ def test_aofrewrite_bloomfilter_metrics(self):
 
         # Delete the scaled bloomfilter to check both filters are deleted and metrics stats are set accordingly
         self.client.execute_command('DEL key1')
+        wait_for_equal(lambda: self.client.execute_command('BF.EXISTS key1 item_prefix1'), 0)
         self.verify_bloom_metrics(self.client.execute_command("INFO bf"), 0, 0, 0, 0, 0)
```
KarthikSubbarao: We can add a newline if it is missing at the end of this file

tests/test_bloom_command.py
```diff
@@ -134,6 +128,10 @@ def test_bloom_command_behavior(self):
         for test_case in basic_behavior_test_case:
             cmd = test_case[0]
             expected_result = test_case[1]
```
KarthikSubbarao, line 130: Is this still `expected_result`? or has it changed to `capacty_to_fill`? If it is the latter, you can rename it according

KarthikSubbarao, line 132: This is not `expected_result` for every scenario any more.


Adding a comment above this to explain what it is varies based on the command - Multi Commands have it as the length (of 0s/1s expected)
Other commands have it as the literal expected result

tests/test_bloom_metrics.py
```diff
@@ -21,11 +21,11 @@ def test_basic_command_metrics(self):
         # Check that other commands don't influence metrics
         assert(self.client.execute_command('BF.EXISTS key item1') == 1)
         assert(self.client.execute_command('BF.ADD key item2') == 1)
-        assert(self.client.execute_command('BF.MADD key item3 item4')== [1, 1])
-        assert(self.client.execute_command('BF.MEXISTS key item3 item5')== [1, 0])
-        assert(self.client.execute_command('BF.CARD key') == 4)
+        assert(len(self.client.execute_command('BF.MADD key item3 item4')) == 2)
+        assert(len(self.client.execute_command('BF.MEXISTS key item3 item5')) == 2)
+        assert(self.client.execute_command('BF.CARD key') >= 1)
```
nnmehta: Should we also add an upper limit here? like <= 4 ?

src/bloom/utils.rs
```diff
@@ -616,14 +618,23 @@ mod tests {
                     fp_count += 1;
                 }
                 Ok(1) => {
+                    if let Some(BloomError::NonScalingFilterFull) = expected_error {
+                        panic!("A non scaling filter that is full shouldn't be able to have new items added successfully");
+                    }
```
KarthikSubbarao: If something like this is possible, it will be more generic: 

```suggestion
                    if let Some(err) = expected_error {
                        panic!("Expected error on the bloom object during during item add: {:?}", err);
                    }
```

tests/test_bloom_command.py
```diff
@@ -134,6 +128,11 @@ def test_bloom_command_behavior(self):
...
```
KarthikSubbarao, line 132:
```suggestion
            # For non multi commands, this is the verbatim expected result. 
            # For multi commands, test_case[1] contains the number of item add/exists results which are expected to be 0 or 1.
            expected_result = test_case[1]
```

PR #40, opened 2025-01-13
Adding a bloom command keyspace test

zackcam:
Created a python test that tests all the write commands and a read command expecting the read command to not trigger any keyspace notifications

### DIFF
--- a/tests/test_bloom_keyspace.py
+++ b/tests/test_bloom_keyspace.py
@@ -1,0 +1,70 @@
+import logging, time
+from valkey_bloom_test_case import ValkeyBloomTestCaseBase
+from valkeytests.conftest import resource_port_tracker
+
+class TestKeyEventNotifications(ValkeyBloomTestCaseBase):
+    RESERVE_KEYSPACE_MESSAGE = {'type': 'pmessage', 'pattern': b'__key*__:*', 'channel': b'__keyspace@0__:intermediate_val', 'data': b'bloom.reserve'}
+    RESERVE_KEYEVENT_MESSAGE = {'type': 'pmessage', 'pattern': b'__key*__:*', 'channel': b'__keyevent@0__:bloom.reserve', 'data': b'intermediate_val'}
+    ADD_KEYSPACE_MESSAGE = {'type': 'pmessage', 'pattern': b'__key*__:*', 'channel': b'__keyspace@0__:intermediate_val', 'data': b'bloom.add'} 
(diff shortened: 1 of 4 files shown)


tests/test_bloom_keyspace.py
```diff
@@ -0,0 +1,79 @@
...
+        while time.time() < timeout:
+            message = self.keyspace_client_subscribe.get_message()
+            if message:
+                if count > 0:
+                    messages.append(message)    
+                count = count + 1
+            time.sleep(0.01)
+        return messages
+    
+    def test_keyspace_bloom_add(self):
```
KarthikSubbarao: To reduce the number of tests (the number of servers started up on the host for testing), we can have a single test that loops through the commands and asserts accordingly. Can we try this out / is this possible?

tests/test_bloom_keyspace.py
```diff
@@ -0,0 +1,79 @@
...
+        self.check_response(result_messages, expected_messages)
+
+    def test_keyspace_bloom_reserve(self):
+        self.create_subscribe_clients()
+        expected_messages = self.create_expected_message_list(True, False)
+        result_messages = self.get_subscribe_client_messages(self.keyspace_client,'BF.RESERVE test 0.01 1000') 
+        self.check_response(result_messages, expected_messages)
+
+    def create_subscribe_clients(self):
+        self.keyspace_client =self.server.get_new_client()
```
KarthikSubbarao:
```suggestion
        self.keyspace_client = self.server.get_new_client()
```

tests/test_bloom_keyspace.py
```diff
@@ -0,0 +1,79 @@
...
+        count=0
+        messages =[]
+        timeout = time.time() + 5
+        while time.time() < timeout:
+            message = self.keyspace_client_subscribe.get_message()
+            if message:
+                if count > 0:
+                    messages.append(message)    
+                count = count + 1
+            time.sleep(0.01)
```
KarthikSubbarao: Can we try to avoid sleep here since that can make this test flaky?

tests/test_bloom_keyspace.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def check_response(self, result_messages, expected_messages):
+        assert len(expected_messages) == len(result_messages), f"Number of messages received differs, Expected messages: {expected_messages}, Messages received: {result_messages}"
+        for message in expected_messages:
+            assert message in result_messages, f"{message} was not found in messages received"
+
+    def get_subscribe_client_messages(self, client, cmd):
+        client.execute_command(cmd)
+        count=0
+        messages =[]
```
KarthikSubbarao:
```suggestion
        count = 0
        messages  = []
```

KarthikSubbarao (approved):

PR #41, opened 2025-01-17
Adding optional arg to BF.INSERT to allow users to check if their bloom filter can reach the desired size

zackcam:
# Overview
This PR is proposing to add an extra optional arg to the BF.INSERT command in order to allow users to know if their creation will be able to scale to what they desire. We will add an optional arg called “VALIDATESCALETO” and validate whether it is possible to achieve X capacity given the bloom object memory usage limit (by default 128MB), fp rate, Tightening, expansion, and scale outs. We also will expand BF.INFO as well to have a new output as well. This output will show what the max capacity of the specified bloom filter is. This field can also be received independently by specifying MAXSCALEDCAPACITY as an arg.
# Amending BF.INSERT command
Adding VALIDATESCALETO as an optional arg of BF.INSERT. This argument will allow the user to specify what they think their filter will scale up to. If the capacity that they think it will scale to isn't possible given the specifications provided then we will throw an error. New bf.insert command will look like:
```
BF.INSERT <key> [CAPACITY capacity] [ERROR fp_error] [EXPANSION expansion] [NOCREATE] [NONSCALING] [VALIDATESCALETO validate_scale_to] [ITEMS item]
```
Due to the nature of this optional argument, specifying NONSCALING and ATLEASTCAPACITY isn't allowed as this new argument is only useful for seeing if a filter can scale to the desired capacity.
There are three new error messages associated with this:
```
pub const VALIDATE_SCALE_TO_EXCEEDS_MAX_SIZE: &str =
    "ERR provided VALIDATESCALETO causes bloom object to exceed memory limit";
pub const VALIDATE_SCALE_TO_FALSE_POSITIVE_INVALID: &str =
    "ERR provided VALIDATESCALETO causes false positive to degrade to 0";
pub const NON_SCALING_AND_VALIDATE_SCALE_TO_IS_INVALID: &str =
    "ERR cannot use NONSCALING and VALIDATESCALETO options together";
```
# Amending BF.INFO command
We will add a new return value to the default result from BF.INFO on a default key, the new output will look like: 
```
'Capacity', 100, 'Size', 384, 'Number of filters', 1, 'Number of items inserted', 1, 'Error rate', b'0.01', 'Expansion rate', 2, 'Max scaled capacity', 26214300]
```
There is also a new optional arg (MAXSCALEDCAPACITY) which can be specified so running `BF.INFO key MAXSCALEDCAPACITY' will return xxxx. Where xxxx is the MAXSCALEDCAPACITY
As this information is only useful for scaling filters this option will be hidden non scaling filters this means that if the user has specified NONSCALING in the creation of their filter we will have not have the Max scaled capacity in the return. We also will not recognize MAXSCALEDCAPACITY as a valid optional arg to the info command in such a case as well.
# Testing
**Integration Tests**
Added checks for these three error messages in `test_bloom_command.py` as well as happy cases to show that different expansion, fp rates and capacity all are valid when using this new optional arg.

Updated the correctness tests to now check for the new field for a scaling filter when it does a BF.INFO call. For the non scaling correctness test we check that the new field does not appear

Added a more in depth correctness test as well where we initially try to create two filters where the VALIDATESCALETO is 1 more than the MAXSCALEDCAPACITY (one fails due to fp_rate degradation, one exceeds memory limit). We then actually create it with  VALIDATESCALETO set to the MAXSCALEDCAPACITY of the two keys. We then do an info call and get the returned MAXSCALEDCAPACITY and add that many items to both bloom filters. From there we then add 1 more item and expect a failure with either a exceeded memory error or fp rate reaches 0 error. At the very end we then compare a new result of a BF.INFO call to make sure that MAXSCALEDCAPACITY has not changed after adding items and scaling out.

**Unit tests:**
Added a unit tests for the calculate_max_scaled_capacity method with 5 different variations. For each variation it does a happy check where we will not receive an error from the command. We check that we get the correct returned capacity it reached after it went past the desired 'validate_scale_to'. Next we do an error case for each, four of them we expect to fail due to memory limit and one we expect to fail on fp_rate.

Also expanded test_seed to double check that random seed is different each time.
# Apendix 
Also suggested in the beginning was a second option: For option 2 we will add a new command that would output the capacity that can be reached before having an error thrown from scaling out given the bloom object memory usage limit (by default 128MB), fp rate, Tightening, expansion.
# Option 2: Create a new command BF.GETMAXCAPACITY 
This would check what the max capacity of a bloom filter is given a capacity, error rate and expansion. This command will look like:
```
BF.GETMAXCAPACITY [CAPACITY capacity] [ERROR fp_error] [EXPANSION expansion]
```
This command would then output the capacity that could be reached before having an error thrown from scaling out. This would **not** create a filter given the specified arguments.


madolson:
All things considered, I think the current naming and API arguments make sense. The only alternative I thought of is we can add it BF.INFO, which might make sense either way. We don't need a new command, and diverging on info should be OK since it's written to be extensible.

zackcam:
> All things considered, I think the current naming and API arguments make sense. The only alternative I thought of is we can add it BF.INFO, which might make sense either way. We don't need a new command, and diverging on info should be OK since it's written to be extensible.

Would we want to add to both BF.INFO and BF.INSERT?
I think that adding to both would make the most sense as adding to BF.INSERT can allow start up confirmation of the size they want is possible. While adding to BF.INFO can provide the max capacity information at any point while also showing how big it could get not just if it can get bigger than what the user wanted.

src/bloom/command_handler.rs
```diff
@@ -553,6 +554,21 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
                     }
                 };
             }
+            "ATLEASTCAPACITY" => {
```
KarthikSubbarao: These names fit well. 
```
ATLEASTCAPACITY
VALIDATECAPACITY
```

However, I am wondering if we should name it to sound specific to scaling cases
So, something like this:
```
CANSCALETO
CANSCALETOCAPACITY
```

src/bloom/utils.rs
```diff
@@ -32,10 +33,16 @@ pub const CAPACITY_LARGER_THAN_0: &str = "ERR (capacity should be larger than 0)
 pub const MAX_NUM_SCALING_FILTERS: &str = "ERR bloom object reached max number of filters";
 pub const UNKNOWN_ARGUMENT: &str = "ERR unknown argument received";
 pub const EXCEEDS_MAX_BLOOM_SIZE: &str = "ERR operation exceeds bloom object memory limit";
+pub const WANTED_CAPACITY_EXCEEDS_MAX_SIZE: &str =
+    "ERR Wanted capacity would go beyond bloom object memory limit";
```
KarthikSubbarao:
```suggestion
    "ERR Provided X causes bloom object to exceed memory limit";
```
X - TBD

src/bloom/utils.rs
```diff
@@ -32,10 +33,16 @@ pub const CAPACITY_LARGER_THAN_0: &str = "ERR (capacity should be larger than 0)
...
+pub const WANTED_CAPACITY_FALSE_POSITIVE_INVALID: &str =
+    "ERR False positive degrades too much to reach wanted capacity";
```
KarthikSubbarao:
```suggestion
    "ERR Provided X causes false positive to degrade to 0";
```
X - TBD

src/bloom/utils.rs
```diff
@@ -32,10 +33,16 @@ pub const CAPACITY_LARGER_THAN_0: &str = "ERR (capacity should be larger than 0)
...
 pub const KEY_EXISTS: &str = "BUSYKEY Target key name already exists.";
 pub const DECODE_BLOOM_OBJECT_FAILED: &str = "ERR bloom object decoding failed";
 pub const DECODE_UNSUPPORTED_VERSION: &str =
     "ERR bloom object decoding failed. Unsupported version";
+pub const NON_SCALING_AND_WANTED_CAPACITY_IS_INVALID: &str =
+    "ERR Specifying NONSCALING and ATLEASTCAPCITY is not allowed";
```
KarthikSubbarao:
```suggestion
    "ERR Cannot use NONSCALING and X options together";
```
X - TBD

src/bloom/utils.rs
KarthikSubbarao: Let's also use a new / similar error for the actual scaling out case

src/bloom/command_handler.rs
```diff
@@ -684,6 +720,28 @@ pub fn bloom_filter_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
                     }
                     Ok(ValkeyValue::Integer(val.expansion() as i64))
                 }
+                "MAXCAPACITY" => {
```
KarthikSubbarao: Can we also return back the fp rate through an "ERROR" option in the BF.INFO command?

src/bloom/utils.rs
```diff
@@ -455,6 +462,74 @@ impl BloomObject {
...
+        wanted_capacity: i64,
+        tightening_ratio: f64,
+        expansion: u32,
+    ) -> Result<i64, ValkeyError> {
+        let mut curr_filter_capacity = capacity;
+        let mut curr_total_capacity = capacity;
+        let mut curr_num_filters: u64 = 0;
+        let mut filters_memory_usage = 0;
+        while curr_total_capacity < wanted_capacity || wanted_capacity == -1 {
+            curr_total_capacity += curr_filter_capacity * expansion as i64;
```
KarthikSubbarao: 1. Can we calculate & update the new `curr_filter_capacity` once? We can avoid calculating it twice
2. Can we use `checked_mul` to ensure we do not exceed any bounds?

src/bloom/utils.rs
```diff
@@ -455,6 +467,80 @@ impl BloomObject {
...
+    ///                       want to check the maximum size we could scale up till
+    /// * `tightening_ratio` - The tightening ratio of the object
+    /// * `expansion` - The expanison rate of the object
+    ///
+    /// # Returns
+    /// * i64 - The maximum capacity that can be reached if called from BF.INFO. If called from BF.INSERT the size it reached after going past max capacity
+    /// * ValkeyError - Can return two different errors:
+    ///     CAN_SCALE_TO_EXCEEDS_MAX_SIZE: When scaling to the wanted capacity would go over the bloom object memory limit
+    ///     CAN_SCALE_TO_FALSE_POSITIVE_INVALID: When scaling to the wanted capcity would cause the false positive rate to reach 0
+    pub fn calculate_if_can_scale_to_is_valid(
```
KarthikSubbarao: Nit: `calculate_max_scaled_capacity`

src/bloom/utils.rs
```diff
@@ -29,13 +30,20 @@ pub const ERROR_RATE_RANGE: &str = "ERR (0 < error rate range < 1)";
...
 pub const CAPACITY_LARGER_THAN_0: &str = "ERR (capacity should be larger than 0)";
-pub const MAX_NUM_SCALING_FILTERS: &str = "ERR bloom object reached max number of filters";
+pub const FALSE_POSITIVE_DEGRADES_TO_O: &str = "ERR false positive degrades to 0 on scale out";
 pub const UNKNOWN_ARGUMENT: &str = "ERR unknown argument received";
 pub const EXCEEDS_MAX_BLOOM_SIZE: &str = "ERR operation exceeds bloom object memory limit";
+pub const CAN_SCALE_TO_EXCEEDS_MAX_SIZE: &str =
+    "ERR provided CANSCALETO causes bloom object to exceed memory limit";
+pub const MAX_NUM_SCALING_FILTERS: &str = "ERR bloom object reached max number of filters";
+pub const CAN_SCALE_TO_FALSE_POSITIVE_INVALID: &str =
+    "ERR provided CANSCALETO causes false positive to degrades to 0";
```
KarthikSubbarao: degrades -> degrade

tests/test_bloom_correctness.py
```diff
@@ -127,3 +130,31 @@ def test_scaling_filter(self):
         info_dict = dict(zip(it, it))
         # Validate correctness on a copy of a scaling bloom filter.
         self.validate_copied_bloom_correctness(client, filter_name, item_prefix, add_operation_idx, expected_fp_rate, fp_margin, info_dict)
+
+    def test_max_and_can_scale_to_correctness(self):
+        can_scale_to_commands = [
+            ('BF.INSERT key ERROR 0.00000001 CANSCALETO 13107101', "provided CANSCALETO causes bloom object to exceed memory limit" ),
+            ('BF.INSERT key EXPANSION 1 CANSCALETO 101601', "provided CANSCALETO causes false positive to degrades to 0" )
```
KarthikSubbarao: Can we expand this to handle variations in the `VALIDATESCALETO` arg to test success cases on variations of bloom objects?
* Expansion 1
* Expansion 2
* Various capacity / fp rate

src/bloom/command_handler.rs
```diff
@@ -462,6 +462,7 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
         true => (None, true),
         false => (Some(configs::FIXED_SEED), false),
     };
+    let mut can_scale_to = -1;
```
KarthikSubbarao: `can_scale_to` -> `test_scale_to` or `validate_scale_to`

src/bloom/command_handler.rs
```diff
@@ -568,6 +584,26 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
         // When the `ITEMS` argument is provided, we expect additional item arg/s to be provided.
         return Err(ValkeyError::WrongArity);
     }
+    // Check if we have a wanted capacity and calculate if we can reach that capacity
+    if can_scale_to > 0 {
+        if expansion == 0 {
```
KarthikSubbarao, line 591: Let's add a comment above to say:
```
Using VALIDATESCALETO and NONSCALING options together is invalid. 
```

src/bloom/command_handler.rs
```diff
@@ -568,6 +584,26 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
...
+            ));
+        }
+        match utils::BloomObject::calculate_if_can_scale_to_is_valid(
+            capacity,
+            fp_rate,
+            can_scale_to,
+            tightening_ratio,
+            expansion,
+        ) {
+            Ok(result) => result,
```
KarthikSubbarao: What is happening to `result` here?

If it is unused, we can do something like this:
```suggestion
            Ok(_) => (),
```

src/bloom/command_handler.rs
```diff
@@ -678,12 +714,31 @@ pub fn bloom_filter_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
...
                 "FILTERS" => Ok(ValkeyValue::Integer(val.num_filters() as i64)),
                 "ITEMS" => Ok(ValkeyValue::Integer(val.cardinality())),
+                "ERROR" => Ok(ValkeyValue::Float(val.fp_rate())),
                 "EXPANSION" => {
                     if val.expansion() == 0 {
                         return Ok(ValkeyValue::Null);
                     }
                     Ok(ValkeyValue::Integer(val.expansion() as i64))
                 }
+                "MAXSCALEDCAPACITY" if val.expansion() > 0 => {
```
KarthikSubbarao, line 727: Let's add a small comment here.
```
Only calculate and expose MAXSCALEDCAPACITY for scaling bloom objects. 
```

src/bloom/command_handler.rs
```diff
@@ -678,12 +714,31 @@ pub fn bloom_filter_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
...
+                    let max_capacity = match utils::BloomObject::calculate_if_can_scale_to_is_valid(
+                        val.filters()
+                            .first()
+                            .expect("Filter will be populated")
+                            .capacity(),
```
KarthikSubbarao: Is there an API on the bloom object to return the starting capacity that we can use here? (Or at least an API to get the first filter?)

If not, can we add one?

src/bloom/command_handler.rs
```diff
@@ -697,13 +752,34 @@ pub fn bloom_filter_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
...
                 result.push(ValkeyValue::Null);
             } else {
                 result.push(ValkeyValue::Integer(val.expansion() as i64));
             }
+            if val.expansion() != 0 {
+                let max_capacity = match utils::BloomObject::calculate_if_can_scale_to_is_valid(
+                    val.filters()
+                        .first()
+                        .expect("Filter will be populated")
+                        .capacity(),
```
KarthikSubbarao: Same comment as above

src/bloom/utils.rs
```diff
@@ -455,6 +467,80 @@ impl BloomObject {
...
+    ///
+    /// * `capacity` - The size of the initial filter in the bloom object.
+    /// * `fp_rate` - the false positive rate for the bloom object
+    /// * `can_scale_to` - the capcity we check to see if it can scale to. If this method is called from BF.INFO this is set as -1 as we
+    ///                       want to check the maximum size we could scale up till
+    /// * `tightening_ratio` - The tightening ratio of the object
+    /// * `expansion` - The expanison rate of the object
+    ///
+    /// # Returns
+    /// * i64 - The maximum capacity that can be reached if called from BF.INFO. If called from BF.INSERT the size it reached after going past max capacity
```
KarthikSubbarao: Does this sound right?
```suggestion
    /// * i64 - If called from BF.INFO, the maximum capacity that can be reached. If called from BF.INSERT, the size it reached before reaching the provided `can_scale_to`.
```

src/bloom/utils.rs
```diff
@@ -455,6 +467,80 @@ impl BloomObject {
             _ => Err(BloomError::DecodeUnsupportedVersion),
         }
     }
+
+    /// This method is called from two different bloom commands BF.INFO and BF.INSERT. The functionality varies slightly on which command it
+    /// is called from. When called from BF.INFO this method is used to find the maximum possible size that the bloom object could scale to
+    /// without throwing an error. When called from BF.INSERT this method is used to determine if it is possible to reach the capacity that
+    /// has been provided.
```
KarthikSubbarao:
```suggestion
    /// This method is called from two different bloom commands: BF.INFO and BF.INSERT. The functionality varies slightly on which command it
    /// is called from. When called from BF.INFO, this method is used to find the maximum possible size that the bloom object could scale to
    /// without throwing an error. When called from BF.INSERT, this method is used to determine if it is possible to reach the provided `can_scale_to`.
```

src/bloom/utils.rs
```diff
@@ -455,6 +467,80 @@ impl BloomObject {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `capacity` - The size of the initial filter in the bloom object.
+    /// * `fp_rate` - the false positive rate for the bloom object
+    /// * `can_scale_to` - the capcity we check to see if it can scale to. If this method is called from BF.INFO this is set as -1 as we
```
KarthikSubbarao:
```suggestion
    /// * `can_scale_to` - the capacity we check to see if it can scale to. If this method is called from BF.INFO this is set as -1 as we
```

src/bloom/utils.rs
KarthikSubbarao: It would help to also unit test this

src/bloom/utils.rs
zackcam: It will be the capacity the moment it passes the `can_scale_to` i.e if we have `can_scale_to = 500` and starting capacity of 100 with expansion 2, we will return 700 (100 + 200 + 400). Maybe say the size it reached when it surpassed (or was greater than) the provided can_scale_to

src/bloom/utils.rs
```diff
@@ -979,6 +1075,53 @@ mod tests {
...
+    #[case(100000, 0.00001, 1000000, 2, 1500000)]
+    #[case(100, 0.00001, 1000, 1, 1000)]
+    fn test_calculate_max_scaled_capacity(
+        #[case] capacity: i64,
+        #[case] fp_rate: f64,
+        #[case] validate_scale_to: i64,
+        #[case] expansion: u32,
+        #[case] resulting_size: i64,
+    ) {
+        // Validate that max scaled capacity returns the correct capacity reached when a valid can_scale to is provided
```
KarthikSubbarao:
```suggestion
        // Validate that max scaled capacity returns the correct capacity reached when a valid validate_scale_to to is provided
```

src/bloom/utils.rs
```diff
@@ -455,6 +474,78 @@ impl BloomObject {
...
+                }
+                return Err(ValkeyError::Str(VALIDATE_SCALE_TO_EXCEEDS_MAX_SIZE));
+            }
+            // Update overall memory usage
+            filters_memory_usage += curr_filter_size;
+            curr_total_capacity += curr_filter_capacity;
+            curr_filter_capacity = match curr_filter_capacity.checked_mul(expansion.into()) {
+                Some(new_capacity) => new_capacity,
+                None => {
+                    // u32:max cannot be reached with 64MB memory usage limit per filter even with a high fp rate (e.g. 0.9).
```
KarthikSubbarao: We can remove both occurrences of this comment.

With a 128MB memory limit for a bloom object overall, it is not possible to reach u32:max capacity.

src/bloom/utils.rs
```diff
@@ -979,6 +1075,53 @@ mod tests {
         assert_eq!(result2.err(), Some(BloomError::ExceedsMaxBloomSize));
     }
 
+    #[rstest]
+    #[case(1000, 0.01, 10000, 2, 15000)]
+    #[case(10000, 0.001, 100000, 4, 210000)]
+    #[case(50000, 0.0001, 500000, 3, 650000)]
+    #[case(100000, 0.00001, 1000000, 2, 1500000)]
+    #[case(100, 0.00001, 1000, 1, 1000)]
+    fn test_calculate_max_scaled_capacity(
```
KarthikSubbarao, line 1089: We can add the -1 case

src/bloom/utils.rs
```diff
@@ -455,6 +474,78 @@ impl BloomObject {
...
+    /// * ValkeyError - Can return two different errors:
+    ///     VALIDATE_SCALE_TO_EXCEEDS_MAX_SIZE: When scaling to the wanted capacity would go over the bloom object memory limit
+    ///     VALIDATE_SCALE_TO_FALSE_POSITIVE_INVALID: When scaling to the wanted capacity would cause the false positive rate to reach 0
+    pub fn calculate_max_scaled_capacity(
+        capacity: i64,
+        fp_rate: f64,
+        validate_scale_to: i64,
+        tightening_ratio: f64,
+        expansion: u32,
+    ) -> Result<i64, ValkeyError> {
```
KarthikSubbarao: We can use BloomError to remove a dependency on the valkeymodule-rs crate when we implement the BloomObject struct

tests/test_bloom_command.py
```diff
@@ -117,6 +123,7 @@ def test_bloom_command_behavior(self):
             ('bf.info TEST expansion', 2),
             ('BF.INFO TEST_EXPANSION EXPANSION', 9),
             ('BF.INFO TEST_CAPACITY CAPACITY', 2000),
+            ('BF.INFO TEST MAXSCALEDCAPACITY', 26214300),
```
KarthikSubbarao, line 126: Can we add the BF.INFO ERROR case?

tests/test_bloom_correctness.py
```diff
@@ -92,6 +94,7 @@ def test_scaling_filter(self):
             assert info_dict[b'Number of filters'] == filter_idx
             assert info_dict[b'Size'] > 0
             assert info_dict[b'Expansion rate'] == expansion
```
KarthikSubbarao, line 98: Can we check for 'Error rate'?

tests/test_bloom_correctness.py
```diff
@@ -74,6 +75,7 @@ def test_scaling_filter(self):
         assert info_dict[b'Number of filters'] == 1
         assert info_dict[b'Size'] > 0
         assert info_dict[b'Expansion rate'] == expansion
+        assert info_dict[b'Max scaled capacity'] == 20470000
```
KarthikSubbarao, line 80: Can we check for 'Error rate'?

src/bloom/data_type.rs
```diff
@@ -86,16 +86,15 @@ impl ValkeyDataType for BloomObject {
...
-                            "Failed to restore bloom object: Reached max number of filters",
-                        );
-                        return None;
-                    }
-                };
+            let new_fp_rate = match Self::calculate_fp_rate(fp_rate, i as i32, tightening_ratio) {
+                Ok(rate) => rate,
+                Err(_) => {
+                    logging::log_warning(
+                        "Failed to restore bloom object: Reached max number of filters",
```
KarthikSubbarao: Update error log

src/bloom/command_handler.rs
```diff
@@ -678,12 +714,29 @@ pub fn bloom_filter_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
                 "SIZE" => Ok(ValkeyValue::Integer(val.memory_usage() as i64)),
                 "FILTERS" => Ok(ValkeyValue::Integer(val.num_filters() as i64)),
                 "ITEMS" => Ok(ValkeyValue::Integer(val.cardinality())),
+                "ERROR" => Ok(ValkeyValue::Float(val.fp_rate())),
```
KarthikSubbarao, line 719: Can we update the test_bloom_basic test for BF.INFO to validate both the new fields?

src/bloom/command_handler.rs
```diff
@@ -462,6 +462,7 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
         true => (None, true),
         false => (Some(configs::FIXED_SEED), false),
     };
+    let mut validate_scale_to = -1;
```
KarthikSubbarao: Nit: Would prefer using `None` here to make this code more Rust like.
```
let mut validate_scale_to = None;
```

Later on, you can use this:
```
validate_scale_to = match input_args[idx].to_string_lossy().parse::<i64>() {
                    Ok(num) if (BLOOM_CAPACITY_MIN..=BLOOM_CAPACITY_MAX).contains(&num) => Some(num),
```

And then:

```
if let Some(validate_scale_to) {

}
```

KarthikSubbarao (approved):

KarthikSubbarao:
Thank you for the change @zackcam . Merging it in now

PR #42, opened 2025-01-24
Add dump and restore test

nnmehta:
Created a python test that tests bloom data using dump and restore commands after deleting a key and making sure the key exists after restoring

### DIFF
--- a/tests/test_bloom_basic.py
+++ b/tests/test_bloom_basic.py
@@ -364,3 +364,15 @@ class TestBloomBasic(ValkeyBloomTestCaseBase):
             assert self.client.execute_command('CONFIG SET bf.bloom-tightening-ratio 1.75') == b'ERR (0 < tightening ratio range < 1)'
         except ResponseError as e:
             assert str(e) == f"CONFIG SET failed (possibly related to argument 'bf.bloom-tightening-ratio') - ERR (0 < tightening ratio range < 1)"
+
+    def test_bloom_dump_and_restore(self):
+        """
+        This is a test that validates the bloom data has same debug digest value before and after using restore command
+        """     
+        client = self.server.get_new_client()
+        client.execute_command('BF.INSERT original error 0.001 capacity 2000 items 1')
+        dump = client.execute_command('DUMP original')
+        dump_digest = client.execute_command('DEBUG DIGEST-VALUE original')
+        client.execute_command('RESTORE', 'copy', 0, dump)
+        restore_digest = client.execute_command('DEBUG DIGEST-VALUE copy')
+        assert restore_digest == dump_digest


tests/test_bloom_basic.py
```diff
@@ -364,3 +364,11 @@ def test_bloom_string_config_set(self):
...
         except ResponseError as e:
             assert str(e) == f"CONFIG SET failed (possibly related to argument 'bf.bloom-tightening-ratio') - ERR (0 < tightening ratio range < 1)"
+
+    def test_bloom_dump_and_restore(self):
+        client = self.server.get_new_client()
+        client.execute_command('BF.INSERT dump error 0.001 capacity 2000 items 1')
+        dump = client.execute_command('DUMP dump')
+        assert client.execute_command('DEL dump') == 1
+        client.execute_command('RESTORE', 'dump', 0, dump)
+        assert client.execute_command('BF.EXISTS dump 1') == 1
```
KarthikSubbarao: Can we run the DEBUG command and check the DEBUG result on the object before and after restore?

tests/test_bloom_basic.py
```diff
@@ -364,3 +364,11 @@ def test_bloom_string_config_set(self):
             assert self.client.execute_command('CONFIG SET bf.bloom-tightening-ratio 1.75') == b'ERR (0 < tightening ratio range < 1)'
         except ResponseError as e:
             assert str(e) == f"CONFIG SET failed (possibly related to argument 'bf.bloom-tightening-ratio') - ERR (0 < tightening ratio range < 1)"
+
+    def test_bloom_dump_and_restore(self):
+        client = self.server.get_new_client()
+        client.execute_command('BF.INSERT dump error 0.001 capacity 2000 items 1')
+        dump = client.execute_command('DUMP dump')
+        assert client.execute_command('DEL dump') == 1
```
KarthikSubbarao: Let's skip out on the deletion for this test

tests/test_bloom_basic.py
```diff
@@ -364,3 +364,11 @@ def test_bloom_string_config_set(self):
             assert self.client.execute_command('CONFIG SET bf.bloom-tightening-ratio 1.75') == b'ERR (0 < tightening ratio range < 1)'
         except ResponseError as e:
             assert str(e) == f"CONFIG SET failed (possibly related to argument 'bf.bloom-tightening-ratio') - ERR (0 < tightening ratio range < 1)"
+
+    def test_bloom_dump_and_restore(self):
```
KarthikSubbarao, line 368: Can we add a small comment to describe the test?

nnmehta: Sure will do

tests/test_bloom_basic.py
nnmehta: Based on this: https://redis.io/docs/latest/commands/restore/
I thought we want to delete the key, then restore the deleted key.
Will update accordingly to just dump and restore the key rather than deleting

tests/test_bloom_basic.py
nnmehta: Sure, will check the Debug commands before and after

tests/test_bloom_basic.py
```diff
@@ -364,3 +364,16 @@ def test_bloom_string_config_set(self):
...
+        """
+        This is a test that validates the bloom data has same debug digest value before and after using restore command
+        """     
+        client = self.server.get_new_client()
+        client.execute_command('BF.INSERT original error 0.001 capacity 2000 items 1')
+        dump = client.execute_command('DUMP original')
+        dump_digest = client.execute_command('DEBUG DIGEST-VALUE original')
+        client.execute_command('RESTORE', 'copy', 0, dump)
+        restore_digest = client.execute_command('DEBUG DIGEST-VALUE copy')
+        assert client.execute_command('BF.EXISTS copy 1') == 1
```
zackcam: Do you think there would be any benefit of comparing the bf.info values as well or is this covered by comparing the digest value?

KarthikSubbarao (approved):

tests/test_bloom_basic.py
KarthikSubbarao: Yes, we can remove this. DIGEST should cover it

zackcam (approved):

PR #45, opened 2025-02-12
Adding usage of must_obey_client in wrapper to improve performace

zackcam:
Adding a wrapper for the ValkeyModule_MustObeyClient function in order to check if commands are from the primary client or AOF client in a more efficient way. If this function does not exist we fall back to checking for replicated flag in the GetContextFlags API as a best effort.

Updating one test to use different specified reserve arguments in order to reduce the amount of items we need to add to fill up the filter which in turn will reduce the run time of the test but will still provide the same testing functionality.


src/wrapper/mod.rs
```diff
@@ -1,2 +1,23 @@
...
+    match unsafe { raw::ValkeyModule_MustObeyClient } {
+        Some(func) => {
+            let status = unsafe { func(ctx_raw) as isize };
+            match status {
+                1 => true,
+                0 => false,
+                _ => ctx.get_flags().contains(ContextFlags::REPLICATED),
+            }
+        }
+        None => ctx.get_flags().contains(ContextFlags::REPLICATED),
```
KarthikSubbarao: Let's add a comment here:

```suggestion
        None => {
            // Fallback to checking for the replicated flag in the GetContextFlags API as a best effort.
            ctx.get_flags().contains(ContextFlags::REPLICATED);
        }
```

src/wrapper/mod.rs
```diff
@@ -1,2 +1,23 @@
+use valkey_module::{raw, Context, ContextFlags, ValkeyModuleCtx};
+
 pub mod bloom_callback;
 pub mod defrag;
+
+/// Wrapper for the ValkeyModule_MustObeyClient function.
+/// Takes in an Context and returns true if the if commands are arriving
+/// from the primary client or AOF client and should never be rejected.
+/// False otherwise.
+pub fn is_must_obey_client(ctx: &Context) -> bool {
```
KarthikSubbarao:
```suggestion
pub fn must_obey_client(ctx: &Context) -> bool {
```

src/wrapper/mod.rs
```diff
@@ -1,2 +1,23 @@
...
+    let ctx_raw = ctx.get_raw() as *mut ValkeyModuleCtx;
+    match unsafe { raw::ValkeyModule_MustObeyClient } {
+        Some(func) => {
+            let status = unsafe { func(ctx_raw) as isize };
+            match status {
+                1 => true,
+                0 => false,
+                _ => ctx.get_flags().contains(ContextFlags::REPLICATED),
```
KarthikSubbarao: This should never be any value other than 0 or 1. 

Instead of silently failing, if this is not 1 or 0, we should know immediately and fix the bug.

So, let us add a panic here
```suggestion
                _ => panic!("We do not expect ValkeyModule_MustObeyClient to return anything other than 1 or 0."),
```

src/bloom/command_handler.rs
```diff
@@ -615,7 +615,7 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
         }
     };
     // Skip bloom filter size validation on replicated cmds.
-    let validate_size_limit = !ctx.get_flags().contains(ContextFlags::REPLICATED);
+    let validate_size_limit = !is_must_obey_client(ctx);
```
KarthikSubbarao: We are missing a similar update in the `bloom_filter_load` function

KarthikSubbarao (changes requested):

KarthikSubbarao:
While looking into the macos failure, saw that it was not able to find the ValkeyModule_MustObeyClient symbol from the redismodule.o object file while linking.

This means we will need this change: https://github.com/valkey-io/valkeymodule-rs/pull/176 for the macos build to work. This PR creates a valkeymodule.o object file file

We will wait until this valkeymodule-rs PR is merged in and released to merge in this change

Cargo.toml
```diff
@@ -38,3 +38,4 @@ debug-assertions = true
 default = ["min-valkey-compatibility-version-8-0"]
 enable-system-alloc = ["valkey-module/enable-system-alloc"]
 min-valkey-compatibility-version-8-0 = []
+valkey_8_0 = []  # Empty feature flag for Valkey 8.0
```
KarthikSubbarao: You can add something like this to provide more context.

```suggestion
valkey_8_0 = []  # valkey-bloom is intended to be loaded on server versions >= Valkey 8.1 and by default it is built this way (unless this flag is provided). It is however compatible with Valkey version 8.0 if the user explicitly provides this feature flag in their cargo build command.
```

src/wrapper/mod.rs
```diff
@@ -1 +1,34 @@
...
+/// from the primary client or AOF client and should never be rejected.
+/// False otherwise.
+pub fn must_obey_client(ctx: &Context) -> bool {
+    // If we are using valkey 8.0 then we cannot use ValkeyModule_MustObeyClient so must go back to the default
+    // of checking for the replicated flag
+    #[cfg(not(feature = "valkey_8_0"))]
+    {
+        let ctx_raw = ctx.get_raw() as *mut valkey_module::ValkeyModuleCtx;
+
+        match unsafe { valkey_module::raw::ValkeyModule_MustObeyClient } {
```
KarthikSubbarao: We should be able to unwrap here on 8.1 because it should exist for sure. We can do something like `valkey_module::raw::ValkeyModule_MustObeyClient.expect("We expect that ValkeyModule_MustObeyClient exists on >= 8.1")`

src/wrapper/mod.rs
```diff
@@ -1 +1,34 @@
...
+                }
+            }
+            // Fallback to checking for replicated flag in the GetContextFlags API as a best effort.
+            None => ctx.get_flags().contains(ContextFlags::REPLICATED),
+        }
+    }
+
+    #[cfg(feature = "valkey_8_0")]
+    {
+        ctx.get_flags().contains(ContextFlags::REPLICATED)
```
KarthikSubbarao:
```suggestion
        // On Valkey 8.0, fallback to checking for replicated flag in the GetContextFlags API as a best effort.
        ctx.get_flags().contains(ContextFlags::REPLICATED)
```

src/wrapper/mod.rs
```diff
@@ -1 +1,34 @@
...
+
 pub mod bloom_callback;
+
+/// Wrapper for the ValkeyModule_MustObeyClient function.
+/// Takes in an Context and returns true if the if commands are arriving
+/// from the primary client or AOF client and should never be rejected.
+/// False otherwise.
+pub fn must_obey_client(ctx: &Context) -> bool {
+    // If we are using valkey 8.0 then we cannot use ValkeyModule_MustObeyClient so must go back to the default
+    // of checking for the replicated flag
```
KarthikSubbarao: We can remove this

src/wrapper/mod.rs
```diff
@@ -1 +1,34 @@
...
+    #[cfg(not(feature = "valkey_8_0"))]
+    {
+        let ctx_raw = ctx.get_raw() as *mut valkey_module::ValkeyModuleCtx;
```
KarthikSubbarao, line 13:
```suggestion
        // ValkeyModule_MustObeyClient exists since Valkey 8.1, so we use this as it is a more performant check.
        let ctx_raw = ctx.get_raw() as *mut valkey_module::ValkeyModuleCtx;
```

src/wrapper/mod.rs
zackcam: Looks like we don't need the expect (doesn't allow an expect). I updated this though so it does the call in a less verbose way.

src/wrapper/mod.rs
zackcam: Removed

KarthikSubbarao (approved):

PR #49, opened 2025-03-05
Making it so github workflows can see if we have memory leaks as well as adding asan capability to ./build.sh

zackcam:
## Adding better ASAN functionality to integration tests

Currently the asan capability we have can't tell if we have memory leaks occurring in the integration tests from new prs. This PR fixes that by parsing an output file which is created as tests are running. We then check for leaks in this ouput file and fail the workflow if there any as well as outputting the test that they occurred in as well. In lieu of this as well I have added capability of doing asan builds locally where it will perform the same as the github workflow would. The asan build can be done by doing setting `ASAN_BUILD ` to anything.

Currently this will cause the workflows to fail as there is a known issue with configuration sets and gets in valkeymodule-rs. Once the fix for this is in we should rerun the workflows and make sure that no new memory leaks are occurring as well.

Below is a screenshot of a what it looks like with memory leaks locally:
image
Below is a screenshot of memory leaks occuring in the workflow:
image
Link to the build the screenshot above is taken from: https://github.com/zackcam/valkey-bloom/actions/runs/13665508262/job/38205917231


build.sh
```diff
@@ -42,7 +42,12 @@ else
     git clone "$REPO_URL"
     cd valkey
     git checkout "$SERVER_VERSION"
-    make -j
+    make distclean
+    if [ ! -z "${ASAN_BUILD}" ]; then
+        make -j4 SANITIZER=address
```
KarthikSubbarao: We don't need to specify the cores, it should already be auto detected and used by the makefile
```suggestion
        make -j SANITIZER=address
```

tests/conftest.py
```diff
@@ -1,5 +1,15 @@
...
+    return request.param
+
+@pytest.fixture(autouse=True, scope="session")
+def cleanup_valkey():
+    yield
+    # After all tests complete
+    try:
+        subprocess.run(['pkill', '-f', 'valkey-se'], check=False)
+    except Exception as e:
+        print(f"Warning: Failed to kill valkey server: {e}")
```
KarthikSubbarao: Question: What is this used for?

KarthikSubbarao:
Great work getting ASAN parsing working. This is very important and useful @zackcam

build.sh
```diff
@@ -64,12 +69,39 @@ fi
...
+        python3 -m pytest --capture=sys --cache-clear -v "$SCRIPT_DIR/tests/" -k $TEST_PATTERN 2>&1 | tee test_output.tmp
+    else
+        echo "TEST_PATTERN is not set. Running all integration tests."
+        python3 -m pytest --capture=sys --cache-clear -v "$SCRIPT_DIR/tests/" 2>&1 | tee test_output.tmp
+    fi
+
+    # Check for memory leaks in the output
+    if grep -q "LeakSanitizer: detected memory leaks" test_output.tmp; then
+        RED='\033[0;31m'
+        echo -e "${RED}Memory leaks detected in the following tests:"
```
KarthikSubbarao, line 84: [Only if it is simple to do - we can ignore and skip this otherwise] - Could we mention how many tests had leaks detected in them?

tests/conftest.py
zackcam: This is done as at the end of the integration tests it would get stuck on the redirecting output for ASAN runs as there was a valkey server that wasn't shutting down properly. This was done to make it so that all valkey servers got cleaned up and it could actually progress. The real fix should be to find the test which isn't tearing down properly and fix it. But this solution also will clean up servers incase people interrupt their teardown as well

build.sh
zackcam: Do you mean have a separate flag to check for memory leaks on ASAN completely. i.e. have a case where asan is done but no memory leaks are looked for and a case where asan is done and memory leaks are looked for?

KarthikSubbarao: Sorry for typos. I mean to ask that when the build OR CICD fails, can we log a line that says "X python integration tests have leaks detected in them" where X can be 5 for example?

zackcam: Example in workflow run I can set it up as an error or colour as needed
image
Example in local run (Did it with a test pattern as well to check that works and because it is quicker):
image

tests/conftest.py
KarthikSubbarao: Thanks

> The real fix should be to find the test which isn't tearing down properly and fix it.

Yes, we can do this separately. I think we identified one or two tests that need this fixed

zackcam: I now have fixed this so we don't need this fixture anymore.

tests/valkeytests/valkey_test_case.py
```diff
@@ -634,7 +634,6 @@ class ReplicationTestCase(ValkeyTestCase):
     num_replicas = 1
 
     def setup_replication(self, num_replicas = num_replicas):
-        super(ReplicationTestCase, self).setup()
```
zackcam, line 637: This was causing some valkey servers to be left up and running and actually doesn't do anything. Removing this in order to not use the fixture I had before on end of tests which killed all running servers

KarthikSubbarao (approved):
Thank you. Merging in

PR #50, opened 2025-03-11
Adding tightening ratio to bf.info and updating allowed range for bloom-expansion config

zackcam:
Two changes in this PR
### First Change
Added it so tightening ratio is returned by BF.INFO.
**Testing**
Expanded two tests that check info to make sure the tightening ratio we get returned is what we expect. Also added a similar check in one place for false positive rate as well. Tested the negative case as well so if we have a scaling filter we should not return the tightening ratio.

### Second Change
Previously we had a lower limit of 1 for setting the expansion rate, this meant we wouldn't allow a user to have by default a non scaling filter. I updated this range to now include 0 at the lower end so now the user can set it so they create non-scaling filters by default
**Testing**
Added a new test that checks that setting configs actually updated the default creations.

### DIFF
--- a/src/bloom/command_handler.rs
+++ b/src/bloom/command_handler.rs
@@ -716,6 +716,9 @@ pub fn bloom_filter_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
                 "FILTERS" => Ok(ValkeyValue::Integer(val.num_filters() as i64)),
                 "ITEMS" => Ok(ValkeyValue::Integer(val.cardinality())),
                 "ERROR" => Ok(ValkeyValue::Float(val.fp_rate())),
+                "TIGHTENING" if val.expansion() > 0 => {
+                    Ok(ValkeyValue::Float(val.tightening_ratio()))
+                }
                 "EXPANSION" => {
                     if val.expansion() == 0 {
                         return Ok(ValkeyValue::Null);
@@ -759,8 +762,9 @@ pub fn bloom_filter_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
                 result.push(ValkeyValue::Null);
             } else {
                 result.push(ValkeyValue::Integer(val.expansion() as i64));
-            }
-            if val.expansion() != 0 {
+                // The following fields are only relevant to scalable filters, so will only be included if expansion is not equal to 0.
+                result.push(ValkeyValue::SimpleStringStatic("Tightening ratio"));
+                result.push(ValkeyValue::Float(val.tightening_ratio()));
                 let max_capacity = match utils::BloomObject::calculate_max_scaled_capacity(
                     val.starting_capacity(),
                     val.fp_rate(),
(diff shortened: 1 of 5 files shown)


src/bloom/command_handler.rs
```diff
@@ -753,6 +754,8 @@ pub fn bloom_filter_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
                 ValkeyValue::Integer(val.cardinality()),
                 ValkeyValue::SimpleStringStatic("Error rate"),
                 ValkeyValue::Float(val.fp_rate()),
+                ValkeyValue::SimpleStringStatic("Tightening ratio"),
+                ValkeyValue::Float(val.tightening_ratio()),
```
KarthikSubbarao: Should we show this only for scalable bloom filters?
Since tightening does not matter to non scaling

zackcam: Good point, I've updated to check if expansion is 0 and if so not output this

tests/test_bloom_command.py
```diff
@@ -154,7 +154,10 @@ def test_bloom_command_behavior(self):
         filter_index = bf_info.index(b'Number of filters') + 1
         item_index = bf_info.index(b'Number of items inserted') + 1
         expansion_index = bf_info.index(b'Expansion rate') + 1
+        error_rate_index = bf_info.index(b'Error rate') + 1
+
         assert bf_info[capacity_index] == self.client.execute_command('BF.INFO BF_INFO CAPACITY') == 2000
         assert bf_info[filter_index] == self.client.execute_command('BF.INFO BF_INFO FILTERS') == 1
         assert bf_info[item_index] == self.client.execute_command('BF.INFO BF_INFO ITEMS') == 0
         assert bf_info[expansion_index] == self.client.execute_command('BF.INFO BF_INFO EXPANSION') == None
+        assert bf_info[error_rate_index] == self.client.execute_command('BF.INFO BF_INFO ERROR') == str(0.5).encode()
```
KarthikSubbarao: 1. Can we also add validation for TIGHTENING RATIO as well (general / overall response)?

2. Can we add a success case in the `basic_behavior_test_case` list above for individually querying tightening ratio for scaling?

3. Can we add an error case in the `basic_error_test_cases` list in the `test_bloom_command_error` test for individually querying tightening ratio for non scaling filters?

zackcam: 1. This was on a non scaling filter I can create a new filter and check for that if we want.
2. Will add
3. Will add

src/bloom/command_handler.rs
```diff
@@ -759,8 +762,9 @@ pub fn bloom_filter_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
                 result.push(ValkeyValue::Null);
             } else {
                 result.push(ValkeyValue::Integer(val.expansion() as i64));
-            }
-            if val.expansion() != 0 {
+                // The following fields are only relevant to scalable filters, so will only be ouput if expanison is not equal to 0
```
KarthikSubbarao: Fixing some typos

```suggestion
                // The following fields are only relevant to scalable filters, so will only be included if expansion is not equal to 0
```

KarthikSubbarao:
Could we update the PR description / title / commit message to explain that there are two changes: (1) Adding tightening ratio to BF.INFO (2) Making BF.BLOOM-EXPANSION also control scaling vs non scaling?

KarthikSubbarao (approved):

PR #51, opened 2025-03-19
Removing unnecessary code form valkey_test_case

zackcam:
Cleaning up valkey_test_case. Consists of:
## [Commit 1](https://github.com/valkey-io/valkey-bloom/pull/51/commits/d671c81bd977ae695a69796b4ab6c79b7a49f4bc)
1. Removing ValkeyInfo and ValkeyClient which had no purpose other than extra function calls which could be removed
2. Updating waiters so now we won't use `@wait` and instead use explicit function calls
3. Added clean up for aof folder and files in that folder after running tests
4. Removed functions which only purpose was to be called by another function for waits

## [Commit 2](https://github.com/valkey-io/valkey-bloom/pull/51/commits/795eb42012c0ef66194b2786d345798901e4cb36)
1. Removed server_id variable
2. Updated port tracker fixture so framework is now compatible with pytest version 7 as well as 6

## [Commit 3](https://github.com/valkey-io/valkey-bloom/pull/51/commits/73d2334ec0fd9432bb0ae654985a1d187d7df499)
1. Removed automatic creation of server with specific args in `valkey_test_case.py`. Instead now the `valkey_bloom_test_case.py` now creates the server and decides what args to use
2. Updated server_path to be more flexible in being set. Now `valkey_test_case.py` will try to find valkey server from just valkey_server if server path is not provided in creating server
3. Consolidated creating, arg setting and starting a server in one function call so now less calls are needed for a test by test basis for creating and starting a server
4. Removed unnecessary variables set in `valkey_test_case.py` eg `num_keys`

### DIFF
--- a/tests/test_bloom_aofrewrite.py
+++ b/tests/test_bloom_aofrewrite.py
@@ -6,10 +6,8 @@ from valkeytests.conftest import resource_port_tracker
 class TestBloomAofRewrite(ValkeyBloomTestCaseBase):
     
     def get_custom_args(self):
-        # test aof rewrite should avoid bloom filter override as rdb. use aof
-        args = super().get_custom_args()
-        args.update({'aof-use-rdb-preamble': 'no', 'appendonly': 'yes'})
-        return args
+        self.args.update({'aof-use-rdb-preamble': 'no', 'appendonly': 'yes'})
+        return self.args
 
     def test_basic_aofrewrite_and_restore(self):
         client = self.server.get_new_client()
--- a/tests/test_bloom_aofrewrite.py
+++ b/tests/test_bloom_aofrewrite.py
@@ -4,10 +4,6 @@ from valkey_bloom_test_case import ValkeyBloomTestCaseBase
 from valkeytests.conftest import resource_port_tracker
 
 class TestBloomAofRewrite(ValkeyBloomTestCaseBase):
-    
-    def get_custom_args(self):
-        self.args.update({'aof-use-rdb-preamble': 'no', 'appendonly': 'yes'})
-        return self.args
 
     def test_basic_aofrewrite_and_restore(self):
         client = self.server.get_new_client()
--- a/tests/test_bloom_aofrewrite.py
+++ b/tests/test_bloom_aofrewrite.py
@@ -19,9 +19,10 @@ class TestBloomAofRewrite(ValkeyBloomTestCaseBase):
         assert bf_exists_result_1 == 1
         bf_info_result_1 = client.execute_command('BF.INFO testSave')
         assert(len(bf_info_result_1)) != 0
-        curr_item_count_1 = client.info_obj().num_keys()
+        curr_item_count_1 = self.server.num_keys()
+
         # cmd debug digest
-        server_digest = client.debug_digest()
+        server_digest = client.execute_command('DEBUG', 'DIGEST')
         assert server_digest != None or 0000000000000000000000000000000000000000
         object_digest = client.execute_command('DEBUG DIGEST-VALUE testSave')
 
@@ -32,13 +33,13 @@ class TestBloomAofRewrite(ValkeyBloomTestCaseBase):
         time.sleep(1)
         self.server.restart(remove_rdb=False, remove_nodes_conf=False, connect_client=True)
         assert self.server.is_alive()
-        restored_server_digest = client.debug_digest()
+        restored_server_digest = client.execute_command('DEBUG', 'DIGEST')
         restored_object_digest = client.execute_command('DEBUG DIGEST-VALUE testSave')
         assert restored_server_digest == server_digest
         assert restored_object_digest == object_digest
 
         # verify restore results
-        curr_item_count_2 = client.info_obj().num_keys()
+        curr_item_count_2 = self.server.num_keys()
         assert curr_item_count_2 == curr_item_count_1
         bf_exists_result_2 = client.execute_command('BF.EXISTS testSave item')
         assert bf_exists_result_2 == 1
@@ -53,7 +54,7 @@ class TestBloomAofRewrite(ValkeyBloomTestCaseBase):
         self.add_items_till_capacity(self.client, "key1", 7500, 1, "item_prefix")
 
         # cmd debug digest
-        server_digest = self.client.debug_digest()
+        server_digest = self.client.execute_command('DEBUG', 'DIGEST')
         assert server_digest != None or 0000000000000000000000000000000000000000
         object_digest = self.client.execute_command('DEBUG DIGEST-VALUE key1')
 
@@ -63,7 +64,7 @@ class TestBloomAofRewrite(ValkeyBloomTestCaseBase):
         # restart server
         self.server.restart(remove_rdb=False, remove_nodes_conf=False, connect_client=True)
         assert self.server.is_alive()
-        restored_server_digest = self.client.debug_digest()
+        restored_server_digest = self.client.execute_command('DEBUG', 'DIGEST')
         restored_object_digest = self.client.execute_command('DEBUG DIGEST-VALUE key1')
         assert restored_server_digest == server_digest
         assert restored_object_digest == object_digest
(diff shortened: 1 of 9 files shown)


tests/valkeytests/valkey_test_case.py
```diff
@@ -244,7 +102,10 @@ def exit(self, remove_rdb=True, remove_nodes_conf=True):
             if "logfile" in self.args and os.path.exists(os.path.join(self.cwd, self.args["logfile"])):
                 os.remove(os.path.join(self.cwd, self.args["logfile"]))
 
-        if remove_rdb and "dbfilename" in self.args and os.path.exists(os.path.join(self.cwd, self.args["dbfilename"])):
+            if test_teardown and "appenddirname" in self.args and os.path.exists(os.path.join(self.cwd, self.args["appenddirname"])):
+                shutil.rmtree(os.path.join(self.cwd, self.args["appenddirname"]))
```
KarthikSubbarao, line 106: Question: Why do we use a different approach for removing the aof dir instead of os.remove?

tests/test_bloom_aofrewrite.py
```diff
@@ -6,10 +6,8 @@
 class TestBloomAofRewrite(ValkeyBloomTestCaseBase):
     
     def get_custom_args(self):
```
KarthikSubbarao: Where is this function used?

tests/test_bloom_replication.py
```diff
@@ -8,12 +8,12 @@ class TestBloomReplication(ReplicationTestCase):
...
 
-    def get_custom_args(self):
-        self.set_server_version(os.environ['SERVER_VERSION'])
-        return {
-            'loadmodule': os.getenv('MODULE_PATH'),
-            'bf.bloom-use-random-seed': self.use_random_seed,
-        }
+    @pytest.fixture(autouse=True)
+    def setup_test(self, setup):
+        self.args = {"maxmemory-policy":"allkeys-random", "activerehashing":"yes", "repl-diskless-sync": "yes", "save": "", "enable-debug-command":"yes", 'loadmodule': os.getenv('MODULE_PATH'),'bf.bloom-use-random-seed': self.use_random_seed}
```
KarthikSubbarao: We can remove the configs which are not needed for the replication tests

tests/valkey_bloom_test_case.py
```diff
@@ -10,12 +10,15 @@ class ValkeyBloomTestCaseBase(ValkeyTestCase):
...
 
-    def get_custom_args(self):
-        self.set_server_version(os.environ['SERVER_VERSION'])
-        return {
-            'loadmodule': os.getenv('MODULE_PATH'),
-            'bf.bloom-use-random-seed': self.use_random_seed,
-        }
+    @pytest.fixture(autouse=True)
+    def setup_test(self, setup):
+        args = {"maxmemory-policy":"allkeys-random", "activerehashing":"yes", "repl-diskless-sync": "yes", "save": "", "enable-debug-command":"yes", 'loadmodule': os.getenv('MODULE_PATH'),'bf.bloom-use-random-seed': self.use_random_seed}
```
KarthikSubbarao: We can remove the configs which are not needed for the bloom tests

tests/valkeytests/valkey_test_case.py
```diff
@@ -277,9 +137,9 @@ def _waitForExit(self):
     def pid(self):
         return self.server.pid
 
-    @wait(timeout = 5)
+    # DO we even need this it does the same as _waitForServerPoll
```
KarthikSubbarao: Yes, is_down seems like a cleaner function

tests/valkey_bloom_test_case.py
```diff
@@ -10,12 +10,15 @@ class ValkeyBloomTestCaseBase(ValkeyTestCase):
...
+        server_path = f"{os.path.dirname(os.path.realpath(__file__))}/.build/binaries/{os.environ['SERVER_VERSION']}/valkey-server"
+
+        self.server, self.client = self.create_server(testdir = self.testdir,  server_path=server_path, args=args)
+
+        print("startup args are: ", args)
```
KarthikSubbarao: For this file, we could use a logger instead of print.

`import logging`

tests/valkeytests/valkey_test_case.py
zackcam: The aof dir has files inside which os import cant clean up itself unless I go file by file. This way i can just remove the directory instead of removing files one by one

tests/valkeytests/valkey_test_case.py
zackcam: Will remove this and update places it is used

tests/test_bloom_aofrewrite.py
zackcam: Does not seem like it is used will remove as it looks like we don't need these arguments anyway for tests to run and pass

tests/valkeytests/valkey_test_case.py
```diff
@@ -51,183 +52,40 @@ def expect(lhs, op, rhs):
...
+    
     def set_startup_args(self, args):
         self.args.update(args)
 
     def get_new_client(self):
-        return ValkeyClient.create_from_server(self)
+        return self.create_from_server(self)
 
-    def exit(self, remove_rdb=True, remove_nodes_conf=True):
+    def exit(self, test_teardown=True, remove_nodes_conf=True):
```
KarthikSubbarao: test_teardown  - > cleanup

tests/valkeytests/valkey_test_case.py
```diff
@@ -51,183 +52,40 @@ def expect(lhs, op, rhs):
...
         self.args["port"] = self.port
         self.args["logfile"] = "logfile_{}".format(port)
         self.args["dbfilename"] = "testrdb-{}.rdb".format(port)
+        self.args["appenddirname"] = "aof-{}".format(port)
         self.cwd = cwd
         self.valkey_path = server_path
 
+    @classmethod
+    def create_from_server(self, server, db=0):
+        print(("Created regular client for port {}".format(server.port)))
```
KarthikSubbarao: We can add a logger here. For the expected code paths, we can use INFO or DEBUG.

We can use warning/error for the unexpected code paths.

And set logging level to warning

tests/valkeytests/valkey_test_case.py
```diff
@@ -51,183 +52,40 @@ def expect(lhs, op, rhs):
...
-    def debug_digest(self):
-        return self.execute_command('DEBUG', 'DIGEST')
-
 class ValkeyServerHandle(object):
     """Handle to a valkey server process"""
     
     DEFAULT_BIND_IP = "0.0.0.0"
 
-    def __init__(self, bind_ip, port, port_tracker, server_path, cwd='.', server_id=0):
+    def __init__(self, bind_ip, port, port_tracker, server_path, cwd='.'):
```
KarthikSubbarao: We can add a default path

tests/valkeytests/valkey_test_case.py
zackcam: Replaced all prints with logging

tests/valkeytests/valkey_test_case.py
```diff
@@ -537,67 +380,48 @@ def wait_until_command(self, server, cmd):
...
-    repl_save_info_size = 83 # Bytes used for saving replication info in RDB aux fields
-    diskless_overhead = 87 # RDB overhead is 2 x 40 byte EOF marker + 7 characters ("$EOF:" + "\r\n") for the beginning of the EOF marker
-    server_path = ".build/binaries/unstable/valkey-server" #Unstable is the default server build
-
-    def set_server_version(self, new_version):
-        self.server_path = f".build/binaries/{new_version}/valkey-server"
+    server_path = "valkey-server" # The default server build is assumed that valkey-server is set
 
     def common_setup(self):
         self.maxmemory = "500MB"
```
KarthikSubbarao: We can remove this

KarthikSubbarao (approved):

PR #57, opened 2025-03-29
Cleaning up usage of valkey-test-framework

KarthikSubbarao:
Cleaning up usage of valkey-test-framework

* .build is renamed to build and is included in .gitignore. This single folder will contain cached valkey and valkey-test-frame-work src code used for integration testing
* Renaming previous folder of valkeytests to be valkeytestframework

Build Script Updates
* clean up / remove valkey dir when done generating server binary

CICD updates:
* Also run 8.1 along with 8.0 and unstable


tests/valkey_bloom_test_case.py
```diff
@@ -86,7 +103,7 @@ def add_items_till_scaling_failure(self, client, filter_name, starting_item_idx,
                     raise RuntimeError("Unexpected return value 1 from BF.ADD")
         except Exception as e:
             if "non scaling filter is full" in str(e):
-                return
+                return new_item_idx
```
zackcam, line 107: Only thing I could see is that when we call `validate_scaling_failure` we then check with index 1 however this method does allow not starting at 1 for the `new_item_idx` so either we could return the new_item_idx - 1 as well and pass both those in to `validate_scaling_failure` or just do a `existing_item = f"{item_prefix}{new_item_idx-1}"` in `validate_scaling_failure`

zackcam (approved):

PR #66, opened 2025-06-10
Removing failing on clippy errors in github ci

zackcam:
Removing failures on clippy errors for the github ci, the build script will still show any clippy errors that are present. The ci will also still show any clippy errors

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,13 +24,13 @@ jobs:
     - name: Run cargo and clippy format check
       run: |
         cargo fmt --check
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -24,13 +24,13 @@ jobs:
     - name: Run cargo and clippy format check
```
KarthikSubbarao, line 24: Can we update the comments / the descriptions accordingly?

KarthikSubbarao: Same comment through out the file

KarthikSubbarao (approved):

PR #67, opened 2025-06-11
Adding clean option for build script and fixing current clippy issues

zackcam:
This will add an easy way of cleaning up the bloom repo instead of running cargo commands. This will remove all build artifacts 
Also fixed the clippy errors that were being emitted

Let me know if we want to clean test-data if that exists as well

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -8,6 +8,15 @@ set -e
 SCRIPT_DIR=$(pwd)
 echo "Script Directory: $SCRIPT_DIR"
 
+if [ "$1" = "clean" ]; then
+  echo "Cleaning build artifacts"
+  rm -rf target/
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -8,6 +8,14 @@ set -e
 SCRIPT_DIR=$(pwd)
 echo "Script Directory: $SCRIPT_DIR"
 
+if [ "$1" = "clean" ]; then
```
KarthikSubbarao, line 11: > Let me know if we want to clean test-data if that exists as well

Yes, that would be nice. Is there anything else pytest related which is not cleaned up?

KarthikSubbarao (approved):

PR #69, opened 2025-06-19
Improving build script to provide more options

zackcam:
Added a new help, integration, and unit argument for the build script this will allow developers/users to target specifically what they want to do with building in an easier manner.

This also has been reflected in the README.md with the new arguments and a quick explanation of what they do

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -5,10 +5,81 @@
 # Exit the script if any command fails
 set -e
 
+function print_usage() {
+cat<<EOF
+Usage: build.sh [--release] [--unit] [--integration] [--clean]
+
+    --help | -h               Print this help message and exit.
+    --release                 Builds the release configuration.
+    --unit                    Builds the unit tests configuration.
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -5,10 +5,73 @@
...
+RUN_INTEGRATION=0
+BUILD_RELEASE=1
+CLEAN_BUILD=0
+
+## Parse command line argument
+while [[ $# -gt 0 ]]; 
+do
+    arg="$1"
+    case $arg in
+        --release)
```
KarthikSubbarao, line 44: Can we add a flag to skip the `clippy` checks? valkey-bundle calls this build.sh script so they would want an option to build the bloom binary without any linter / formatting checks since Rust linter updates may occur at a higher frequency than our releases

Nikhil-Manglore: The bundle doesn't require this flag anymore, you can still do it if you'd like though

PR #72, opened 2025-08-04
Adding Bloom support for external server

Nikhil-Manglore:
This change adds the ability to run tests against an external Valkey server instead of always starting a new one.

External server mode: If we set the environment variable,  `VALKEY_EXTERNAL_SERVER=true , when we run the bloom pytests we will connect to an existing Valkey server at the specified host and port.

Local server mode: If not set or false, we will continue with the original behavior of starting a new Valkey server with the module loaded

### DIFF
--- a/tests/test_bloom_basic.py
+++ b/tests/test_bloom_basic.py
@@ -10,8 +10,12 @@ class TestBloomBasic(ValkeyBloomTestCaseBase):
         client = self.server.get_new_client()
         # Validate that the valkey-bloom module is loaded.
         module_list_data = client.execute_command('MODULE LIST')
-        module_list_count = len(module_list_data)
-        assert module_list_count == 1
+        
+        # Only check count for local servers (bundle has multiple modules)
+        use_external = os.environ.get("VALKEY_EXTERNAL_SERVER", "false").lower() == "true"
+        if not use_external:
+            module_list_count = len(module_list_data)
+            assert module_list_count == 1
         module_loaded = False
         for module in module_list_data:
             if (module[b'name'] == b'bf'):
--- a/tests/test_bloom_basic.py
+++ b/tests/test_bloom_basic.py
@@ -9,13 +9,7 @@ class TestBloomBasic(ValkeyBloomTestCaseBase):
     def test_basic(self):
         client = self.server.get_new_client()
         # Validate that the valkey-bloom module is loaded.
-        module_list_data = client.execute_command('MODULE LIST')
-        
-        # Only check count for local servers (bundle has multiple modules)
-        use_external = os.environ.get("VALKEY_EXTERNAL_SERVER", "false").lower() == "true"
-        if not use_external:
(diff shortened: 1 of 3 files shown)


tests/valkey_bloom_test_case.py
```diff
@@ -13,11 +13,23 @@ class ValkeyBloomTestCaseBase(ValkeyTestCase):
...
-        server_path = f"{os.path.dirname(os.path.realpath(__file__))}/build/binaries/{os.environ['SERVER_VERSION']}/valkey-server"
-
-        self.server, self.client = self.create_server(testdir = self.testdir,  server_path=server_path, args=args)
-        logging.info("startup args are: %s", args)
+        use_external = os.environ.get("VALKEY_EXTERNAL_SERVER", "false").lower() == "true"
+        
+        if use_external:
+            # Use external server
+            external_host = os.environ.get("VALKEY_HOST", "localhost")
+            external_port = int(os.environ.get("VALKEY_PORT", "6379"))
```
KarthikSubbarao, line 23: Does this mean every test will run against the same server? Is that the intention?

There is no description on the PR. It would help if you briefly summarize it there :D

Nikhil-Manglore: We'll make a new server for each test, that will be done over on Valkey-Bundle. I'll update the description

KarthikSubbarao: Yes, but for the test framework, it will use a single server, right? 

On the valkey-bundle, you are starting up docker containers that will run a new instance of the test framework (each with a single external server) from what it sounds like

Nikhil-Manglore: Yes that's correct

KarthikSubbarao:
Question: Does this PR handle replication tests?

Nikhil-Manglore:
> Question: Does this PR handle replication tests?

As of now no that's the test I left out, will have to figure that out in the future

Edit: The PR now handles replication tests

tests/test_bloom_basic.py
```diff
@@ -10,8 +10,12 @@ def test_basic(self):
         client = self.server.get_new_client()
         # Validate that the valkey-bloom module is loaded.
         module_list_data = client.execute_command('MODULE LIST')
-        module_list_count = len(module_list_data)
-        assert module_list_count == 1
+        
+        # Only check count for local servers (bundle has multiple modules)
+        use_external = os.environ.get("VALKEY_EXTERNAL_SERVER", "false").lower() == "true"
+        if not use_external:
```
zackcam: Could potentially just save on the if and not check for module count as we check if module is loaded anyway

zackcam (approved):

tests/test_bloom_basic.py
Nikhil-Manglore: Note: We talked offline, will remove this check

PR #75, opened 2025-10-13
Fix AOF rewrite and restore BF tests

### LINKED ISSUE #74: AOF related integration tests failing for branch 1.0 and unstable
When I run `./build.sh`, I see the following tests are failing

```
tests/test_bloom_aofrewrite.py::TestBloomAofRewrite::test_basic_aofrewrite_and_restore[random-seed]
tests/test_bloom_aofrewrite.py::TestBloomAofRewrite::test_basic_aofrewrite_and_restore[fixed-seed] tests/test_bloom_aofrewrite.py::TestBloomAofRewrite::test_aofrewrite_bloomfilter_metrics[random-seed]
tests/test_bloom_aofrewrite.py::TestBloomAofRewrite::test_aofrewrite_bloomfilter_metrics[fixed-seed]
```

Also, I see there are some indentation errors that are leading to failures in tests.

arpitbbhayani:
The AOF rewrite tests were failing because bloom filter data was not being properly restored after server restart. The root cause was twofold:

1. BF.LOAD command was replicating during AOF loading, causing duplicate entries
2. Tests were not properly configuring AOF persistence for server restart

Changes:
- Modified BF.LOAD handler to skip replication when loading from AOF by checking must_obey_client() context
- Updated AOF tests to enable appendonly before adding data
- Added wait for initial AOF rewrite to prevent concurrent rewrites
- Set appendonly in server startup args for proper AOF loading on restart

Also includes minor fixes:
- Changed build.sh shebang to bash for better compatibility
- Updated pytest requirement to 7.4.3
- Fixed indentation in utils.rs documentation

Fixes #74

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env sh
 
 # Script to run format checks valkey-bloom module, build it and generate .so files, run unit and integration tests.
 
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/bin/bash
 
 # Script to run format checks valkey-bloom module, build it and generate .so files, run unit and integration tests.
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 valkey
-pytest==6
+pytest==7.4.3
--- a/src/bloom/utils.rs
+++ b/src/bloom/utils.rs
@@ -488,15 +488,15 @@ impl BloomObject {
     /// * `capacity` - The size of the initial filter in the bloom object.
     /// * `fp_rate` - the false positive rate for the bloom object
     /// * `validate_scale_to` - the capacity we check to see if it can scale to. If this method is called from BF.INFO this is set as -1 as we
-    ///                       want to check the maximum size we could scale up till
+    ///   want to check the maximum size we could scale up till
     /// * `tightening_ratio` - The tightening ratio of the object
     /// * `expansion` - The expanison rate of the object
     ///
     /// # Returns
     /// * i64 - The maximum capacity that can be reached if called from BF.INFO. If called from BF.INSERT the size it reached when it became greater than `validate_scale_to`
     /// * ValkeyError - Can return two different errors:
-    ///     VALIDATE_SCALE_TO_EXCEEDS_MAX_SIZE: When scaling to the wanted capacity would go over the bloom object memory limit
-    ///     VALIDATE_SCALE_TO_FALSE_POSITIVE_INVALID: When scaling to the wanted capacity would cause the false positive rate to reach 0
+    ///   VALIDATE_SCALE_TO_EXCEEDS_MAX_SIZE: When scaling to the wanted capacity would go over the bloom object memory limit
+    ///   VALIDATE_SCALE_TO_FALSE_POSITIVE_INVALID: When scaling to the wanted capacity would cause the false positive rate to reach 0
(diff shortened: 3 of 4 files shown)


src/bloom/command_handler.rs
```diff
@@ -817,24 +817,33 @@ pub fn bloom_filter_load(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
         None => {
             // if filter not exists, create it.
             let hex = value.to_vec();
-            let validate_size_limit = !must_obey_client(ctx);
+            let is_aof_loading = must_obey_client(ctx);
```
zackcam: Do we need this intermediary step here of having the variable of `is_aof_loading` could we not use `validate_size_limit ` on line 829?

enjoy-binbin:
@arpitbbhayani can you fix the DCO?

src/bloom/command_handler.rs
```diff
@@ -817,24 +817,33 @@ pub fn bloom_filter_load(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
...
+                    seed: bloom.seed(),
+                    items: &input_args[idx..],
+                })
+            } else {
+                None
             };
             match filter_key.set_value(&BLOOM_TYPE, bloom) {
                 Ok(_) => {
-                    replicate_and_notify_events(ctx, filter_name, false, true, replicate_args);
+                    // Only replicate if not loading from AOF/replication
```
KarthikSubbarao: From what I remember reading in the core, calling the ValkeyModule_Replicate Module API in the core has no effect if the role is a replica (if we received it from the replication stream). I would assume it is the same for AOF. Could we check if this is neccesary?

src/bloom/command_handler.rs
```diff
@@ -817,24 +817,33 @@ pub fn bloom_filter_load(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyRe
...
+                    if let Some(args) = replicate_args {
+                        replicate_and_notify_events(ctx, filter_name, false, true, args);
+                    }
```
KarthikSubbarao: One side effect from this change is that, we only post a notification on bloom object creations on primary nodes. Not on replica nodes

src/bloom/command_handler.rs
arpitbbhayani: I went through the code and realized we do not need this change. I wrongly analyzed the flow, but now I have reverted the change and validated the flow.

src/bloom/command_handler.rs
arpitbbhayani: Yes. I went though the code and understood the flow end-to-end and realized this was a redundant check. given that it passed the test I thought this is needed. Now, I have reverted this change. thanks.

src/bloom/command_handler.rs
arpitbbhayani: yup. removed this change. ref comment above.

arpitbbhayani:
Thank you so much for reviewing the code. Apologies that it took some time for me to come back.
I have addressed all the concerns and made the changes.

Now, the PR only fixes the AOF tests that were failing and other changes have been reverted.

build.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/bin/bash
```
KarthikSubbarao: Suggesting we retain `#!/usr/bin/env sh` for better compatibility 

```suggestion
#!/usr/bin/env sh
```

KarthikSubbarao:
Thank you for your help with this PR, @arpitbbhayani :D

PR #78, opened 2025-11-23
feat: Add command metadata and update tests

bravo1goingdark:
# Add `#[valkey_command]` Metadata & KeySpecs for Bloom Filter Commands

This PR adds the `#[valkey_command]` macro to all Bloom Filter commands in `src/lib.rs`, enabling proper `SetCommandInfo` support through the `valkeymodule-rs` crate.  
The macro now provides Valkey with complete metadata, including:

- Command name  
- Flags  
- Arity  
- **KeySpec definitions** (key position, read/write semantics, key ranges)

As a result, `COMMAND INFO BF.*` now correctly reports the KeySpec metadata for each Bloom Filter command.

---

### Added `#[valkey_command]` to all Bloom Filter commands:
- `BF.ADD`  
- `BF.MADD`  
- `BF.EXISTS`  
- `BF.MEXISTS`  
- `BF.CARD`  
- `BF.RESERVE`  
- `BF.INFO`  
- `BF.INSERT`  
- `BF.LOAD`  

### Implemented complete KeySpecs for each command:
- `begin_search`
- `find_keys`
- Key flags (`rw`, `insert`, `update`, `access`)

### Updated tests in `tests/test_bloom_command.py`:
- Adjusted expected arities  
- Added KeySpec validation checks  

### Minor cleanup:
- Improved string formatting in:
  - `src/bloom/data_type.rs`
  - `src/bloom/utils.rs`

---

## Notes / Limitations

The current version of `valkeymodule-rs` does **not** support argument metadata (`args:`) in `SetCommandInfo`.  
Because of this limitation, this PR implements all supported metadata (KeySpecs), but cannot yet add full CLI argument autocomplete like the built-in `SET` command.

Once `args` support is added upstream in `valkeymodule-rs`, I can follow up with another PR to provide complete argument specifications for full autocomplete.

---

## Verification

Example output of `COMMAND INFO BF.ADD`, showing that KeySpec metadata is now present:

KeySpec Output Screenshot

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -54,9 +54,9 @@ fn deinitialize(_ctx: &Context) -> Status {
     flags: [ReadOnly, Fast],
     arity: 3,
     key_spec: [{
-        flags: [ReadOnly, Access],
         begin_search: Index({ index: 1 }),
         find_keys: Range({ last_key: 1, steps: 1, limit: 0 }),
+        flags: [ReadOnly, Access],
     }],
 })]
 fn bloom_exists_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
@@ -72,9 +72,9 @@ fn bloom_exists_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult 
     flags: [ReadOnly, Fast],
     arity: -3,
     key_spec: [{
-        flags: [ReadOnly, Access],
         begin_search: Index({ index: 1 }),
         find_keys: Range({ last_key: 1, steps: 1, limit: 0 }),
+        flags: [ReadOnly, Access],
     }],
 })]
 fn bloom_mexists_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
@@ -90,9 +90,9 @@ fn bloom_mexists_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult
     flags: [Write, DenyOOM, Fast],
     arity: 3,
     key_spec: [{
-        flags: [ReadWrite, Insert, Update],
         begin_search: Index({ index: 1 }),
         find_keys: Range({ last_key: 1, steps: 1, limit: 0 }),
+        flags: [ReadWrite, Insert],
     }],
 })]
 fn bloom_add_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
@@ -108,9 +108,9 @@ fn bloom_add_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     flags: [Write, DenyOOM, Fast],
     arity: -3,
     key_spec: [{
-        flags: [ReadWrite, Insert, Update],
         begin_search: Index({ index: 1 }),
         find_keys: Range({ last_key: 1, steps: 1, limit: 0 }),
+        flags: [ReadWrite, Insert],
     }],
 })]
 fn bloom_madd_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
@@ -126,9 +126,9 @@ fn bloom_madd_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     flags: [ReadOnly, Fast],
     arity: 2,
     key_spec: [{
-        flags: [ReadOnly, Access],
         begin_search: Index({ index: 1 }),
         find_keys: Range({ last_key: 1, steps: 1, limit: 0 }),
+        flags: [ReadOnly, Access],
     }],
 })]
 fn bloom_card_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
@@ -144,9 +144,9 @@ fn bloom_card_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     flags: [Write, DenyOOM, Fast],
     arity: -4,
     key_spec: [{
-        flags: [ReadWrite, Insert],
         begin_search: Index({ index: 1 }),
         find_keys: Range({ last_key: 1, steps: 1, limit: 0 }),
+        flags: [ReadWrite, Insert],
     }],
 })]
 fn bloom_reserve_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
@@ -162,9 +162,9 @@ fn bloom_reserve_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult
     flags: [ReadOnly, Fast],
     arity: -2,
     key_spec: [{
-        flags: [ReadOnly, Access],
         begin_search: Index({ index: 1 }),
         find_keys: Range({ last_key: 1, steps: 1, limit: 0 }),
+        flags: [ReadOnly, Access],
     }],
 })]
 fn bloom_info_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
@@ -181,9 +181,9 @@ fn bloom_info_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     flags: [Write, DenyOOM, Fast],
     arity: -2,
     key_spec: [{
-        flags: [ReadWrite, Insert, Update],
         begin_search: Index({ index: 1 }),
         find_keys: Range({ last_key: 1, steps: 1, limit: 0 }),
+        flags: [ReadWrite, Insert],
     }],
 })]
 fn bloom_insert_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
@@ -200,9 +200,9 @@ fn bloom_insert_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult 
     flags: [Write, DenyOOM],
     arity: 3,
     key_spec: [{
-        flags: [ReadWrite, Insert],
         begin_search: Index({ index: 1 }),
         find_keys: Range({ last_key: 1, steps: 1, limit: 0 }),
+        flags: [ReadWrite, Insert],
     }],
 })]
 fn bloom_load_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -111,17 +229,7 @@ valkey_module! {
     acl_categories: [
         "bloom",
     ]
-    commands: [
-        ["BF.ADD", bloom_add_command, "write fast deny-oom", 1, 1, 1, "fast write bloom"],
```
KarthikSubbarao, line 117: Does this remove the `bloom` and other ACL categories from all the commands? Can you run the integration tests to validate if this passes?

You can check the `valkey_module!` macro to see how the command creation works in there. Specifically, how the ACL category is set through it

bravo1goingdark:
> Added per-command ACL categories in `initialize( )` since `#[valkey_command]` doesn't set them. All 98 integration tests pass

@KarthikSubbarao

bravo1goingdark:
should i close this pr ?

KarthikSubbarao:
We will take a look at the latest PR here.....

@zackcam - Do you have some time this week to help review this?

zackcam:
> We will take a look at the latest PR here.....
> 
> @zackcam - Do you have some time this week to help review this?

I should be able to take a look this week at it. I'll try and make time to look at it

src/lib.rs
```diff
@@ -45,48 +67,165 @@ fn deinitialize(_ctx: &Context) -> Status {
...
+#[valkey_command({
+    name: "BF.EXISTS",
+    summary: "Determines if the bloom filter contains the specified item",
+    complexity: "O(N), where N is the number of hash functions used by the bloom filter.",
+    since: "1.0.0",
+    flags: [ReadOnly, Fast],
+    arity: 3,
+    key_spec: [{
+        begin_search: Index({ index: 1 }),
+        find_keys: Range({ last_key: 1, steps: 1, limit: 0 }),
```
zackcam, line 79: Maybe I'm reading wrong: https://valkey.io/topics/key-specs/ but should these have last_key as 0? so like 0, 1, 0 would be the same for all I think?

zackcam: I was looking at:
>The SET command has a range of 0, 1 and 0.

Which I think would be the same sort as we do?

tests/test_bloom_command.py
```diff
@@ -9,14 +9,24 @@ def verify_command_arity(self, command, expected_arity):
...
+        self.verify_command_arity('BF.EXISTS', 3)
+        self.verify_command_arity('BF.ADD', 3)
+        self.verify_command_arity('BF.MEXISTS', -3)
+        self.verify_command_arity('BF.MADD', -3)
+        self.verify_command_arity('BF.CARD', 2)
+        self.verify_command_arity('BF.RESERVE', -4)
+        self.verify_command_arity('BF.INFO', -2)
+        self.verify_command_arity('BF.INSERT', -2)
+
+    def test_bloom_command_keyspecs_present(self):
```
zackcam, line 21: For this do we want to check that keyspecs are correct not just present?

src/lib.rs
```diff
@@ -45,48 +67,165 @@ fn deinitialize(_ctx: &Context) -> Status {
...
+    }],
+})]
 fn bloom_mexists_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     command_handler::bloom_filter_exists(ctx, &args, true)
 }
 
 /// Command handler for BF.ADD <key> <item>
+#[valkey_command({
+    name: "BF.ADD",
+    summary: "Add a single item to a bloom filter; creates the filter if it does not exist",
```
zackcam, line 108: I think we want these to be the same as commands json files. If those aren't punctually correct I'm fine with us changing those to match these. (This one is same idea just different wording on it)

zackcam (commented):
Just two actual comments really, testing one is a nit. Once these are done I'll do another review and if it looks good I'll approve!

Thanks for working on this!

src/lib.rs
bravo1goingdark: You were reading it right: `lastkey` is relative to `begin_search`, so `0` means only the first key. Updated all 9 BF.* commands to `Range({ last_key: 0, steps: 1, limit: 0 })`.

This also fixes the legacy values derived from it — `COMMAND INFO` now reports first/last/step as `1, 1, 1` instead of `1, 2, 1` (previously the value argument was reported as a key). The test now asserts both the key spec range and those legacy positions.

tests/test_bloom_command.py
bravo1goingdark: Good call — replaced `test_bloom_command_keyspecs_present` with `test_bloom_command_keyspecs`, which now checks the actual values for all 9 commands instead of presence for 4:

- key spec flags (RW/insert for writes, RO/access for reads)
- `begin_search` type `index` with index 1
- `find_keys` type `range` with `0, 1, 0`
- legacy `first_key_pos`/`last_key_pos`/`step_count` = 1/1/1

Verified the assertions actually bite: with `last_key` back at 1 the test fails on `last_key_pos`.

src/lib.rs
bravo1goingdark: Aligned the three that differed: `BF.ADD`, `BF.MADD` and `BF.LOAD` summaries now use the exact wording from `src/commands/*.json` (the other six already matched).

While diffing the two, the JSON arity for `BF.MADD`/`BF.MEXISTS` said `3` but both are variadic, so I changed those to `-3` to match the macro and what the server reports. Happy to split that into a separate change if you'd rather keep the JSON edits out of this PR.

bravo1goingdark:
Addressed all three review comments in 42fcf5f:

1. **Key spec range** — all 9 BF.* commands now use `last_key: 0` (`0, 1, 0`), matching `SET`. Legacy `COMMAND INFO` positions are now 1/1/1 instead of 1/2/1.
2. **Key spec test** — now validates flags, `begin_search`, `find_keys` range and legacy key positions for all 9 commands (was: presence check for 4). Confirmed it fails if `last_key` is reverted to 1. Also added the missing `BF.LOAD` arity assertion.
3. **Summaries** — `BF.ADD`, `BF.MADD`, `BF.LOAD` now use the wording from the `src/commands/*.json` files; the other six already matched. Also corrected `arity` for `BF.MADD`/`BF.MEXISTS` in those JSONs from `3` to `-3` since both are variadic.

tests/test_bloom_command.py
```diff
@@ -9,14 +9,93 @@ def verify_command_arity(self, command, expected_arity):
...
+        assert range_spec['lastkey'] == 0, \
+            f"lastkey for '{command}' is {range_spec['lastkey']}, expected 0 (a single key)"
+        assert range_spec['keystep'] == 1, \
+            f"keystep for '{command}' is {range_spec['keystep']}, expected 1"
+        assert range_spec['limit'] == 0, \
+            f"limit for '{command}' is {range_spec['limit']}, expected 0"
+
+    def test_bloom_command_keyspecs(self):
+        for command, expected_flags in self.EXPECTED_BLOOM_KEYSPECS.items():
+            self.verify_command_keyspec(command, expected_flags)
```
zackcam:
```
        keys = self.client.execute_command('COMMAND', 'GETKEYS', command, 'k', 'item')
        assert keys == [b'k'], f"{command} GETKEYS returned {keys}, expected [b'k']"

        info = self.client.execute_command('COMMAND', 'INFO', command)[command]

        assert (info['first_key_pos'], info['last_key_pos'], info['step_count']) \
            == (1, 1, 1), f"{command} legacy key positions wrong: {info}"

        specs = info['key_specifications']
        assert len(specs) == 1, f"{command} expected 1 key spec, got {len(specs)}"
        flags = {f.decode() if isinstance(f, bytes) else f
                 for f in specs[0]['flags']}
        assert flags == expected_flags, \
            f"{command} flags {sorted(flags)}, expected {sorted(expected_flags)}"
```
Maybe something like this would be cleaner for testing. Right now seems quite long of a test.

zackcam (commented):
Just one comment on the test, I think utilizing the https://valkey.io/commands/command-getkeys/ might be a nicer way of checking

tests/test_bloom_command.py
bravo1goingdark: Done in a9344bf — GETKEYS is the main check now and I dropped the field-by-field key spec assertions, so the test went from ~75 lines to ~30. It still catches the old bug: with `last_key` at 1, `GETKEYS BF.ADD k item` returns `[k, item]`.

Two heads-ups from running it: sending GETKEYS as split args (`execute_command('COMMAND', 'GETKEYS', ...)`) errors because valkey-py runs its COMMAND INFO parser over the reply — one string works, same as the ACL test. And GETKEYS validates arity, so each command needs its own sample args.

zackcam (approved):

tests/test_bloom_command.py
zackcam: Perfect thank you, yeah I didn't actually make the changes locally to test my snippet sorry just wanted to give the general idea but looks good now. Will run the workflows and see about getting it merged

bravo1goingdark: ok :+1:

tests/test_bloom_replication.py
```diff
@@ -178,8 +178,9 @@ def test_replication_behavior(self):
             except ResponseError as e:
                 pass
             primary_cmd_stats = self.client.info("Commandstats")['cmdstat_' + prefix]
-            assert primary_cmd_stats["calls"] == 1
-            assert primary_cmd_stats["failed_calls"] == 1
+            was_rejected = primary_cmd_stats.get("rejected_calls", 0) == 1
+            was_failed = primary_cmd_stats.get("calls", 0) == 1 and primary_cmd_stats.get("failed_calls", 0) == 1
+            assert was_rejected or was_failed
```
KarthikSubbarao: Why did this have to change from the original?

bravo1goingdark: Because this PR gives the commands real arity metadata, so the server now catches the two arity mistakes before the module ever runs.

Before, all four reached the module: `calls=1, failed_calls=1`. Now `BF.ADD key item1 item2` and `BF.MADD key` are rejected at the server: `rejected_calls=1, calls=0`. `BF.RESERVE` (bad error rate) and `BF.INSERT` (bad capacity) still run and fail inside the module, so they keep the original numbers.

I've replaced the `rejected or failed` with the exact expectation per command in 06f3e5e, so the reason is visible in the test. Checked on valkey 8.0, 8.1 and unstable — the replication test passes on all three.

PR #83, opened 2026-02-23
Merge from unstable into 1.0

KarthikSubbarao:
Merge from unstable into 1.0

### DIFF
--- a/00-RELEASENOTES
+++ b/00-RELEASENOTES
@@ -1,5 +1,26 @@
-Valkey Bloom 1.0.0 release notes
+Valkey Bloom 1.0 release notes
 ========================
+
+Upgrade urgency levels:
+
+| Level    | Meaning                                                             |
+|----------|---------------------------------------------------------------------|
+| LOW      | No need to upgrade unless there are new features you want to use.   |
+| MODERATE | Program an upgrade of the server, but it's not urgent.              |
+| HIGH     | There is a critical bug that may affect a subset of users. Upgrade! |
+| CRITICAL | There is a critical bug affecting MOST USERS. Upgrade ASAP.         |
+| SECURITY | There are security fixes in the release.                            |
+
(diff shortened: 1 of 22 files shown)


build.sh
```diff
@@ -22,7 +31,7 @@ if [ -z "$SERVER_VERSION" ]; then
     export SERVER_VERSION="unstable"
 fi
 
-if [ "$SERVER_VERSION" != "unstable" ] && [ "$SERVER_VERSION" != "8.0" ] && [ "$SERVER_VERSION" != "8.1" ]; then
+if [ "$SERVER_VERSION" != "unstable" ] && [ "$SERVER_VERSION" != "8.0" ] && [ "$SERVER_VERSION" != "8.1" ] && [ "$SERVER_VERSION" != "9.0" ]; then
```
zackcam, line 34: I wonder of we should just make this a greater than. Save work in the future

00-RELEASENOTES
```diff
@@ -1,6 +1,27 @@
-Valkey Bloom 1.0.0 release notes
+Valkey Bloom 1.0 release notes
```
zackcam, line 1: Should this be 1.0.1?

zackcam (approved):

00-RELEASENOTES
KarthikSubbarao: Oh that's just the title of the file - major and minor version. (Following the convention from the core)

The contents of the file include all the patch versions of this category

build.sh
KarthikSubbarao: Yes, we should make that change in the unstable branch.

Also, I think we should just test valkey-bloom against the latest engine version (9.0) and unstable

and have engine past versions (8.0 and 8.1) tested from a GitHub daily job.

Otherwise, there will be too many versions to test

PR #86, opened 2026-04-07
Fix replica crash on partial-success BF.MADD/BF.INSERT replication

madolson:
## Problem

When `BF.MADD` or `BF.INSERT` partially succeeds on a nearly-full nonscaling bloom filter (some items added, then `NonScalingFilterFull` error), the command is replicated verbatim to replicas. On the replica, the error-triggering items cause `VM_ReplyWithError` back to the primary — a fatal protocol violation that crashes the replica process.

Stack trace from the crash (from `BUG.md`):
```
Guru Meditation: This replica panicked sending an error to its primary
after processing the command 'BF.MADD'
```

## Root Cause

`handle_bloom_add` in multi mode would add items one by one, and on error push a `ValkeyValue::StaticError` into the result array and `break`. Since some items were already added (`add_succeeded = true`), the command was replicated verbatim via `replicate_verbatim()`. The replica would then hit the same `NonScalingFilterFull` error and call `VM_ReplyWithError`, which is not allowed on replicas.

The errors `NonScalingFilterFull` and `MaxNumScalingFilters` are NOT gated by `validate_size_limit`, so they fire on replicas too (unlike `ExceedsMaxBloomSize` which is safely skipped).

## Fix

Added `BloomObject::validate_add_items()` which counts how many items are genuinely new (not already in the filter) and verifies the filter can accommodate them all before any are added. If validation fails, the entire command is rejected atomically on the primary — no items added, no replication, no replica crash.

The validation only checks for errors that are NOT gated by `validate_size_limit` (`NonScalingFilterFull`, `MaxNumScalingFilters`, `FalsePositiveReachesZero`, `BadCapacity`) since `ExceedsMaxBloomSize` is already safely skipped on replicas.

This matches the documented behavior for `BF.MADD`: "An error will occur if... an item is being added to a full non scaling filter."

## Testing

- Two new replication tests (`test_madd_nonscaling_filter_full_replication`, `test_insert_nonscaling_filter_full_replication`) that set `propagation-error-behavior panic` on the replica and verify it stays alive
- Updated `validate_nonscaling_failure` to expect atomic rejection (top-level error) instead of partial inline errors
- Updated `test_too_large_bloom_obj` to handle both `ResponseError` and inline errors
- All 94 integration tests and 18 unit tests pass

### DIFF
--- a/src/bloom/command_handler.rs
+++ b/src/bloom/command_handler.rs
@@ -29,41 +29,52 @@ fn handle_bloom_add(
 ) -> Result<ValkeyValue, ValkeyError> {
     match multi {
         true => {
-            // Count how many unique items are genuinely new (not already in the filter).
+            // Determine which items are genuinely new (not already in the filter).
             // Duplicates within the same command are only counted once since the first
             // occurrence will be added and subsequent ones will return 0 (already exists).
+            let items: Vec<&[u8]> = args
+                .iter()
+                .take(argc)
+                .skip(item_idx)
+                .map(|a| a.as_slice())
+                .collect();
             let mut new_item_count: i64 = 0;
             let mut seen: std::collections::HashSet<&[u8]> =
-                std::collections::HashSet::with_capacity(argc - item_idx);
-            for arg in args.iter().take(argc).skip(item_idx) {
-                let item = arg.as_slice();
-                if seen.insert(item) && !bf.item_exists(item) {
+                std::collections::HashSet::with_capacity(items.len());
+            // is_new[i] is true if items[i] is the first unique occurrence AND not in the filter.
+            let mut is_new = Vec::with_capacity(items.len());
+            for &item in &items {
+                let new = seen.insert(item) && !bf.item_exists(item);
+                if new {
                     new_item_count += 1;
                 }
+                is_new.push(new);
             }
             // Validate that all new items can be added without error.
             // If this fails, the entire command is rejected — no partial adds.
             if let Err(err) = bf.validate_add_items(new_item_count) {
                 return Err(ValkeyError::Str(err.as_str()));
             }
-            // All items are guaranteed to succeed. Add them.
-            let mut result = Vec::with_capacity(argc - item_idx);
-            let mut curr_cmd_idx = item_idx;
-            while curr_cmd_idx < argc {
-                let item = args[curr_cmd_idx].as_slice();
-                match bf.add_item(item, validate_size_limit) {
-                    Ok(add_result) => {
-                        if add_result == 1 {
-                            *add_succeeded = true;
+            // All items are guaranteed to succeed. Add them using add_item_unchecked
+            // to avoid calling item_exists again.
+            let mut result = Vec::with_capacity(items.len());
+            for (i, &item) in items.iter().enumerate() {
+                if is_new[i] {
+                    match bf.add_item_unchecked(item, validate_size_limit) {
+                        Ok(add_result) => {
+                            if add_result == 1 {
+                                *add_succeeded = true;
+                            }
+                            result.push(ValkeyValue::Integer(add_result));
                         }
-                        result.push(ValkeyValue::Integer(add_result));
+                        Err(err) => {
+                            // This should not happen after validation, but handle defensively.
+                            return Err(ValkeyError::Str(err.as_str()));
+                        }
                     }
-                    Err(err) => {
-                        // This should not happen after validation, but handle defensively.
-                        return Err(ValkeyError::Str(err.as_str()));
-                    }
-                };
-                curr_cmd_idx += 1;
+                } else {
+                    result.push(ValkeyValue::Integer(0));
+                }
             }
             Ok(ValkeyValue::Array(result))
         }
--- a/src/bloom/command_handler.rs
+++ b/src/bloom/command_handler.rs
@@ -13,6 +13,11 @@ use valkey_module::{Context, ValkeyError, ValkeyResult, ValkeyString, ValkeyValu
 /// Helper function used to add items to a bloom object. It handles both multi item and single item add operations.
 /// It is used by any command that allows adding of items: BF.ADD, BF.MADD, and BF.INSERT.
 /// Returns the result of the item add operation on success as a ValkeyValue and a ValkeyError on failure.
+///
+/// For multi-item operations, the function validates upfront that all new items can be added
+/// without error (e.g. NonScalingFilterFull). If validation fails, the entire command is rejected
+/// before any items are added. This ensures atomicity and prevents partial-success replication
+/// that would cause replicas to crash.
 fn handle_bloom_add(
     args: &[ValkeyString],
     argc: usize,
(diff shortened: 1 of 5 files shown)


madolson:
@copilot apply changes based on the comments in [this thread](https://github.com/valkey-io/valkey-bloom/pull/86#pullrequestreview-4065232050)

src/bloom/command_handler.rs
```diff
@@ -24,6 +29,21 @@ fn handle_bloom_add(
...
+                let item = arg.as_slice();
+                if !bf.item_exists(item) {
+                    new_item_count += 1;
+                }
+            }
+            // Validate that all new items can be added without error.
+            // If this fails, the entire command is rejected — no partial adds.
+            if let Err(err) = bf.validate_add_items(new_item_count) {
+                return Err(ValkeyError::Str(err.as_str()));
+            }
```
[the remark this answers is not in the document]
madolson: I did think about this, and originally decided it was an OK tradeoff for performance. Everything on this path will degrade performance, but checking locally it didn't make much of a difference, so will incorporate it.

tests/test_bloom_replication.py
```diff
@@ -182,6 +182,117 @@ def test_replication_behavior(self):
             assert primary_cmd_stats["failed_calls"] == 1
             assert ('cmdstat_' + prefix) not in self.replicas[0].client.info("Commandstats")
 
+    def _find_new_items(self, key, count, start_offset=1000):
```
zackcam, line 185: Other utility functions for tests are in valkey_bloom_test_case but think this is fine as not used anywhere else atm

zackcam (approved):

src/bloom/command_handler.rs
```diff
@@ -24,6 +29,24 @@ fn handle_bloom_add(
...
         true => {
+            // Count how many unique items are genuinely new (not already in the filter).
+            // Duplicates within the same command are only counted once since the first
+            // occurrence will be added and subsequent ones will return 0 (already exists).
+            let mut new_item_count: i64 = 0;
+            let mut seen: std::collections::HashSet<&[u8]> =
+                std::collections::HashSet::with_capacity(argc - item_idx);
+            for arg in args.iter().take(argc).skip(item_idx) {
+                let item = arg.as_slice();
+                if seen.insert(item) && !bf.item_exists(item) {
```
KarthikSubbarao: It would be nice to not have to call `item_exists` twice per item.

Now, it is called here and also in `add_item`


`item_exists`'s cost gets larger with scaling filters and also with larger number of hash fns

src/bloom/utils.rs
```diff
@@ -303,6 +303,55 @@ impl BloomObject {
...
+    ///
+    /// ExceedsMaxBloomSize is intentionally NOT checked here because it is already gated by
+    /// validate_size_limit and skipped on replicas.
+    pub fn validate_add_items(&self, count: i64) -> Result<(), BloomError> {
+        if count == 0 {
+            return Ok(());
+        }
+        let last_filter = match self.filters.last() {
+            Some(f) => f,
+            None => return Ok(()),
```
KarthikSubbarao: Nit: Unexpected

src/bloom/utils.rs
```diff
@@ -303,6 +303,55 @@ impl BloomObject {
...
+        while items_remaining > 0 {
+            if num_filters == configs::BLOOM_NUM_FILTERS_PER_OBJECT_LIMIT_MAX {
+                return Err(BloomError::MaxNumScalingFilters);
+            }
+            let _new_fp_rate =
+                Self::calculate_fp_rate(self.fp_rate, num_filters, self.tightening_ratio)?;
+            let new_capacity = match prev_capacity.checked_mul(self.expansion.into()) {
+                Some(c) if c > 0 => c,
+                _ => return Err(BloomError::BadCapacity),
+            };
```
KarthikSubbarao: Nit: This block of code is common to add_item

And it is also missing the memory usage checks.

It might be worth extracting this out and using it across two places

src/bloom/utils.rs
```diff
@@ -303,6 +303,55 @@ impl BloomObject {
...
+    /// Check whether the bloom object can accommodate `count` new item additions without hitting
+    /// errors that are not gated by validate_size_limit (i.e., errors that would also fire on
+    /// replicas and cause them to crash). Specifically:
+    /// - NonScalingFilterFull: nonscaling filter has no room
+    /// - MaxNumScalingFilters: too many sub-filters
+    /// - FalsePositiveReachesZero / BadCapacity: scale-out arithmetic failures
+    ///
+    /// ExceedsMaxBloomSize is intentionally NOT checked here because it is already gated by
+    /// validate_size_limit and skipped on replicas.
+    pub fn validate_add_items(&self, count: i64) -> Result<(), BloomError> {
```
KarthikSubbarao, line 311: Looks like this function is missing memory usage object checks.

A scale out can fails due to exceeding the memory usage limit per bloom object (configured)

src/bloom/command_handler.rs
```diff
@@ -24,6 +29,24 @@ fn handle_bloom_add(
...
+                    new_item_count += 1;
+                }
+            }
+            // Validate that all new items can be added without error.
+            // If this fails, the entire command is rejected — no partial adds.
+            if let Err(err) = bf.validate_add_items(new_item_count) {
```
KarthikSubbarao, line 55: Can we check if we really need a new function or if we can try and re-use `calculate_max_scaled_capacity` here?

For context:
```
    /// This method is called from two different bloom commands: BF.INFO and BF.INSERT. The functionality varies slightly on which command it
    /// is called from. When called from BF.INFO, this method is used to find the maximum possible size that the bloom object could scale to
    /// without throwing an error. When called from BF.INSERT, this method is used to determine if it is possible to reach the provided `validate_scale_to`.
```

src/bloom/utils.rs
KarthikSubbarao: (This should be covered if we use `calculate_max_scaled_capacity`)

src/bloom/utils.rs
```diff
@@ -303,13 +303,49 @@ impl BloomObject {
...
+            return Err(BloomError::NonScalingFilterFull);
+        }
+        let needed_total = self.cardinality() + count;
+        Self::calculate_max_scaled_capacity(
+            self.starting_capacity(),
+            self.fp_rate,
+            needed_total,
+            self.tightening_ratio,
+            self.expansion,
+        )?;
```
KarthikSubbarao: This looks good for the validation logic. But using `?` will return back an error message which might not be specific to the `BF.MADD` command used by a customer.

Example:
```
-ERR provided VALIDATESCALETO causes bloom object to exceed memory limit
```

Or:
```
-ERR provided VALIDATESCALETO causes false positive to degrade to 0
```

What do you think? 
Should we fix the error message in these cases / convert it accordingly into the following?

```
-ERR operation exceeds bloom object memory limit
-ERR false positive degrades to 0 on scale out
```

Or should we change the `calculate_max_scaled_capacity` fn itself to use consistent error messages for both cases?

madolson: Added `map_err` to translate them back to the user-facing variants                    
  (`ExceedsMaxBloomSize`, `FalsePositiveReachesZero`) so clients see the same messages they'd get from `add_item`.

KarthikSubbarao (approved):

Issue #90, opened 2026-04-17
RESTORE with out-of-range tightening_ratio crashes server on BF.INFO (SIGABRT)


madolson:
## Summary

A crafted RESTORE payload can inject a bloom object with `tightening_ratio=2.0`, which is outside the valid range. Subsequent `BF.INFO` on that key triggers an assertion failure in `calculate_max_scaled_capacity` → `calculate_fp_rate`, causing SIGABRT and crashing the server.

## Root Cause

`load_from_rdb` in `src/bloom/data_type.rs` does **not** validate `tightening_ratio` after reading it from the RDB stream. In contrast, `decode_object` in `src/bloom/utils.rs` (used by `BF.LOAD`) correctly validates:

```rust
if !(values.2 > BLOOM_TIGHTENING_RATIO_MIN
    && values.2 < BLOOM_TIGHTENING_RATIO_MAX)
{
    return Err(BloomError::ErrorRateRange);
}
```

This check is missing from `load_from_rdb`, so RESTORE and RDB file loading accept invalid values.

When `BF.INFO` is called, `calculate_max_scaled_capacity` computes `fp_rate * tightening_ratio^N`. With `tightening_ratio=2.0`, the false positive rate grows past 1.0, hitting an assertion failure → `abort()`.

## Reproducer

```python
#!/usr/bin/env python3
"""Crash reproducer: RESTORE bloom with tightening_ratio=2.0, then BF.INFO."""
import socket, struct, time, sys

PORT = int(sys.argv[1]) if len(sys.argv) > 1 else 6379

def resp_send(*args):
    s = socket.socket(); s.settimeout(3); s.connect(('127.0.0.1', PORT))
    hdr = f'*{len(args)}\r\n'.encode()
    body = b''
    for a in args:
        if isinstance(a, str): a = a.encode()
        body += f'${len(a)}\r\n'.encode() + a + b'\r\n'
    s.sendall(hdr + body); time.sleep(0.5)
    try: r = s.recv(4096)
    except: r = b''
    s.close(); return r

def alive():
    try:
        s = socket.socket(); s.settimeout(2); s.connect(('127.0.0.1', PORT))
        s.sendall(b'*1\r\n$4\r\nPING\r\n'); time.sleep(0.3)
        r = s.recv(4096); s.close(); return b'PONG' in r
    except: return False

# CRC64 (Valkey polynomial)
POLY = 0xad93d23594c935a9
def _reflect64(v):
[...]

PR #91, opened 2026-04-17
Adding validation on tightening_ratio and fp_rate for load

zackcam:
Fix for https://github.com/valkey-io/valkey-bloom/issues/90, this will check for both doubles loaded to make sure they fall in the valid range for bloom filters

### DIFF
--- a/src/bloom/data_type.rs
+++ b/src/bloom/data_type.rs
@@ -1,6 +1,9 @@
 use crate::bloom::utils::BloomFilter;
 use crate::bloom::utils::BloomObject;
 use crate::configs;
+use crate::configs::{
+    BLOOM_FP_RATE_MAX, BLOOM_FP_RATE_MIN, BLOOM_TIGHTENING_RATIO_MAX, BLOOM_TIGHTENING_RATIO_MIN,
+};
 use crate::wrapper::bloom_callback;
 use crate::MODULE_NAME;
 use std::os::raw::c_int;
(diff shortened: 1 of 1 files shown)


src/bloom/data_type.rs
KarthikSubbarao, line 93: Should we validate if the RDB claims the seed is not random, but it does not match the fixed seed?

KarthikSubbarao: Oh, I think that is already covered below...
```
          if !is_seed_random && filter.seed() != configs::FIXED_SEED {
                logging::log_warning("Failed to restore bloom object: Object in fixed seed mode, but seed does not match FIXED_SEED.");
                return None;
            }
```

src/bloom/data_type.rs
```diff
@@ -69,10 +72,21 @@ impl ValkeyDataType for BloomObject {
         let Ok(fp_rate) = raw::load_double(rdb) else {
             return None;
         };
-
+        if !(fp_rate > BLOOM_FP_RATE_MIN && fp_rate < BLOOM_FP_RATE_MAX) {
```
KarthikSubbarao, line 75: Should we add the same validation for BF.LOAD for the fp/tightening ratio params?

zackcam: We should already have this coverage in BF.LOAD. Currently it looks like:

```
if !(values.1 > BLOOM_FP_RATE_MIN && values.1 < BLOOM_FP_RATE_MAX) {
    return Err(BloomError::ErrorRateRange);
}
if !(values.2 > BLOOM_TIGHTENING_RATIO_MIN
    && values.2 < BLOOM_TIGHTENING_RATIO_MAX)
{
    return Err(BloomError::ErrorRateRange);
}
```

KarthikSubbarao (approved):

PR #92, opened 2026-04-22
Fixing bug in bloom defrag, adding digest checks to integration test, create new unit test that catches before mentioned bug

zackcam:
## Defrag issue fix
One of the last steps we do in defragmention is a `into_boxed_slice` specifically `let filters_ptr = Box::into_raw(filters_vec.into_boxed_slice()) as *mut c_void;`. This function will drop excess capacity based on documentation [listed here](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.into_boxed_slice). However as rust vec's scale they don't always have capacity equal to length so at some point if we scale enough we could hit where capacity > len. For example lets say length of 3 but capacity of 4 this will drop the extra 1 capacity that we had and just make a vec of 3. This becomes an issue as after this we then recreate the vec with the following:
```
        *bloom_object.filters_mut() = unsafe {
            Vec::from_raw_parts(
                defragged_filters_ptr as *mut Box<BloomFilter>,
                num_filters,
                filters_capacity,
            )
        }; 
``` 
This recreates the vec but with the original capacity. However our `let defragged_filters_ptr = defrag.alloc(filters_ptr);` will only take memory of length not capacity. This could lead to the bloom filter Vec thinking that it owned more memory than it actually did potentially causing memory corruptions on future scale outs. The fix for this is simple we instead want to recreate the Vec with the length as capacity instead of capacity. Which would give us the new code of:
```
        *bloom_object.filters_mut() = unsafe {
            Vec::from_raw_parts(
                filters_ptr as *mut Box<BloomFilter>,
                num_filters,
                num_filters,
            )
        };
```

## Bloom testing

Amended the original bloom test to also include digest checks

Also adds `test_bloom_defrag_then_scale` to show the defrag callback across a variety of bloom filter configurations in a single run:

1. Non-scaling filters (NONSCALING)
2. Scaling with expansion 1, 2, and 3
3. Varying capacity (30, 50, 100, 150, 200)
4. fp_rate (0.05, 0.01, 0.001)

The test creates 200 keys per configuration, deletes half to create fragmentation, runs defrag, verifies data integrity via digests and BF.MEXISTS, then adds more items to force further scale-out. This targets the potential corruption path where defrag's use of into_boxed_slice() could produce a Vec with an incorrect capacity. Finally, BGSAVE + restart verifies the data survives RDB persistence as well.

Latest push:
~~Added two more production style filters, which have large capacity and lower false positive rate. Will not scale the scaling filter one as it will increase test running time significantly if we have to add and check multiple times~~ Removed due to OOM issues, large capacity compared to smaller capacity should have no effect on defrag logic, and we have a range of capacities checked

### DIFF
--- a/src/wrapper/bloom_callback.rs
+++ b/src/wrapper/bloom_callback.rs
@@ -167,15 +167,15 @@ lazy_static! {
 fn external_vec_defrag(vec: Vec<u8>) -> Vec<u8> {
     let defrag = Defrag::new(core::ptr::null_mut());
     let len = vec.len();
-    let capacity = vec.capacity();
+    // into_boxed_slice will remove any excess capacity which means we need to recreate the Vec with the correct capacity of len
     let vec_ptr = Box::into_raw(vec.into_boxed_slice()) as *mut c_void;
     let defragged_filters_ptr = unsafe { defrag.alloc(vec_ptr) };
     if !defragged_filters_ptr.is_null() {
         metrics::BLOOM_DEFRAG_HITS.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
-        unsafe { Vec::from_raw_parts(defragged_filters_ptr as *mut u8, len, capacity) }
+        unsafe { Vec::from_raw_parts(defragged_filters_ptr as *mut u8, len, len) }
     } else {
         metrics::BLOOM_DEFRAG_MISSES.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
-        unsafe { Vec::from_raw_parts(vec_ptr as *mut u8, len, capacity) }
+        unsafe { Vec::from_raw_parts(vec_ptr as *mut u8, len, len) }
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/wrapper/bloom_callback.rs
```diff
@@ -167,15 +167,14 @@ lazy_static! {
 fn external_vec_defrag(vec: Vec<u8>) -> Vec<u8> {
     let defrag = Defrag::new(core::ptr::null_mut());
     let len = vec.len();
-    let capacity = vec.capacity();
     let vec_ptr = Box::into_raw(vec.into_boxed_slice()) as *mut c_void;
```
KarthikSubbarao, line 171: Let's add a comment for why we use len and not capacity due to `into_boxed_slice` dropping excess capacity

zackcam: I'll add the same comment I had for other place:
` // into_boxed_slice will remove any excess capacity which means we need to recreate the Vec with the correct capacity of len`. This one also has capacity and len the same from my understanding so technically doesn't need changed but for correctness of terms I am

KarthikSubbarao (approved):

KarthikSubbarao (approved):

PR #93, opened 2026-04-23
Moving metric increment to very end of decode

zackcam:
Moving the metric increment to the very end of the decode path. This means that any late error where we fail the decode now will not increment metrics


KarthikSubbarao (approved):
Thanks, Looks good

PR #94, opened 2026-05-19
remove 'validate_scale_to' check in replica

cjx-zar:
`BF.INSERT ... VALIDATESCALETO ...` against an existing bloom uses `replicate_verbatim()`, so the replica re-runs the projection against its own `bloom-memory-usage-limit`. 
If the replica's limit is lower than the primary's, the same command is silently rejected on the replica only — the item is written on the primary but not on the replica → primary/replica divergence.

The error message when `tightening_ratio` is out of range has been fixed, by the way.

### DIFF
--- a/src/bloom/command_handler.rs
+++ b/src/bloom/command_handler.rs
@@ -595,18 +595,25 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
                 utils::NON_SCALING_AND_VALIDATE_SCALE_TO_IS_INVALID,
             ));
         }
-        match utils::BloomObject::calculate_max_scaled_capacity(
-            capacity,
-            fp_rate,
-            scale_to,
-            tightening_ratio,
-            expansion,
-        ) {
-            Ok(_) => (),
-            Err(err) => {
-                return Err(ValkeyError::Str(err.as_str()));
-            }
-        };
+        // Skip the size projection on must-obey clients (replication, AOF replay,
+        // import slot migration). The source already validated against its own
+        // bloom-memory-usage-limit; re-running the check on this node may
+        // produce a different verdict and silently drop a write that succeeded
+        // on the primary, causing primary/replica divergence.
+        if !must_obey_client(ctx) {
+            match utils::BloomObject::calculate_max_scaled_capacity(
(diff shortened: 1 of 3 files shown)


zackcam (approved):
Can we update the pr description to include the change in utils.rs for the error message change as well

src/bloom/command_handler.rs
```diff
@@ -631,18 +595,25 @@ pub fn bloom_filter_insert(ctx: &Context, input_args: &[ValkeyString]) -> Valkey
...
-                return Err(ValkeyError::Str(err.as_str()));
-            }
-        };
+        // Skip the size projection on must-obey clients (replication, AOF replay,
+        // import slot migration). The source already validated against its own
+        // bloom-memory-usage-limit; re-running the check on this node may
+        // produce a different verdict and silently drop a write that succeeded
+        // on the primary, causing primary/replica divergence.
+        if !must_obey_client(ctx) {
+            match utils::BloomObject::calculate_max_scaled_capacity(
```
KarthikSubbarao: Hmm.. The normal BF.INSERT path already has this `must_obey_client` based protection logic (on replicas for e.g.). See [here](https://github.com/valkey-io/valkey-bloom/blob/c6c1558fe045cfd291a032da075fd087a995b9a3/src/bloom/command_handler.rs#L656)
```
 let validate_size_limit = !must_obey_client(ctx);
```

This is the VALIDATE_SCALE_TO path which users call to see if scaling will fail.

So, this PR is for ensuring we skip the X path on replicated commands.


>  and silently drop a write that succeeded

It would not be a silent failure, it would be a crash I believe if a replica rejects a command propagated from primary

src/bloom/command_handler.rs
KarthikSubbarao: Can we move this up if we want to skip it in the `VALIDATE_SCALE_TO` path?

You can move this above block you updated

src/bloom/command_handler.rs
cjx-zar: As far as I know, whether the replica crashes depends on the `propagation-error-behavior` configuration

cjx-zar:
I have updated the pr description and optimized the code :)

KarthikSubbarao (approved):

PR #96, opened 2026-06-01
Add TOPK.RESERVE command and TopK data type

detemmienation:
First slice of TopK support, structured to mirror the existing bloom module. Only TOPK.RESERVE is implemented
## Syntax
`TOPK.RESERVE key topk [SEED seed] [width depth decay] [SEED seed]`
Defaults: width=8, depth=7, decay=0.9. Valid arities: 3, 5, 6, 8.

## Notes
- New dependency: heavykeeper = "0.6" for the underlying sketch.
- Replication uses the canonical 8-arg form so primary and replica build identical sketches.
## Tests
Five integration tests covering the four valid arities, key-already-exists, wrong-arity, parameter validation, and wrong-type.


src/commands/topk.reserve.json
zackcam, line 1: We should probably hold back on adding these .json files until commands are functional as adding here will cause them to appear on the valkey website.

src/topk/command_handler.rs
```diff
@@ -0,0 +1,180 @@
...
+/// always optional; the literal token disambiguates the otherwise-ambiguous
+/// trailing-seed case. When the user does not supply a seed, we generate a
+/// random one on the primary. The resolved seed is always replicated so
+/// replicas build a byte-identical sketch.
+///
+/// Valid argument counts: 3, 5, 6, 8.
+pub fn topk_reserve(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
+    let argc = input_args.len();
+    // Valid arities:
+    //   3 = key topk
```
zackcam, line 63: We should allow seed to be not just at the end. i.e so could be key topk SEED seed width depth decay

src/topk/utils.rs
```diff
@@ -0,0 +1,82 @@
+use heavykeeper::CuckooTopK;
+
+/// KeySpace Notification Events
+pub const RESERVE_EVENT: &str = "topk.reserve";
+
+/// Default sketch parameters used when the user only provides `topk`. These
+/// match RedisBloom's defaults so client behavior is consistent for users
```
zackcam: Don't need to mention matching redis-bloom in comments like this

src/wrapper/topk_callback.rs
```diff
@@ -0,0 +1,81 @@
...
+/// Drop the TopKObject when its key is deleted or replaced.
+pub unsafe extern "C" fn topk_free(value: *mut c_void) {
+    drop(Box::from_raw(value.cast::<TopKObject>()));
+}
+
+/// # Safety
+/// Approximate memory usage for the MEMORY USAGE command. Currently reports
+/// only the size of the wrapper struct; the heap allocations CuckooTopK
+/// performs internally (lobby + heavy slots) are not yet accounted for.
+/// Refine when TOPK.ADD lands and we have a stable view into the sketch.
```
zackcam, line 15: We should leave this out of this pr for now if it doesn't get used or tested really for now

src/lib.rs
```diff
@@ -122,7 +133,8 @@ valkey_module! {
         ["BF.RESERVE", bloom_reserve_command, "write fast deny-oom", 1, 1, 1, "fast write bloom"],
         ["BF.INFO", bloom_info_command, "readonly fast", 1, 1, 1, "fast read bloom"],
         ["BF.INSERT", bloom_insert_command, "write fast deny-oom", 1, 1, 1, "fast write bloom"],
-        ["BF.LOAD", bloom_load_command, "write deny-oom", 1, 1, 1, "write bloom"]
+        ["BF.LOAD", bloom_load_command, "write deny-oom", 1, 1, 1, "write bloom"],
```
zackcam, line 136: Lets add a comment seperating the different probablistic types just for readability

zackcam (commented):
Still working through but quick comments so you can see and update the changes requested. IMO we should archive most of the rdb work for now (Keep it around in a stash don't just delete as it will be needed later). This will allow for better incremental reviews as that code can't be fully tested as is now anyway as we can't check items being added and how they are affected on rdb. With it in now it is creating a lot more changes which reviewing now is not necessary

tests/test_topk_command.py
```diff
@@ -0,0 +1,117 @@
...
+        ) == b'OK'
+
+        # arity 8: full tuning plus explicit seed.
+        assert self.client.execute_command(
+            'TOPK.RESERVE tk5 10 200 5 0.5 SEED 42'
+        ) == b'OK'
+
+        assert self.client.execute_command('DBSIZE') == 5
+
+    def test_topk_reserve_key_already_exists(self):
```
zackcam, line 54: Maybe we can extend this a little bit to test that reserve will work after we delete the key. So can add something like `del dup` then try and create it again

src/topk/command_handler.rs
```diff
@@ -0,0 +1,196 @@
...
+    ctx: &Context,
+    key_name: &ValkeyString,
+    k: u32,
+    width: u32,
+    depth: u32,
+    decay: f64,
+    seed: u64,
+) {
+    let k_str = ValkeyString::create_from_slice(std::ptr::null_mut(), k.to_string().as_bytes());
+    let width_str =
```
zackcam: I think these should be val eg `width_val`. Additionally can we follow the current way we have it in bloom. I.e create a struct ReplicateArgs and pass that in then take out the properties. Link to code: https://github.com/valkey-io/valkey-bloom/blob/unstable/src/bloom/command_handler.rs#L122. Can keep it pretty generic for now as not all the functionality will be in as no items yet

src/topk/command_handler.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    let mut idx = 1;
+    let key_name = &input_args[idx];
+    idx += 1;
+
+    let k = parse_positive_u32(&input_args[idx], utils::BAD_TOPK, utils::TOPK_LARGER_THAN_0)?;
+    idx += 1;
+
+    // Optional leading SEED <n> block, immediately after `topk`.
+    let mut user_seed: Option<u64> = None;
```
zackcam, line 83: A better way could be to do a match statement in a while loop. This way we don't need to have two different checks for SEED. Can look at bf.insert or bf.info for an example

src/topk/command_handler.rs
```diff
@@ -0,0 +1,196 @@
...
+            VALKEY_OK
+        }
+        Err(_) => Err(ValkeyError::Str(utils::ERROR)),
+    }
+}
+
+/// Parse an arg as a u32 that must be strictly positive. Returns the
+/// `bad_format` error for unparseable input (including negatives) and
+/// `not_positive` for explicit zero.
+fn parse_positive_u32(
```
zackcam: We could use an if in a match instead so places we call this would be something like names could be better in my example:
```
                depth = match input_args[idx].to_string_lossy().parse::<u32>() {
                    Ok(num) if (K_MIN..=K_MAX).contains(&num) => num,
                    _ => {
                        return Err(ValkeyError::Str(utils::BAD_K));
                    }
```

This is becasue we most liekly will have different limits for each of the arguments. So having just a generic check might not work in the long term. We can leave the match generic top end for now until we add to the configs. But maybe make a note that we will need to update

src/topk/data_type.rs
```diff
@@ -0,0 +1,40 @@
...
+use valkey_module::native_types::ValkeyType;
+use valkey_module::raw;
+
+/// Module data type encoding version for TopK. Bump this whenever the
+/// on-disk layout changes. RDB save/load callbacks are intentionally left
+/// `None` until TOPK.ADD lands and we can round-trip sketch contents.
+const TOPK_TYPE_ENCODING_VERSION: i32 = 1;
+
+pub static TOPK_TYPE: ValkeyType = ValkeyType::new(
+    "topk-type",
```
zackcam, line 11: Good name!

src/topk/command_handler.rs
```diff
@@ -0,0 +1,196 @@
...
+    args[idx]
+        .to_string_lossy()
+        .parse::<u64>()
+        .map_err(|_| ValkeyError::Str(utils::INVALID_SEED))
+}
+
+/// Generate a u64 seed using stdlib only. Mixes a high-resolution timestamp
+/// with the address of a stack local through DefaultHasher. Cheap and
+/// non-cryptographic; sufficient for sketch hash diversification.
+fn random_seed() -> u64 {
```
zackcam, line 175: I took a quick look, i think the builder for cuckoo can choose a random seed without us having to create one. Might be better to use that

src/lib.rs
```diff
@@ -122,7 +133,9 @@ valkey_module! {
...
+        // Commands for topk
```
zackcam: Nit `TOPK Commands ` to me seems nicer, plus follows what the proposed cuckoo filter pr is doing

tests/test_topk_command.py
```diff
@@ -0,0 +1,134 @@
+from valkey_bloom_test_case import ValkeyBloomTestCaseBase
+from valkeytestframework.conftest import resource_port_tracker  # noqa: F401
```
zackcam, line 2: What is the `# noqa: F401` just curious

zackcam (commented):
Really nice start on this! Left some comments, feel free to push back if you implemented certain things I asked for a specific reason

src/topk/command_handler.rs
zackcam: Nevermind the seed is not exposed doing it this way is best! Nice!

tests/test_topk_command.py
detemmienation: it suppresses the "imported but unused" warning 
resource_port_tracker doesn't need to be referenced in the code but the import needs to be there for pytest

src/topk/command_handler.rs
detemmienation: Thank you ^.^

tests/test_topk_command.py
detemmienation: Added!

src/topk/command_handler.rs
detemmienation: Done. Renamed the locals to _val and introduced the struct ReplicateArgs 🫡

src/topk/command_handler.rs
detemmienation: Done. Collapsed the 2 checks into one while loop.

src/topk/command_handler.rs
detemmienation: Each arg now has its own match and bounds; max is a placeholder until we add topk configs.

zackcam (approved):
Looks good to me! Will run workflows then merge in

PR #97, opened 2026-06-03
Support reusing an externally built valkey-server binary for test

enjoy-binbin:
Add VALKEY_SERVER_PATH to skip building Valkey from source and copy
the provided binary into the integration test directory instead.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -31,7 +31,7 @@ if [ -z "$SERVER_VERSION" ]; then
     export SERVER_VERSION="unstable"
 fi
 
-if [ "$SERVER_VERSION" != "unstable" ] && [ "$SERVER_VERSION" != "8.0" ] && [ "$SERVER_VERSION" != "8.1" ] && [ "$SERVER_VERSION" != "9.0" ]; then
+if [ "$SERVER_VERSION" != "unstable" ] && [ "$SERVER_VERSION" != "8.0" ] && [ "$SERVER_VERSION" != "8.1" ] && [ "$SERVER_VERSION" != "9.0" ] && [ "$SERVER_VERSION" != "9.1" ]; then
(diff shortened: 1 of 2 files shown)


enjoy-binbin:
@KarthikSubbarao Would you be able to take a look?

build.sh
```diff
@@ -47,6 +47,20 @@ fi
 REPO_URL="https://github.com/valkey-io/valkey.git"
 BINARY_PATH="tests/build/binaries/$SERVER_VERSION/valkey-server"
 CACHED_VALKEY_PATH="tests/build/valkey"
+
+# Optional: path to an externally built valkey-server binary. When set,
+# we skip building Valkey from source and copy the provided binary into
+# the integration test directory instead.
+if [ -n "$VALKEY_SERVER_PATH" ]; then
+    if [ ! -f "$VALKEY_SERVER_PATH" ]; then
```
zackcam: Do we want to check if this is executable here not just a regular file? I know we check just below but that check would then fall back to building valkey from source which might get missed by whoever was using the flag pretty easily. So would be instead:
```
    if [ ! -x "$VALKEY_SERVER_PATH" ]; then
```

enjoy-binbin: Yes, make sense, i actually did check at the first time but somehow i remove it. Added, thanks for the review.

KarthikSubbarao (commented):
@enjoy-binbin  - can we fix the DCO?

enjoy-binbin:
@KarthikSubbarao I have fixed the DCO. Sorry, i didn't realize there was a commit that hadn't been squashed in.

build.sh
zackcam: Nice thanks!

zackcam (approved):

PR #99, opened 2026-06-03
Add TOPK.ADD command

detemmienation:
## Summary
Implements `TOPK.ADD key item [item ...]`. Adds items to a sketch created by `TOPK.RESERVE` and returns an array with one entry per item: nil if nothing was displaced from the top-k, otherwise the bulk string of the evicted item.

## Changes
- `TopKObject::add` wraps heavykeeper's `add_with_evicted` to surface the displaced item.
- `topk_add` handler: arity check, missing-key error (no auto-create), wrong-type handling, per-item reply assembly.
- Registered `TOPK.ADD` in `lib.rs`.
- Reorganized `test_topk_command.py` into table-driven arity/error/behavior tests covering RESERVE and ADD.

## Testing
- `pytest tests/test_topk_command.py`.
- `cargo build --release` clean.

## Note
Depends on a new `add_with_evicted` API not yet in published heavykeeper;`Cargo.toml` pins a fork branch for now and needs to move to an upstreamed version before merge.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,11 +16,12 @@ valkey-module = { version = "0.1.5", features = ["min-valkey-compatibility-versi
 valkey-module-macros = "0"
 linkme = "0"
 bloomfilter = { version = "3.0.1", features = ["serde"] }
-heavykeeper = "0.6"
+heavykeeper = { git = "https://github.com/detemmienation/heavykeeper-rs.git", branch = "evicted" }
 lazy_static = "1.4.0"
 libc = "0.2"
 serde = { version = "1.0", features = ["derive"] }
 bincode = "1.3"
+
 
 [dev-dependencies]
 rand = "0.8"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,12 +16,12 @@ valkey-module = { version = "0.1.5", features = ["min-valkey-compatibility-versi
 valkey-module-macros = "0"
 linkme = "0"
 bloomfilter = { version = "3.0.1", features = ["serde"] }
+# TODO: This is a temporary solution until we add this functionality to the original heavykeeper crate.
 heavykeeper = { git = "https://github.com/detemmienation/heavykeeper-rs.git", branch = "evicted" }
 lazy_static = "1.4.0"
 libc = "0.2"
 serde = { version = "1.0", features = ["derive"] }
 bincode = "1.3"
-
 
 [dev-dependencies]
 rand = "0.8"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,7 +95,7 @@ fn bloom_load_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
 }
 
 /// Command handler for:
-/// TOPK.RESERVE key topk [SEED seed] [width depth decay] [SEED seed]
+/// TOPK.RESERVE key topk [width depth decay] [SEED seed]
 fn topk_reserve_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     topk_command_handler::topk_reserve(ctx, &args)
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,9 +95,14 @@ fn bloom_load_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
 }
 
 /// Command handler for:
-/// TOPK.RESERVE key topk [width depth decay] [SEED seed]
+/// TOPK.RESERVE key topk [SEED seed] [width depth decay] [SEED seed]
 fn topk_reserve_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     topk_command_handler::topk_reserve(ctx, &args)
+}
+
+/// Command handler for TOPK.ADD <key> <item> [<item> ...]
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -16,12 +16,13 @@ valkey-module = { version = "0.1.5", features = ["min-valkey-compatibility-versi
 valkey-module-macros = "0"
 linkme = "0"
 bloomfilter = { version = "3.0.1", features = ["serde"] }
-heavykeeper = "0.6"
+heavykeeper = { git = "https://github.com/detemmienation/heavykeeper-rs.git", branch = "evicted" }
```
zackcam, line 20: Can we add a comment saying this is a temporary solution until we add this functionality to the heavykeeper original crate

Cargo.toml
```diff
@@ -16,12 +16,13 @@ valkey-module = { version = "0.1.5", features = ["min-valkey-compatibility-versi
...
 lazy_static = "1.4.0"
 libc = "0.2"
 serde = { version = "1.0", features = ["derive"] }
 bincode = "1.3"
 
+
```
zackcam: We should remove this extra line 😄 (Sorry know its my fault from showing you some things)

src/topk/command_handler.rs
```diff
@@ -146,13 +153,51 @@ pub fn topk_reserve(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult
...
+///
+/// Syntax:
+///     TOPK.ADD key item [item ...]
+///
+/// Returns an array, one entry per input item, in order:
+///   - Null if no heavy-slot resident was displaced by the insertion.
+///   - The bulk-string of the displaced item otherwise.
+///
+/// The key must already hold a TOPK reserved via TOPK.RESERVE; missing keys
+/// are an error rather than auto-created (matching upstream RedisBloom).
```
zackcam: Can remove the matching part of the comment

src/lib.rs
```diff
@@ -95,11 +95,16 @@ fn bloom_load_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
 }
 
 /// Command handler for:
-/// TOPK.RESERVE key topk [width depth decay] [SEED seed]
+/// TOPK.RESERVE key topk [SEED seed] [width depth decay] [SEED seed]
```
zackcam: I don't think we need to show the seed in front as well. Optional args are typically thought of being able to be moved around as is

tests/test_topk_command.py
```diff
@@ -1,104 +1,27 @@
...
-        # arity 6: full sketch tuning, server-generated seed.
-        assert self.client.execute_command(
-            'TOPK.RESERVE tk4 5 50 4 0.9'
-        ) == b'OK'
 
-        # arity 8: full tuning plus explicit seed.
-        assert self.client.execute_command(
-            'TOPK.RESERVE tk5 10 200 5 0.5 SEED 42'
-        ) == b'OK'
+    def verify_command_arity(self, command, expected_arity):
```
zackcam: We shoudl move this and the one in test_bloom_command to valkey_bloom_test_case so we can reduce repeated code

tests/test_topk_command.py
```diff
@@ -116,23 +39,71 @@ def test_topk_reserve_bad_params(self):
...
-        # Reserving on a key that already holds a non-TOPK value should fail.
-        assert self.client.execute_command('SET strkey hello') == b'OK'
-        try:
-            self.client.execute_command('TOPK.RESERVE strkey 5')
-            assert False, 'TOPK.RESERVE on a string key should fail'
-        except ResponseError as e:
-            # WRONGTYPE is the standard Valkey error for this case.
-            assert 'WRONGTYPE' in str(e), f'unexpected error: {e}'
+        # ---- TOPK.ADD: no eviction ----
+        # With k=3 and only three distinct items, nothing is ever displaced.
```
zackcam: This file we more want to just check that command parsing is working as intended so having the commands in one test case list might be better. This would then not bother checking with the eviction logic here. I would move checking the eviction logic to a new test file. 
Also with the eviction as it is not deterministic random seed and probability we might need to just check the length of the return not the actual values

src/topk/utils.rs
```diff
@@ -79,4 +81,13 @@ impl TopKObject {
     pub fn sketch_mut(&mut self) -> &mut CuckooTopK<Vec<u8>> {
         &mut self.sketch
     }
+
+    /// Add `increment` occurrences of `item` to the sketch and return the
+    /// heavy-slot resident displaced by this insertion (if any). At most one
+    /// item can be evicted per call.
+    pub fn add(&mut self, item: &[u8], increment: u64) -> Option<Vec<u8>> {
```
zackcam, line 88: The increment will be so we can utilize this in the topk.incrby command right?

detemmienation: Yes, increment is actually native to the heavykeeper crate's add API and we can use it for TOPK.INCRBY in the future :D

tests/test_topk_command.py
detemmienation: Done as suggested 🫡

tests/test_topk_command.py
detemmienation: Done as suggested 👯

zackcam (approved):

PR #100, opened 2026-06-04
Add TOPK.INFO, TOPK.INCRBY, and TOPK.LIST commands

detemmienation:
## Summary

Adds three read/write TOPK commands:

- **TOPK.INFO key** — returns the sketch parameters: `k`, `width`, `depth`, and `decay`. Read-only.
- **TOPK.INCRBY key item increment [item increment ...]** — like TOPK.ADD but with an explicit per-item increment. Returns one entry per pair: the evicted item (bulk string) or nil.
- **TOPK.LIST key [WITHCOUNT]** — returns the tracked items ordered by descending count (Top `k`). With WITHCOUNT, each item is followed by its estimated count. Read-only.

## Dependency change

`Cargo.toml` now points at the published `heavykeeper = "0.6.6"`  
`add_with_evicted` returns `Option<T>` in this version; the sketch adapter was adjusted accordingly.

## Testing

Full Python integration suite passes, including new TOPK.INFO/INCRBY/LIST coverage in `tests/test_topk_command.py`: arity, error paths (missing key, wrong type, bad arguments, wrong arity), and behavior (reply shape, ranked order, WITHCOUNT pairs, case-insensitive token, at-most-k guarantee).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,6 +107,12 @@ fn topk_add_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
 }
 
 /// Command handler for:
+///  TOPK.INCRBY <key> <item> <increment> [<item> <increment> ...]
+fn topk_incrby_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
+    topk_command_handler::topk_incrby(ctx, &args)
+}
+
+/// Command handler for:
 /// TOPK.INFO key
 fn topk_info_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     topk_command_handler::topk_info(ctx, &args)
@@ -149,6 +155,7 @@ valkey_module! {
         // TOPK Commands
         ["TOPK.RESERVE", topk_reserve_command, "write fast deny-oom", 1, 1, 1, "fast write topk"],
         ["TOPK.ADD", topk_add_command, "write fast deny-oom", 1, 1, 1, "fast write topk"],
+        ["TOPK.INCRBY", topk_incrby_command, "write fast deny-oom", 1, 1, 1, "fast write topk"],
         ["TOPK.INFO", topk_info_command, "readonly fast", 1, 1, 1, "fast read topk"],
     ],
     configurations: [
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -100,9 +100,16 @@ fn topk_reserve_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult 
     topk_command_handler::topk_reserve(ctx, &args)
 }
 
-/// Command handler for TOPK.ADD <key> <item> [<item> ...]
+/// Command handler for:
+///  TOPK.ADD <key> <item> [<item> ...]
 fn topk_add_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     topk_command_handler::topk_add(ctx, &args)
+}
+
+/// Command handler for:
+/// TOPK.INFO key
+fn topk_info_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
+    topk_command_handler::topk_info(ctx, &args)
 }
 
 ///
@@ -142,6 +149,7 @@ valkey_module! {
         // TOPK Commands
         ["TOPK.RESERVE", topk_reserve_command, "write fast deny-oom", 1, 1, 1, "fast write topk"],
         ["TOPK.ADD", topk_add_command, "write fast deny-oom", 1, 1, 1, "fast write topk"],
+        ["TOPK.INFO", topk_info_command, "readonly fast", 1, 1, 1, "fast read topk"],
     ],
     configurations: [
         i64: [
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,38 +88,32 @@ fn bloom_insert_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult 
     command_handler::bloom_filter_insert(ctx, &args)
 }
 
-/// Command handler for:
-/// BF.LOAD <key> data
+/// Command handler for BF.LOAD <key> data
 fn bloom_load_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     command_handler::bloom_filter_load(ctx, &args)
 }
 
-/// Command handler for:
-/// TOPK.RESERVE key topk [width depth decay] [SEED seed]
+/// Command handler for TOPK.RESERVE key topk [width depth decay] [SEED seed]
 fn topk_reserve_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     topk_command_handler::topk_reserve(ctx, &args)
 }
 
-/// Command handler for:
-///  TOPK.ADD <key> <item> [<item> ...]
+/// Command handler for TOPK.ADD <key> <item> [<item> ...]
 fn topk_add_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     topk_command_handler::topk_add(ctx, &args)
 }
 
-/// Command handler for:
-///  TOPK.INCRBY <key> <item> <increment> [<item> <increment> ...]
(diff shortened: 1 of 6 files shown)


src/topk/command_handler.rs
```diff
@@ -195,6 +194,136 @@ pub fn topk_add(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
...
+///     TOPK.INCRBY key item increment [item increment ...]
+///
+/// Like TOPK.ADD, but each item carries an explicit increment.
+/// Returns an array, one entry per item/increment pair, in order:
+///   - Null if no heavy-slot resident was displaced by the insertion.
+///   - The bulk-string of the displaced item otherwise.
+pub fn topk_incrby(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
+    let argc = input_args.len();
+    // key + at least one item/increment pair, and the pairs must be complete.
+    if argc < 4 || !(argc - 2).is_multiple_of(2) {
```
zackcam: Nit: We don't really need to do a minus 2 here do we?

src/topk/command_handler.rs
```diff
@@ -195,6 +194,136 @@ pub fn topk_add(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
...
+        let item = &pair[0];
+        let increment = match pair[1].to_string_lossy().parse::<u64>() {
+            Ok(0) => return Err(ValkeyError::Str(utils::BAD_INCREMENT)),
+            Ok(num) => num,
+            Err(_) => return Err(ValkeyError::Str(utils::BAD_INCREMENT)),
+        };
+        parsed.push((item, increment));
+    }
+
+    let mut result: Vec<ValkeyValue> = Vec::with_capacity(parsed.len());
```
zackcam: This is pretty similar to what we have in add. I wonder if we cna extract this from add and incrby and make it a generic where we pass in the count?

src/lib.rs
```diff
@@ -100,11 +100,30 @@ fn topk_reserve_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult
     topk_command_handler::topk_reserve(ctx, &args)
 }
 
-/// Command handler for TOPK.ADD <key> <item> [<item> ...]
+/// Command handler for:
```
zackcam: Nit: Can probably be same line, look above we only really do separate if we have lots of arguments

tests/test_topk_command.py
```diff
@@ -85,3 +117,52 @@ def test_topk_command_behavior(self):
...
+        assert info[b'decay'] == b'0.5'
+
+        # TOPK.LIST returns tracked items by descending count, at most k.
+        assert self.client.execute_command('TOPK.RESERVE tk_list 3 50 4 0.9 SEED 42') == b'OK'
+        self.client.execute_command('TOPK.INCRBY tk_list apple 10 banana 5 cherry 2')
+
+        listed = self.client.execute_command('TOPK.LIST tk_list')
+        assert listed == [b'apple', b'banana', b'cherry']
+
+        # WITHCOUNT interleaves each item with its estimated count.
```
zackcam, line 163: We should check if this really is estimated count or not. Due to cuckoo type I wonder if this is the real count. As evicting from heavy would require an accurate count I would think otherwise comparing the heavy slot to the lobby slot would not always be accurate.

zackcam (commented):
Majority of the code is looking good. As we have a more flushed out utils.rs now can we implement some unit tests for this?

Also for a couple test cases can we use some number/non just straight letter  keynames and items  just for completeness

src/lib.rs
detemmienation: Gotcha 🫡

src/topk/command_handler.rs
detemmienation: YEAH 😂

src/topk/command_handler.rs
detemmienation: Done as suggested :D

tests/test_topk_command.py
detemmienation: It should be estimated account based on our discussion ^^
The initial count carried over from the lobby can already be an underestimate due to decay, and that's never corrected after promotion

src/lib.rs
```diff
@@ -88,38 +88,32 @@ fn bloom_insert_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult
...
-/// Command handler for:
-///  TOPK.INCRBY <key> <item> <increment> [<item> <increment> ...]
+/// Command handler for TOPK.INCRBY <key> <item> <increment> [<item> <increment> ...]
 fn topk_incrby_command(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     topk_command_handler::topk_incrby(ctx, &args)
 }
 
-/// Command handler for:
-/// TOPK.INFO key
+/// Command handler for TOPK.INFO key
```
zackcam, line 111: Nit: missing the colons for these

src/topk/utils.rs
```diff
@@ -99,3 +99,139 @@ impl TopKObject {
...
+        // The constructor should round-trip every parameter through the getters.
+        let topk = TopKObject::new_reserved(5, 50, 4, 0.9, 42);
+        assert_eq!(topk.k(), 5);
+        assert_eq!(topk.width(), 50);
+        assert_eq!(topk.depth(), 4);
+        assert_eq!(topk.decay(), 0.9);
+        assert_eq!(topk.seed(), 42);
+    }
+
+    #[rstest(seed, case::seed_a(1), case::seed_b(42), case::seed_c(123456789))]
```
zackcam, line 119: Nice tests!
One note we can probably change this later to have two cases of fixed and random seed. Same with other tests. Can update this when we have configs made though!

src/topk/utils.rs
```diff
@@ -99,3 +99,139 @@ impl TopKObject {
             .collect()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rstest::rstest;
+
+    fn new_topk(k: u32, seed: u64) -> TopKObject {
```
zackcam: I think I personally like having this creation even if we are using generic numbers, this way we can see the params in each test easier

src/topk/command_handler.rs
```diff
@@ -159,6 +159,21 @@ pub fn topk_reserve(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult
         Err(_) => Err(ValkeyError::Str(utils::ERROR)),
     }
 }
+
+fn apply_increments<'a>(
```
zackcam: Maybe a rename to add_with_increments?

src/topk/utils.rs
```diff
@@ -99,3 +99,139 @@ impl TopKObject {
...
+            (b"c".as_slice(), 8),
+            (b"d".as_slice(), 7),
+            (b"e".as_slice(), 6),
+        ] {
+            topk.add(item, incr);
+        }
+        assert!(topk.list().len() <= 3);
+    }
+
+    #[rstest(
```
zackcam, line 200: Can we have some of these in the integration tests as well?

src/topk/utils.rs
detemmienation: Done.

src/topk/command_handler.rs
detemmienation: Done.

src/topk/utils.rs
detemmienation: Done as suggested!

tests/test_topk_basic.py
```diff
@@ -34,3 +34,23 @@ def test_topk_add_eviction_count(self):
         evictions = [r for r in result if r is not None]
         assert len(evictions) == 1
 
+    def test_topk_incrby_accumulates_count(self):
```
zackcam, line 37: Sorry don't think i was clear in the other comment. I more meant expanding [these cases](https://github.com/valkey-io/valkey-bloom/blob/topk/tests/test_topk_command.py#L83) in the command file to have more types of keys

But these tests are also nice to have here so don't remove them!

zackcam (approved):

zackcam:
Thank you! Looks good to me!

PR #102, opened 2026-06-05
Add TOPK.COUNT, TOPK.QUERY commands

detemmienation:
## Summary

Adds two read-only TOPK commands:

- `TOPK.COUNT key item [item ...]` — returns the estimated count for each item in order. Untracked items return 0. Counts are estimates that may be lower than the true count but never higher.
- `TOPK.QUERY key item [item ...]` — returns whether each item is currently in the Top-K list, as 1 (in) or 0 (not), in order. 

## Dependency change

`TOPK.QUERY` relies on a new `query_topk_items` API in the heavykeeper crate. This PR temporarily points the dependency at a fork branch. This needs to move to a published release before merge.

## Tests

- Rust unit tests for `count()` and `query()` (seed-parametrized), including the eviction case where a displaced item reports not-in-top-k.
- Python integration tests: arity, error cases (missing key, wrong type, wrong arity), per-item results in order, mixed tracked/untracked, agreement with TOPK.LIST, and the count `<=` true-count invariant.

### DIFF
--- a/src/topk/utils.rs
+++ b/src/topk/utils.rs
@@ -90,6 +90,12 @@ impl TopKObject {
         self.sketch.add_with_evicted(item, increment)
     }
 
+    /// Return the estimated count for `item`, or 0 if it has no residual
+    /// presence in the priority queue or sketch cells.
+    pub fn count(&self, item: &[u8]) -> u64 {
+        self.sketch.count(item)
+    }
+
     /// Return the Top-K items
     pub fn list(&self) -> Vec<(Vec<u8>, u64)> {
         self.sketch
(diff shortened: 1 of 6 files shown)


tests/test_topk_command.py
```diff
@@ -174,3 +190,43 @@ def info_dict(key):
...
+        self.client.execute_command('TOPK.INCRBY tk_count apple 10 banana 5 cherry 2')
+        assert self.client.execute_command('TOPK.COUNT tk_count apple') == [10]
+        assert self.client.execute_command('TOPK.COUNT tk_count apple banana cherry') == [10, 5, 2]
+        assert self.client.execute_command('TOPK.COUNT tk_count missing') == [0]
+        assert self.client.execute_command('TOPK.COUNT tk_count apple missing banana') == [10, 0, 5]
+        assert self.client.execute_command('TOPK.COUNT tk_count apple apple') == [10, 10]
+
+        # Estimates never exceed true counts. Use a small, narrow
+        # sketch so collisions are likely, then check the invariant holds.
+        assert self.client.execute_command('TOPK.RESERVE tk_inv 5 4 2 0.9 SEED 42') == b'OK'
```
zackcam: Nit: This sort of testing should be in test top k basic

src/topk/utils.rs
```diff
@@ -90,6 +90,17 @@ impl TopKObject {
         self.sketch.add_with_evicted(item, increment)
     }
 
+    /// Return the estimated count for `item`, or 0 if it has no residual
+    /// presence in the priority queue or sketch cells.
```
zackcam, line 94: What are sketch cells?

src/topk/utils.rs
```diff
@@ -213,6 +224,97 @@ mod tests {
         assert_eq!(counts.get(item_b), Some(&5));
     }
 
+    #[rstest(seed, case::seed_a(1), case::seed_b(42), case::seed_c(123456789))]
+    fn test_count_untracked_item_is_zero(seed: u64) {
+        let topk = TopKObject::new_reserved(3, 256, 4, DEFAULT_DECAY, seed);
+        assert_eq!(topk.count(b"apple"), 0);
+    }
+
+    #[rstest(seed, case::seed_a(1), case::seed_b(42), case::seed_c(123456789))]
```
zackcam, line 227: Can probably combine this and test below

src/topk/utils.rs
```diff
@@ -213,6 +224,97 @@ mod tests {
...
+        let mut topk = TopKObject::new_reserved(5, 256, 4, DEFAULT_DECAY, 42);
+        topk.add(b"apple", 10);
+        topk.add(b"banana", 5);
+        let counts: std::collections::HashMap<Vec<u8>, u64> = topk.list().into_iter().collect();
+        assert_eq!(topk.count(b"apple"), counts[b"apple".as_slice()]);
+        assert_eq!(topk.count(b"banana"), counts[b"banana".as_slice()]);
+    }
+
+    #[rstest(seed, case::seed_a(1), case::seed_b(42), case::seed_c(123456789))]
+    fn test_query_untracked_item_is_false(seed: u64) {
```
zackcam: We test this below as well, don't think we need this unit test

zackcam (commented):
We mostly want to keep the command test file for error messages, and checking that the command returns a valid result. For more specific tests on the commands we can use the other test file we have for now


Also, do you know how many cells are in heavy? Is there a default or is it based on k?

tests/test_topk_command.py
detemmienation: Done as suggested!

src/topk/utils.rs
detemmienation: heavy cell/lobby cell https://github.com/pmcgleenon/heavykeeper-rs/blob/1e1c610e9bf9afd75e1a812e75535b5944e4c507/src/cuckoo.rs#L94
Yeah the wording was a bit unclear, updated.

src/topk/utils.rs
detemmienation: Combined :D

src/topk/utils.rs
detemmienation: Deleted

detemmienation:
> We mostly want to keep the command test file for error messages, and checking that the command returns a valid result. For more specific tests on the commands we can use the other test file we have for now
> 
> Also, do you know how many cells are in heavy? Is there a default or is it based on k?

I think the number of heavy cells is based on width and depth:`width * depth` https://github.com/pmcgleenon/heavykeeper-rs/blob/1e1c610e9bf9afd75e1a812e75535b5944e4c507/src/cuckoo.rs#L520
It organized as `width` buckets, each with `depth` heavy slots.

zackcam (approved):

PR #103, opened 2026-06-06
Fix lazy-delete timing race in test_bloom_metrics

detemmienation:
CI run fails intermittently at the metric assertion after `UNLINK` key: [CI run](https://github.com/valkey-io/valkey-bloom/actions/runs/27044533835/job/79827647149?pr=102#step:12:82)

- Cause: `UNLINK` frees the object asynchronously, but metrics only decrement in` Drop `(when the free actually runs). The test reads `INFO bf` immediately, so it sometimes sees stale values.

- Fix: Poll until the metric settles before asserting:

```
self.client.execute_command('UNLINK key')
wait_for_equal(lambda: self.parse_valkey_info("bf").get('bf_bloom_num_objects'), '2')
```

### DIFF
--- a/tests/test_bloom_metrics.py
+++ b/tests/test_bloom_metrics.py
@@ -55,7 +55,6 @@ class TestBloomMetrics(ValkeyBloomTestCaseBase):
 
         # Delete a default key and make sure the metrics are still correct
         self.client.execute_command('UNLINK key')
-        wait_for_equal(lambda: self.parse_valkey_info("bf").get('bf_bloom_num_objects'), '2')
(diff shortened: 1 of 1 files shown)


zackcam:
The ci run failed on the command:
```
self.verify_bloom_metrics(self.client.execute_command("INFO bf"), 0, 0, 0, 0, 0)
``` 
The change made in this PR is before the failure by a few checks, I think if the unlink was the problem we would have failed earlier on. The issue here seems to be with the flushdb

detemmienation:
> The ci run failed on the command:
> 
> ```
> self.verify_bloom_metrics(self.client.execute_command("INFO bf"), 0, 0, 0, 0, 0)
> ```
> 
> The change made in this PR is before the failure by a few checks, I think if the unlink was the problem we would have failed earlier on. The issue here seems to be with the flushdb

 Thanks for the feedback! I've updated the fix to add `wait_for_equal` on `bf_bloom_num_objects` after FLUSHDB instead.

zackcam (approved):

PR #104, opened 2026-06-07
`BF.INFO`: documentation and reabability nits

somechris:
* `bf.info`: Harmonize sorting of info types

  The order of the types between the web documentation `Usage` section,
  `INFO FIELDS` section, and the code did not agree.

  We sort all three alike to make things more readable.

* `commands/bf.info`: Add missing `ITEMS` entry

  `ITEMS` was missing from the web documentation's `Usage` section. So
  we add it.


zackcam:
Hi, I was wondering if we could have the PR description updated?

somechris:
> Hi, I was wondering if we could have the PR description updated?

I've added more details to the title and the description.

Is this what you had in mind?

zackcam (approved):

zackcam:
Awesome exactly that. Thanks!

PR #105, opened 2026-06-09
Add keyspace notification and ACL category test coverage for TopK

detemmienation:
## Summary

Adds integration test coverage for TopK keyspace notifications and ACL categories.

### `tests/test_topk_keyspace.py` 
Modeled on `test_bloom_keyspace.py`. Verifies:
- `TOPK.RESERVE` emits `topk.reserve`.
- `TOPK.ADD` and `TOPK.INCRBY` both emit `topk.add` (one event per command, regardless of item count).
- Read-only commands (represented by `TOPK.QUERY`) emit no notification.
- `DEL` of a TopK key emits the native `del` keyspace event.

Unlike `BF.ADD`, TopK commands do not auto-create the key, so each mutative command is preceded by its own `TOPK.RESERVE`. The psubscribe confirmation is skipped by filtering on message `type == 'pmessage'` rather than matching on command name.

### `tests/test_topk_acl_category.py` 
Modeled on `test_bloom_acl_category.py`. Verifies:
- Users without `@topk` access (via `-@topk` / `-@all`) are denied every TopK command.
- Users granted access (`+@all`, `+@topk`, `+@write +@read`, `+@write +@topk`) can run them.
- Each command exposes the expected flags and ACL categories via `COMMAND INFO` (`TOPK.LIST` is `readonly` without `fast`; the rest carry`fast`, with write commands also `denyoom`).

### DIFF
--- a/tests/test_topk_keyspace.py
+++ b/tests/test_topk_keyspace.py
@@ -1,0 +1,72 @@
+import time
+from valkey_bloom_test_case import ValkeyBloomTestCaseBase
+from valkeytestframework.conftest import resource_port_tracker
+
+class TestTopkKeyEventNotifications(ValkeyBloomTestCaseBase):
+    RESERVE_KEYSPACE_MESSAGE = {'type': 'pmessage', 'pattern': b'__key*__:*', 'channel': b'__keyspace@0__:intermediate_val', 'data': b'topk.reserve'}
+    RESERVE_KEYEVENT_MESSAGE = {'type': 'pmessage', 'pattern': b'__key*__:*', 'channel': b'__keyevent@0__:topk.reserve', 'data': b'intermediate_val'}
+    ADD_KEYSPACE_MESSAGE = {'type': 'pmessage', 'pattern': b'__key*__:*', 'channel': b'__keyspace@0__:intermediate_val', 'data': b'topk.add'}
+    ADD_KEYEVENT_MESSAGE = {'type': 'pmessage', 'pattern': b'__key*__:*', 'channel': b'__keyevent@0__:topk.add', 'data': b'intermediate_val'}
+
+    def create_expected_message_list(self, reserve_expected, add_expected, key_name):
+        expected_messages = []
+        self.RESERVE_KEYSPACE_MESSAGE['channel'] = f"__keyspace@0__:{key_name}".encode('utf-8')
+        self.RESERVE_KEYEVENT_MESSAGE['data'] = f"{key_name}".encode('utf-8')
+        self.ADD_KEYSPACE_MESSAGE['channel'] = f"__keyspace@0__:{key_name}".encode('utf-8')
+        self.ADD_KEYEVENT_MESSAGE['data'] = f"{key_name}".encode('utf-8')
+        if reserve_expected:
+            expected_messages.append(self.RESERVE_KEYEVENT_MESSAGE)
+            expected_messages.append(self.RESERVE_KEYSPACE_MESSAGE)
+        if add_expected:
+            expected_messages.append(self.ADD_KEYSPACE_MESSAGE)
+            expected_messages.append(self.ADD_KEYEVENT_MESSAGE)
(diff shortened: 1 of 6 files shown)


tests/test_topk_keyspace.py
```diff
@@ -0,0 +1,72 @@
...
+            expected_messages = self.create_expected_message_list(reserve_expected, add_expected, command.split()[1]) if reserve_expected or add_expected else []
+            result_messages = self.get_subscribe_client_messages(self.keyspace_client, command, expected_message_count)
+            self.check_response(result_messages, expected_messages)
+
+        # test del
+        del_messages = self.get_subscribe_client_messages(self.keyspace_client, 'DEL add_test', 2)
+        assert {'type': 'pmessage', 'pattern': b'__key*__:*', 'channel': b'__keyspace@0__:add_test', 'data': b'del'} in del_messages
+        assert {'type': 'pmessage', 'pattern': b'__key*__:*', 'channel': b'__keyevent@0__:del', 'data': b'add_test'} in del_messages
+
+    def create_subscribe_clients(self):
```
zackcam: I wonder if we can make some of these methods generic and just call them for both bloom and topk to save on code lines

zackcam (commented):
Mostly looks good to me. I would say lets look at putting some common functions that data types would use in a generic place

tests/test_topk_keyspace.py
detemmienation: 🫡

tests/valkey_bloom_test_case.py
```diff
@@ -308,3 +309,94 @@ def restart_external_server(self, server, remove_rdb=True, remove_nodes_conf=Tru
...
+        client = self.server.get_new_client()
+        list_of_commands = client.execute_command(f"COMMAND LIST FILTERBY ACLCAT {category}")
+        # Users without access to the category.
+        client.execute_command(f"ACL SETUSER non{category}user1 on >{category}_pass -@{category}")
+        client.execute_command(f"ACL SETUSER non{category}user2 on >{category}_pass -@all")
+        # Users with access via different grant combinations.
+        client.execute_command(f"ACL SETUSER {category}user1 on >{category}_pass ~* &* +@all ")
+        client.execute_command(f"ACL SETUSER {category}user2 on >{category}_pass ~* &* -@all +@{category} ")
+        client.execute_command(f"ACL SETUSER {category}user3 on >{category}_pass ~* &* -@all +@write +@read ")
+        client.execute_command(f"ACL SETUSER {category}user4 on >{category}_pass ~* &* -@all +@write +@{category}")
```
zackcam: I think I would still keep this part in the respective branches might make it hard to read the tests. Also makes it less restrictive incase other module types want to try differetn acl combinations. So could pass in user and pass for later on methods

tests/test_topk_keyspace.py
```diff
@@ -1,72 +1,31 @@
...
-            ('TOPK.RESERVE add_test 3 50 4 0.9', True, False, 2),
-            ('TOPK.ADD add_test apple banana', False, True, 2),
-            ('TOPK.RESERVE incr_test 3 50 4 0.9', True, False, 2),
-            ('TOPK.INCRBY incr_test apple 5 banana 3', False, True, 2),
-            ('TOPK.QUERY add_test apple', False, False, 0)
+            ('TOPK.RESERVE add_test 3 50 4 0.9', ['topk.reserve'], 2),
+            ('TOPK.ADD add_test apple banana', ['topk.add'], 2),
+            ('TOPK.RESERVE incr_test 3 50 4 0.9', ['topk.reserve'], 2),
+            ('TOPK.INCRBY incr_test apple 5 banana 3', ['topk.add'], 2),
+            ('TOPK.QUERY add_test apple', [], 0),
```
zackcam, line 18: Maybe we can check all the read only commands should be not much additional time for robustness

zackcam (approved):

Issue #84, opened 2026-03-30
[Feature Request] Count-Min Sketch Commands


zmccoy:
Hello, sifting through the original Valkey-Bloom Valkey issue it seems like a consensus was had for adding other probabilistic datastructure commands to valkey-bloom.  

Would the 6 CSM commands in the Redis protocol today be applicable to be included here?  We're users of Redis' count-min sketch commands today and would like to migrate to Valkey, but this is one of our hold ups.  I'd be happy to take guidance on how adding another structure into this crate should look, and happy to work through getting it added.  

Thank you

KarthikSubbarao:
> Would the 6 CSM commands in the Redis protocol today be applicable to be included here?

Hi @zmccoy , the CSM commands are not currently in the valkey-bloom module

> We're users of Redis' count-min sketch commands today and would like to migrate to Valkey, but this is one of our hold ups. I'd be happy to take guidance on how adding another structure into this crate should look, and happy to work through getting it added.

That sounds great! I would be happy to help talk to you about how we can go about adding this data type to the module and provide suggestions / learnings from adding the bloom data type
Are you on the Valkey [Slack](https://valkey.io/slack)?

zmccoy:
Hi @KarthikSubbarao, just joined the Slack a few minutes ago.  I'd love to talk about how to go about this and the learnings you've had.

zmccoy:
Adding some context of CMS Implementation detail library bits here in case folk have opinions on other things to look at and I'm of course happy to hear ideas.  By no means am I an expert in CMS libraries and would love some thoughts. 

The most utilized, and the same author as the Bloomfilter we're using is [rust-count-min-sketch](https://github.com/jedisct1/rust-count-min-sketch).  

For usability in this there are a few things that need changed and possibly a few issues: 
1.  Width and Depth would need to be exposed via getter methods as today there is no way to inspect the size after creation which is done in power of 2 sizing (more on that later)
2. A new_by_dimension function would need to be added which would clash a bit with the prior mentioned sizing.  Currently during creation it uses an [optimal_width](https://github.com/jedisct1/rust-count-min-sketch/blob/master/src/lib.rs#L139) function which sizes it to a power of 2.   I believe this is what allows it utilize the bitwise operators that it's using.  I assume this behavior is too much of a deviation of behavior from how the CMS commands work in Redis? 
[...]

PR #106, opened 2026-06-12
Count min sketch Module Start - CMS.INITBYDIM & CMS.INITBYPROB

zmccoy:
This is the start of the count min sketch module, just broken up into more specific bite sized PRs to make it easier to review and talk about trade offs on.  If a maintainer of Valkey-Bloom could create a count-min-sketch branch in the repository I can point this and the next PRs towards it to have a flow of approval into that feature branch.  

This PR adds the CMS.INITBYDIM and CMS.INITBYPROB commands to Valkey-Bloom.  I decided to ignore the seeding properties that we will need later for MERGE until that time, which also requires the underlying libraries to support doing so which is outlined below.  

This PR has Valkey-bloom take a dependency on flowstats, but I believe the correct dependency to be on rust-count-min-sketch which is outlined here: https://github.com/valkey-io/valkey-bloom/issues/84.  This comes with a trade-off outlined in that issue, needing the PRs for the methods of dimension initialization as well as having the width as a power of 2.  I'm happy to PR these to that repository if we feel that's the best chocie.

I'm more than happy to make changes and take suggestions as much of Valkey's internals were new to me.  Thank you

### DIFF
--- a/src/cms/cms_command_handler.rs
+++ b/src/cms/cms_command_handler.rs
@@ -1,7 +1,58 @@
-use valkey_module::{Context, ValkeyError, ValkeyResult, ValkeyString, VALKEY_OK};
+use valkey_module::{Context, NotifyEvent, ValkeyError, ValkeyResult, ValkeyString, VALKEY_OK};
 
 use crate::cms::data_type::CMS_TYPE;
 use crate::cms::utils::{self, CMSObject};
+
+struct ReplicateArgs {
+    seed: [u8; 32],
+    args: Replications,
+}
+
+enum Replications {
+    ReplicateArgsDim { width: u64, depth: u64 },
+    ReplicateArgsProb { error: f64 },
+}
+
+fn replicate_and_notify_events(
+    ctx: &Context,
+    key_name: &ValkeyString,
+    init_operation: bool,
+    incr_operation: bool,
+    args: ReplicateArgs,
+) {
+    if init_operation {
+        let seed_str = ValkeyString::create_from_slice(std::ptr::null_mut(), "SEED".as_bytes());
+        let seed_val = ValkeyString::create_from_slice(std::ptr::null_mut(), &args.seed);
+
+        match args.args {
+            Replications::ReplicateArgsDim { width, depth } => {
+                let width_val = ValkeyString::create_from_slice(
+                    std::ptr::null_mut(),
+                    width.to_string().as_bytes(),
+                );
+                let depth_val = ValkeyString::create_from_slice(
+                    std::ptr::null_mut(),
+                    depth.to_string().as_bytes(),
+                );
+                let cmd = vec![&width_val, &depth_val, &seed_str, &seed_val];
+                ctx.replicate("CMS.INITBYDIM", cmd.as_slice());
+                ctx.notify_keyspace_event(NotifyEvent::GENERIC, utils::INITBYDIM_EVENT, key_name);
+            }
+            Replications::ReplicateArgsProb { error } => {
+                let error_val = ValkeyString::create_from_slice(
+                    std::ptr::null_mut(),
+                    error.to_string().as_bytes(),
+                );
+                let cmd = vec![&error_val, &seed_str, &seed_val];
+                ctx.replicate("CMS.INITBYPROB", cmd.as_slice());
+                ctx.notify_keyspace_event(NotifyEvent::GENERIC, utils::INITBYPROB_EVENT, key_name);
+            }
+        }
+    } else if incr_operation {
+        ctx.replicate_verbatim();
+        ctx.notify_keyspace_event(NotifyEvent::GENERIC, utils::INCR_EVENT, key_name);
+    }
+}
 
 /// Function that implements logic to handle the CMS.INITBYDIM command.
 pub fn cms_initialize_by_dimensions(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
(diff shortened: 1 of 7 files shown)


zackcam:
Hi Zach! Thanks for starting this off! I'll look at creating a branch over the weekend or Monday and start helping out on reviews!

zackcam:
Hi, I just created a count-min-sketch branch, for us to use. I'll also try and take some time and review in the next day or so!

src/cms/utils.rs
```diff
@@ -0,0 +1,120 @@
...
+
+impl CMSObject {
+    pub fn new_by_dimension(width: u64, depth: u64) -> Result<CMSObject, CMSError> {
+        if width < 1 {
+            return Err(CMSError::InvalidWidth);
+        }
+
+        if depth < 1 {
+            return Err(CMSError::InvalidDepth);
+        }
```
KarthikSubbarao, line 55: Just checking. Are there any max ranges to enforce?

KarthikSubbarao: (if yes, we can add configs in the following PRs for this. But let's add a TODO)

src/cms/utils.rs
```diff
@@ -0,0 +1,120 @@
...
+pub const ITEM_EXISTS: &str = "ERR item exists";
+pub const BAD_WIDTH: &str = "ERR bad width";
+pub const BAD_DEPTH: &str = "ERR bad depth";
+pub const BAD_ERROR_RATE: &str = "ERR bad error rate";
+pub const ERROR_RATE_RANGE: &str = "ERR error rate should be between 0 and 1";
+pub const BAD_PROBABILITY: &str = "ERR bad probability";
+pub const PROBABILITY_RANGE: &str = "ERR probability rate should be between 0 and 1";
+pub const KEY_EXISTS: &str = "ERR Target key name already exists.";
+
+///Keyspace Notification Events
```
KarthikSubbarao, line 15: This is good for now. In the future, a merge also seems like a good event to add once it is supported

src/cms/cms_command_handler.rs
```diff
@@ -0,0 +1,164 @@
...
+
+fn replicate_and_notify_events(
+    ctx: &Context,
+    key_name: &ValkeyString,
+    init_operation: bool,
+    incr_operation: bool,
+    args: ReplicateArgs,
+) {
+    if init_operation {
+        //For later when we need seeding discussions (cms.merge)
```
KarthikSubbarao: Don't we need seeding even now? To ensure creations are deterministic on the replica as they were on the primary, we will need to pass the seed, right? (It is not just for merge AFAIK)

src/cms/cms_command_handler.rs
```diff
@@ -0,0 +1,164 @@
...
+enum Replications {
+    ReplicateArgsDim { width: u64, depth: u64 },
+    ReplicateArgsProb { error_rate: f64, fp_rate: f64 },
+}
+
+fn replicate_and_notify_events(
+    ctx: &Context,
+    key_name: &ValkeyString,
+    init_operation: bool,
+    incr_operation: bool,
```
KarthikSubbarao: Maybe an ENUM is better since we also have a merge operation to consider (unless we want to handle Merge by init)

KarthikSubbarao: In bloom, we have add_operation and reserve_operation as two separate bools since it is possible to do both in one operation. But in CMS, I think you can only do one or the other, right?

src/cms/data_type.rs
```diff
@@ -0,0 +1,36 @@
+use valkey_module::native_types::ValkeyType;
+use valkey_module::raw;
+
+const CMS_TYPE_ENCODING_VERSION: i32 = 1;
+
+//Note this is mocked out for now.
```
KarthikSubbarao, line 6: Sounds good, we can have these implemented as follow up PRs

src/cms/cms_command_handler.rs
KarthikSubbarao: Actually, I am reading the code...
https://github.com/vnvo/flowstats/blob/main/src/frequency/count_min.rs#L94

this is the package , right?

Looks like they use a fixed seed already. So, we don't need to replicate the seed. we can replicate the other stuff

src/cms/cms_command_handler.rs
zmccoy: I believe that's correct, I'll go and double check that.  I followed that pattern a little blindly, and that makes sense as to why they were separate bools rather than a sum type then.

src/cms/cms_command_handler.rs
zmccoy: > Don't we need seeding even now? To ensure creations are deterministic on the replica as they were on the primary, we will need to pass the seed, right? (It is not just for merge AFAIK)

Yes, that's true that was an oversight by myself.  


> Actually, I am reading the code...
[vnvo/flowstats@main/src/frequency/count_min.rs#L94](https://github.com/vnvo/flowstats/blob/main/src/frequency/count_min.rs#L94)
this is the package , right?
Looks like they use a fixed seed already. So, we don't need to replicate the seed. we can replicate the other stuff

This was written with the intent of using https://github.com/jedisct1/rust-count-min-sketch/blob/master/src/lib.rs which does not have a static seed, and also needs the addition of the methods noted in the PR.  If we're okay with keeping using flowstats I'm happy to as well and will push forward doing so.

src/cms/utils.rs
zmccoy: Yep, I plan to add them as we add each ability.  I was avoiding having a bunch of unused values cluttering compilation.

src/cms/utils.rs
zmccoy: I don't see any restrictions on sizing in the library itself for the creation of the cms.  Are you thinking of anything else specific?   I added a TODO to come back to this in case we want to set some maxes via config as well.

zmccoy:
@KarthikSubbarao The comments should be fixed up by the latest commit.  Thank you.

KarthikSubbarao (approved):

PR #107, opened 2026-06-17
Add TopK core metrics, deep-copy COPY support, and basic test coverage

detemmienation:
## Summary

Builds on the existing TopK commands by adding observability metrics, making the `COPY` command produce a true deep copy, and broadening integration test coverage.

### Metrics
- Expose four TopK metrics in a new `topk_core_metrics` INFO section:
  - `topk_total_memory_bytes` – total bytes across all TopK objects
  - `topk_num_objects` – number of TopK keys
  - `topk_num_items_across_objects` – total increments (ADD = +1, INCRBY = +increment)
  - `topk_sum_k_across_objects` – sum of `k` across all objects
- Metrics are incremented on object create / item add and decremented on `Drop`, so deleting a key removes its full contribution.

### Deep-copy COPY
- `TopKObject::create_copy_from` now clones the sketch contents (heavy/lobby cells and priority queue) and carries over the running item count.
- Requires `Clone` on the upstream `CuckooTopK`.

### Memory tracking
- `topk_total_memory_bytes`, `MEMORY USAGE`, and the new `TOPK.INFO` `size` field all derive from one source: `TopKObject::memory_usage()` = `size_of::<TopKObject>()` + the sketch's reported heap usage.

- **Before:** we used the upstream crate's `mem_bytes()`, which counts the sketch's container allocations (lobby/heavy cell arrays, decay-threshold table, priority-queue containers) but only `size_of::<T>()` for each tracked item. Since our items are `Vec<u8>`, that counts the 24-byte vector header but **not the bytes each item points to**, so a sketch tracking short items and one tracking long items reported the same size. This was the largest source of under-counting, and it grew with item length and `k` (up to ~16% in measurements below).

- **After:** the upstream crate now exposes `mem_bytes_with(|item| ...)`, which takes a closure for the per-item heap size. Since our `T` is fixed to `Vec<u8>`, we pass `|item| item.capacity()`, so the variable-length item contents are now counted. This drops the under-count to ~1–4%.

- **Why it is still an estimate:** the remaining gap is the priority queue's `HashMap`. Std's `HashMap` is hashbrown under the hood, but hashbrown does not expose its real allocation size as public API. We could model it exactly (bucket count = `next_power_of_two(capacity * 8 / 7)` plus the control-byte region) and hit 100% accuracy vs dhat — but that hardcodes hashbrown internals (the 7/8 load factor, the 16-byte SSE2 group width) that aren't part of any public contract, so it would break silently on other platforms or if std changes its implementation.

To quantify both estimates, I used [dhat](https://docs.rs/dhat) as the global allocator and compared its live-heap bytes against the sketch's reported usage, varying one dimension at a time off a baseline (`Vec<u8>` items, depth=4, decay=0.9, 200k distinct items):

| Config | `mem_bytes()` | dhat actual | Under-count | `mem_bytes_with()` | Under-count |
|---|--:|--:|--:|--:|--:|
| baseline (k=1k, w=8k, item=12) | 793,048 B | 827,560 B | +34,512 B (+4.2%) | 817,048 B | +10,512 B (+1.3%) |
| long items (item=32) | 793,048 B | 867,560 B | +74,512 B (+8.6%) | 857,048 B | +10,512 B (+1.2%) |
| long items (item=64) | 793,048 B | 931,560 B | +138,512 B (+14.9%) | 921,048 B | +10,512 B (+1.1%) |
| large k (k=10k) | 1,811,352 B | 2,135,336 B | +323,984 B (+15.2%) | 2,051,352 B | +83,984 B (+3.9%) |
| large k (k=50k) | 5,814,680 B | 6,951,856 B | +1,137,176 B (+16.4%) | 6,615,968 B | +335,888 B (+4.8%) |
| small width (w=1k) | 219,608 B | 254,120 B | +34,512 B (+13.6%) | 243,608 B | +10,512 B (+4.1%) |
| large width (w=64k) | 5,380,568 B | 5,415,080 B | +34,512 B (+0.6%) | 5,404,568 B | +10,512 B (+0.2%) |

`mem_bytes()`'s under-count grows with item length and `k`, since neither is reflected in `size_of::<T>()`. `mem_bytes_with()` removes the item-length component entirely, only scales with `k` (the `HashMap` rounding described above).

### TOPK.INFO
- Add a `size` field reporting the object's estimated memory in bytes, so per-object memory is observable from the command path. Comes from the same `TopKObject::memory_usage()` described above.

### Tests
- `test_topk_metrics.py`: metric accounting across RESERVE/ADD/INCRBY, read-only commands, DEL, COPY, and FLUSH.
- `test_topk_basic.py`: module load & command registration, data type name/encoding, generic key access, MULTI/EXEC, Lua, DEL/UNLINK/FLUSHALL, expiration, and WRONGTYPE errors.
- `test_topk_command.py`: asserts the `TOPK.INFO` `size` field is positive and within `MEMORY USAGE`.

### TODO: TOPK.ADD/INCRBY overhead

Counting per-item bytes means memory now changes on add, so the add path diffs `memory_usage()` before/after to keep the metric correct because `add_with_evicted` doesn't tell us whether an item was inserted. This makes add O(width·depth + k) ×2 instead of O(1). 
Follow-up: have the upstream `add_with_evicted` report whether an item was inserted  so the gauge can update in O(item bytes) and add stays O(1).

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -12,7 +12,7 @@ lazy_static! {
     pub static ref BLOOM_DEFRAG_MISSES: AtomicU64 = AtomicU64::new(0);
     pub static ref TOPK_NUM_OBJECTS: AtomicU64 = AtomicU64::new(0);
     pub static ref TOPK_OBJECT_TOTAL_MEMORY_BYTES: AtomicUsize = AtomicUsize::new(0);
-    pub static ref TOPK_NUM_ITEMS_ACROSS_OBJECTS: AtomicU64 = AtomicU64::new(0);
+    pub static ref TOPK_TOTAL_ITEMS_ADDED_ACROSS_OBJECTS: AtomicU64 = AtomicU64::new(0);
     pub static ref TOPK_SUM_K_ACROSS_OBJECTS: AtomicU64 = AtomicU64::new(0);
 }
 
@@ -77,8 +77,8 @@ pub fn topk_info_handler(ctx: &InfoContext) -> ValkeyResult<()> {
             TOPK_NUM_OBJECTS.load(Ordering::Relaxed).to_string(),
         )?
         .field(
-            "topk_num_items_across_objects",
-            TOPK_NUM_ITEMS_ACROSS_OBJECTS
+            "topk_total_items_added_across_objects",
+            TOPK_TOTAL_ITEMS_ADDED_ACROSS_OBJECTS
                 .load(Ordering::Relaxed)
                 .to_string(),
         )?
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -10,6 +10,10 @@ lazy_static! {
     pub static ref BLOOM_CAPACITY_ACROSS_OBJECTS: AtomicU64 = AtomicU64::new(0);
     pub static ref BLOOM_DEFRAG_HITS: AtomicU64 = AtomicU64::new(0);
     pub static ref BLOOM_DEFRAG_MISSES: AtomicU64 = AtomicU64::new(0);
+    pub static ref TOPK_NUM_OBJECTS: AtomicU64 = AtomicU64::new(0);
+    pub static ref TOPK_OBJECT_TOTAL_MEMORY_BYTES: AtomicUsize = AtomicUsize::new(0);
+    pub static ref TOPK_NUM_ITEMS_ACROSS_OBJECTS: AtomicU64 = AtomicU64::new(0);
+    pub static ref TOPK_SUM_K_ACROSS_OBJECTS: AtomicU64 = AtomicU64::new(0);
 }
 
 pub fn bloom_info_handler(ctx: &InfoContext) -> ValkeyResult<()> {
@@ -52,6 +56,37 @@ pub fn bloom_info_handler(ctx: &InfoContext) -> ValkeyResult<()> {
         .field(
             "bloom_defrag_misses",
             BLOOM_DEFRAG_MISSES.load(Ordering::Relaxed).to_string(),
+        )?
+        .build_section()?
+        .build_info()?;
+
+    Ok(())
+}
+
+pub fn topk_info_handler(ctx: &InfoContext) -> ValkeyResult<()> {
+    ctx.builder()
+        .add_section("topk_core_metrics")
+        .field(
+            "topk_total_memory_bytes",
+            TOPK_OBJECT_TOTAL_MEMORY_BYTES
+                .load(Ordering::Relaxed)
+                .to_string(),
+        )?
+        .field(
+            "topk_num_objects",
+            TOPK_NUM_OBJECTS.load(Ordering::Relaxed).to_string(),
+        )?
+        .field(
+            "topk_num_items_across_objects",
+            TOPK_NUM_ITEMS_ACROSS_OBJECTS
+                .load(Ordering::Relaxed)
+                .to_string(),
+        )?
+        .field(
+            "topk_sum_k_across_objects",
+            TOPK_SUM_K_ACROSS_OBJECTS
+                .load(Ordering::Relaxed)
+                .to_string(),
         )?
         .build_section()?
         .build_info()?;
--- a/src/topk/utils.rs
+++ b/src/topk/utils.rs
@@ -84,6 +84,15 @@ impl TopKObject {
         topk
     }
 
+    /// Estimated total memory (in bytes) of this object: the wrapper struct
+    /// plus the heap the underlying sketch owns (lobby/heavy cell arrays, the
+    /// decay-threshold table, and the priority-queue allocations). Excludes the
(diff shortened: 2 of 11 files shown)


src/topk/utils.rs
```diff
@@ -44,21 +46,51 @@ pub struct TopKObject {
     decay: f64,
     seed: u64,
     sketch: CuckooTopK<Vec<u8>>,
+    num_items: u64,
```
zackcam, line 49: I know we need this for tracking number of items added, this should only be an 8 byte increase per topk object correct?

tests/test_topk_basic.py
```diff
@@ -1,15 +1,146 @@
...
+        command_cmd_result = client.execute_command('COMMAND')
+        topk_cmds = ["TOPK.RESERVE", "TOPK.ADD", "TOPK.INCRBY", "TOPK.INFO", "TOPK.LIST", "TOPK.COUNT", "TOPK.QUERY"]
+        assert all(item in command_cmd_result for item in topk_cmds)
+        # Basic create, add, and membership query.
+        assert client.execute_command('TOPK.RESERVE tk 3 50 4 0.9 SEED 42') == b'OK'
+        assert client.execute_command('TOPK.ADD tk apple banana cherry') == [None, None, None]
+        assert client.execute_command('TOPK.QUERY tk apple') == [1]
+        assert client.execute_command('TOPK.QUERY tk missing') == [0]
+
+    def test_copy(self):
```
zackcam, line 26: Can we expand this test a little to do some adds and evicts after in order to make sure seeding and everything is copied as we want. Pretty sure it is but worth making sure it is

tests/test_topk_basic.py
```diff
@@ -1,15 +1,146 @@
...
+        self.verify_server_key_count(client, 1)
+        # cmd persist
+        assert client.execute_command('TOPK.RESERVE TEST_PERSIST 3 50 4 0.9') == b'OK'
+        assert client.execute_command('TTL TEST_PERSIST') == -1
+        assert client.execute_command(f'EXPIREAT TEST_PERSIST {curr_time + 100000}') == 1
+        assert client.execute_command('TTL TEST_PERSIST') > 0
+        assert client.execute_command('PERSIST TEST_PERSIST') == 1
+        assert client.execute_command('TTL TEST_PERSIST') == -1
+
+    def test_topk_wrong_type(self):
```
zackcam: We already kinda cover this in the test for commands. I think maybe we can skip this one? Let me know how you feel on this

src/metrics.rs
```diff
@@ -58,3 +62,34 @@ pub fn bloom_info_handler(ctx: &InfoContext) -> ValkeyResult<()> {
...
+            TOPK_OBJECT_TOTAL_MEMORY_BYTES
+                .load(Ordering::Relaxed)
+                .to_string(),
+        )?
+        .field(
+            "topk_num_objects",
+            TOPK_NUM_OBJECTS.load(Ordering::Relaxed).to_string(),
+        )?
+        .field(
+            "topk_num_items_across_objects",
```
zackcam: I think we might need to disambiguate here a little. We want to clarify that its not current items across object. I'm up for suggestions on name but something like `topk_total_items_added_across_objects` maybe?

zackcam (commented):
Is there a way to get a best guess of memory usage for now per topk object. I wonder if we should do somehting like:
calculate cell arrays: width * depth * size_of::<Cell>() for both lobbies and heavy (the lobby/heavy layout), plus decay_thresholds (depth * size_of::<u64>() or similar).
Then add the k-bounded priority queue: roughly k * (size_of::<entry>() + average_key_len).

tests/test_topk_basic.py
detemmienation: Yeah I agree, removed.

src/metrics.rs
detemmienation: Done as suggested

src/topk/utils.rs
detemmienation: Yeah, 8 bytes per project

src/topk/utils.rs
```diff
@@ -84,13 +84,21 @@ impl TopKObject {
         topk
     }
 
+    /// Estimated total memory (in bytes) of this object: the wrapper struct
+    /// plus the heap the underlying sketch owns (lobby/heavy cell arrays, the
+    /// decay-threshold table, and the priority-queue allocations). Excludes the
+    /// variable byte contents of individual tracked keys, which `mem_bytes`
```
zackcam: Can we have a more in depth explanation of why the tracking can't be accurate, in the pr description so this way others can also understand the tracking a bit better

detemmienation: Updated, also add `topk.info size` related description

zackcam (approved):

PR #108, opened 2026-06-29
Add TopK RDB save/load support

detemmienation:
## Summary

Adds RDB persistence for TopK so sketches survive a server restart, plus the supporting changes: a faster `TOPK.ADD`/`TOPK.INCRBY` metric path, exact memory tracking, a richer `TOPK.INFO`, and a digest callback.

### RDB save/load
- Wire up the `rdb_save` / `rdb_load` type callbacks .
- **Save** persists only what cannot be recomputed: `seed`, `num_items`, and the serialized sketch bytes (`sketch.to_bytes()`).
- **Load** rebuilds the sketch via `CuckooTopK::from_bytes(bytes, seed)` and reads `k` / `width` / `depth` / `decay` back off the restored sketch, so only three values live in the on-disk layout.
- Load is defensive: it rejects an encoding version newer than the module supports, a sketch that fails to deserialize, and `k` / `width` / `depth` / `decay` values that fall outside their valid ranges — logging a warning and failing the load rather than constructing a corrupt object.
- `test_topk_save_and_restore.py`: basic round-trip (params, item list via `TOPK.LIST`, and `num_items` survive a BGSAVE + restart) and a 500-key bulk-survival test.

### O(1) `TOPK.ADD` / `TOPK.INCRBY`
- The add path no longer diffs `memory_usage()` before/after to keep the memory metric correct. The upstream `add_with_evicted` now reports whether an item was inserted, so the gauge updates from the actual byte delta in O(item bytes) and add returns to O(1).

### Memory tracking
- `TopKObject::memory_usage()` now uses the upstream `mem_bytes(|item| item.capacity())`, which counts each tracked item's heap (the `Vec<u8>` contents). The upstream crate now models its `HashMap` allocation exactly, so the reported size matches actual heap usage.

### TOPK.INFO
- Add a `total items added` field, reporting the running item count (`num_items`) that is persisted in the RDB. This is the value tracked by the `topk_total_items_added_across_objects` metric, now observable per-object.
- Add an optional field argument: `TOPK.INFO key [K | WIDTH | DEPTH | DECAY | SIZE | TOTALITEMSADDED]` returns just that single value instead of the full reply. An unrecognized field token returns `invalid information value`.
- `test_topk_command.py`: `total items added` accounting and single-field `TOPK.INFO` queries, plus the unrecognized-token error path.
- `test_topk_acl_category.py`: updated `TOPK.INFO` field count for the new field.

### Digest
- Wire up the `digest` callback so `DEBUG DIGEST` / `DEBUG DIGEST-VALUE` work on TopK keys. The digest covers `seed`, `num_items`, and the serialized sketch bytes, so two equal sketches digest identically.
- `test_topk_save_and_restore.py`: assert `DEBUG DIGEST` and `DIGEST-VALUE` round-trip unchanged across an RDB save and restart.
- `test_topk_basic.py`: `test_copy_and_exists_cmd` verifies `COPY` yields a digest-equal object while a second key changes the whole-server digest and diverging contents digest differently.
- Replication tests to follow.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ valkey-module-macros = "0"
 linkme = "0"
 bloomfilter = { version = "3.0.1", features = ["serde"] }
 # TODO: This is a temporary solution until we add this functionality to the original heavykeeper crate.
-heavykeeper = { git = "https://github.com/detemmienation/heavykeeper-rs.git", branch = "mem" }
+heavykeeper = { git = "https://github.com/detemmienation/heavykeeper-rs.git", branch = "mem-rdb" }
 lazy_static = "1.4.0"
 libc = "0.2"
 serde = { version = "1.0", features = ["derive"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ valkey-module-macros = "0"
 linkme = "0"
 bloomfilter = { version = "3.0.1", features = ["serde"] }
 # TODO: This is a temporary solution until we add this functionality to the original heavykeeper crate.
-heavykeeper = { git = "https://github.com/detemmienation/heavykeeper-rs.git", branch = "rdb" }
+heavykeeper = { git = "https://github.com/detemmienation/heavykeeper-rs.git", branch = "rdb-rng" }
 lazy_static = "1.4.0"
 libc = "0.2"
 serde = { version = "1.0", features = ["derive"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ valkey-module-macros = "0"
 linkme = "0"
 bloomfilter = { version = "3.0.1", features = ["serde"] }
 # TODO: This is a temporary solution until we add this functionality to the original heavykeeper crate.
-heavykeeper = { git = "https://github.com/detemmienation/heavykeeper-rs.git", branch = "mem-rdb" }
+heavykeeper = { git = "https://github.com/detemmienation/heavykeeper-rs.git", branch = "rdb" }
 lazy_static = "1.4.0"
 libc = "0.2"
 serde = { version = "1.0", features = ["derive"] }
--- a/src/topk/command_handler.rs
+++ b/src/topk/command_handler.rs
@@ -255,8 +255,9 @@ pub fn topk_incrby(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
 /// Syntax:
 ///     TOPK.INFO key
 ///
-/// Returns the number of required items (k), width, depth, decay, and the
-/// estimated memory size (in bytes) of the sketch stored at `key`.
+/// Returns the number of required items (k), width, depth, decay, the
+/// estimated memory size (in bytes), and the total number of items added
+/// (total_items_added) for the sketch stored at `key`.
 pub fn topk_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
     if input_args.len() != 2 {
         return Err(ValkeyError::WrongArity);
@@ -281,6 +282,8 @@ pub fn topk_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
         ValkeyValue::Float(topk.decay()),
         ValkeyValue::SimpleStringStatic("size"),
         ValkeyValue::Integer(topk.memory_usage() as i64),
+        ValkeyValue::SimpleStringStatic("total items added"),
+        ValkeyValue::Integer(topk.num_items() as i64),
     ];
     Ok(ValkeyValue::Array(result))
 }
--- a/src/topk/command_handler.rs
+++ b/src/topk/command_handler.rs
@@ -289,17 +289,17 @@ pub fn topk_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
     }
 
     let result = vec![
-        ValkeyValue::SimpleStringStatic("k"),
+        ValkeyValue::SimpleStringStatic("K"),
         ValkeyValue::Integer(topk.k() as i64),
-        ValkeyValue::SimpleStringStatic("width"),
+        ValkeyValue::SimpleStringStatic("Width"),
         ValkeyValue::Integer(topk.width() as i64),
-        ValkeyValue::SimpleStringStatic("depth"),
+        ValkeyValue::SimpleStringStatic("Depth"),
         ValkeyValue::Integer(topk.depth() as i64),
-        ValkeyValue::SimpleStringStatic("decay"),
+        ValkeyValue::SimpleStringStatic("Decay"),
         ValkeyValue::Float(topk.decay()),
-        ValkeyValue::SimpleStringStatic("size"),
+        ValkeyValue::SimpleStringStatic("Size"),
         ValkeyValue::Integer(topk.memory_usage() as i64),
-        ValkeyValue::SimpleStringStatic("total items added"),
+        ValkeyValue::SimpleStringStatic("Total items added"),
         ValkeyValue::Integer(topk.num_items() as i64),
     ];
     Ok(ValkeyValue::Array(result))
--- a/src/topk/command_handler.rs
+++ b/src/topk/command_handler.rs
@@ -253,13 +253,15 @@ pub fn topk_incrby(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
 /// Handle TOPK.INFO.
 ///
 /// Syntax:
-///     TOPK.INFO key
+///     TOPK.INFO key [K | WIDTH | DEPTH | DECAY | SIZE | TOTALITEMSADDED]
 ///
-/// Returns the number of required items (k), width, depth, decay, the
-/// estimated memory size (in bytes), and the total number of items added
-/// (total_items_added) for the sketch stored at `key`.
+/// Without a field argument, returns the number of required items (k), width,
+/// depth, decay, the estimated memory size (in bytes), and the total number of
+/// items added for the sketch stored at `key`.
+/// With a field argument, returns only that single value.
 pub fn topk_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
-    if input_args.len() != 2 {
+    let argc = input_args.len();
+    if !(2..=3).contains(&argc) {
         return Err(ValkeyError::WrongArity);
     }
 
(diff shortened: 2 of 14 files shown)


src/topk/command_handler.rs
```diff
@@ -281,6 +299,8 @@ pub fn topk_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
         ValkeyValue::Float(topk.decay()),
         ValkeyValue::SimpleStringStatic("size"),
         ValkeyValue::Integer(topk.memory_usage() as i64),
+        ValkeyValue::SimpleStringStatic("total items added"),
```
zackcam: Nit: We should probably update all of these to start capitalized

src/topk/utils.rs
```diff
@@ -131,20 +158,15 @@ impl TopKObject {
...
-        let evicted = self.sketch.add_with_evicted(item, increment);
-        let mem_after = self.memory_usage();
-        if mem_after >= mem_before {
+        let (evicted, inserted) = self.sketch.add_with_evicted(item, increment);
+        let added = if inserted { item.len() } else { 0 };
+        let removed = evicted.as_ref().map_or(0, Vec::len);
+        if added >= removed {
             metrics::TOPK_OBJECT_TOTAL_MEMORY_BYTES
-                .fetch_add(mem_after - mem_before, Ordering::Relaxed);
+                .fetch_add(2 * (added - removed), Ordering::Relaxed);
```
zackcam, line 168: We should add a note why we do a *2 here. I know its the right approach but should document it

tests/test_topk_save_and_restore.py
```diff
@@ -0,0 +1,55 @@
...
+        assert item_count_after == item_count_before
+        assert client.execute_command('TOPK.INFO testSave') == info_before
+        assert client.execute_command('TOPK.LIST testSave WITHCOUNT') == list_before
+
+    def test_basic_save_many(self):
+        client = self.server.get_new_client()
+        count = 500
+        for i in range(0, count):
+            name = str(i) + "key"
+            assert client.execute_command('TOPK.RESERVE ' + name + ' 5 50 4 0.9 SEED 42') == b'OK'
```
zackcam: Could we perhaps randomize the values in this? Or maybe not randomize but make it so theres a couple different values

tests/test_topk_save_and_restore.py
zackcam, line 1: Can we also try a test where we load an rdb on non module enabled server?

tests/test_topk_command.py
```diff
@@ -172,6 +177,20 @@ def info_dict(key):
         assert info[b'decay'] == b'0.5'
         assert info[b'size'] > 0
         assert info[b'size'] <= self.client.execute_command('MEMORY USAGE tk5')
+        assert info[b'total items added'] == 0
+
+        # tk_incr accumulates every increment:
+        # 1 + (5+3) + (1+2+3) + (7+2) + 1000000 + (2+3) = 1000029.
```
zackcam: Maybe we can store this as a variable so in case tests change later we only need to update in one place

zackcam (commented):
For the debug digest can we do a quick manual test of a primary and replica where we do large number of additions into the topk set and then compare the digests of the replica and primary? this way we can make sure that the logic between replica and primary is being done equally

tests/test_topk_save_and_restore.py
detemmienation: Done in https://github.com/valkey-io/valkey-bloom/pull/108/changes/3752f4f2a7c602f7c4239818b735a96e73805e86, also extend `test_basic_save_and_restore` to first add 3 items, then add more to exceed 5 for better test coverage.

tests/test_topk_save_and_restore.py
detemmienation: Done in https://github.com/valkey-io/valkey-bloom/pull/108/changes/a322cc63bd45e86ef269defb9f6ca94f898e1415 🫡

zackcam (commented):
For the RDB load as well, i took a look at the underlying crate changes. I think we way have some divergence issues which this didn't introduce but made more noticeable. Lets take the following scenario:

Lets say we have a two node shard of a primary and replica (Can be any number of replicas still same issue). The replica has fallen too far behind, which means it will now need to do a fullsync. So immediately after the full sync, the two nodes hold identical cells and an identical top-k list (the RDB restored those exactly) but their RNGs are at different positions. The primary will have the rng where it was from live traffic while the replica will start the rng from the beginning causing the different positions. eg.
- Primary draws random numbers from phase N, N+1, N+2, etc
- Replica draws from phase 0, 1, 2, etc
As these draws are different, probabilistic decay outcomes differ and over time the replica's top-k list can diverge from the primary's.

To fix we should take a look at a way to preserve the location of this rng, I looked through SmallRng and did not see an easy way to do this. Even though it implements clone it doesn't allow serialization of this which is still a blocker. May need to look at a different crate if so https://docs.rs/rand_xoshiro/latest/rand_xoshiro/struct.Xoshiro256PlusPlus.html contains a state call which can be used

tests/test_topk_command.py
detemmienation: Done in https://github.com/valkey-io/valkey-bloom/pull/108/changes/ad41ea77a035b79a0e45590fbe0fd930bab4d63c

detemmienation:
> For the debug digest can we do a quick manual test of a primary and replica where we do large number of additions into the topk set and then compare the digests of the replica and primary? this way we can make sure that the logic between replica and primary is being done equally

Done in https://github.com/valkey-io/valkey-bloom/pull/108/changes/036bf73c417a2c1919103694b79982364cb971c0
Added replication tests: `test_replication_behavior` and `test_topk_digest_consistency_after_many_adds`
Skipped deterministic test that bloom has because TopK has no implicit configs and all params are explicit `TOPK.RESERVE` arguments, so this test has no TopK equivalent. The generated-seed case is already covered by many_adds test

Also added `TopkFixedSeedMixin` to skip redundant seed parameterization in commit https://github.com/valkey-io/valkey-bloom/pull/108/commits/25416c8bda665f5b9972d2be5b5340a7d2241803

detemmienation:
> For the RDB load as well, i took a look at the underlying crate changes. I think we way have some divergence issues which this didn't introduce but made more noticeable. Lets take the following scenario:
> 
> Lets say we have a two node shard of a primary and replica (Can be any number of replicas still same issue). The replica has fallen too far behind, which means it will now need to do a fullsync. So immediately after the full sync, the two nodes hold identical cells and an identical top-k list (the RDB restored those exactly) but their RNGs are at different positions. The primary will have the rng where it was from live traffic while the replica will start the rng from the beginning causing the different positions. eg.
> 
> * Primary draws random numbers from phase N, N+1, N+2, etc
> * Replica draws from phase 0, 1, 2, etc
>   As these draws are different, probabilistic decay outcomes differ and over time the replica's top-k list can diverge from the primary's.
> 
> To fix we should take a look at a way to preserve the location of this rng, I looked through SmallRng and did not see an easy way to do this. Even though it implements clone it doesn't allow serialization of this which is still a blocker. May need to look at a different crate if so https://docs.rs/rand_xoshiro/latest/rand_xoshiro/struct.Xoshiro256PlusPlus.html contains a state call which can be used

Following up on this: we likely do need to switch to the new crate. 🙌 
 I've updated the RNG-related parts in the upstream crate in commit https://github.com/detemmienation/heavykeeper-rs/commit/ce78d71f06a92dea85cd69fab41149b7f79ddd64 on the rdb-rng branch, and updated the local dependency to point to that branch in commit https://github.com/valkey-io/valkey-bloom/pull/108/commits/980519e3c50659c5426ce472950007a689bd7079.

Cargo.toml
```diff
@@ -17,7 +17,7 @@ valkey-module-macros = "0"
 linkme = "0"
 bloomfilter = { version = "3.0.1", features = ["serde"] }
 # TODO: This is a temporary solution until we add this functionality to the original heavykeeper crate.
-heavykeeper = { git = "https://github.com/detemmienation/heavykeeper-rs.git", branch = "rdb" }
+heavykeeper = { git = "https://github.com/detemmienation/heavykeeper-rs.git", branch = "rdb-rng" }
```
zackcam, line 20: Could we perhaps do a quick perf check on the change in rng to make sure theres not a big degradation when we calculate decay?

tests/test_topk_replication.py
```diff
@@ -0,0 +1,146 @@
+import pytest, os
+from valkey import ResponseError
+from valkeytestframework.valkey_test_case import ReplicationTestCase
+from valkeytestframework.conftest import resource_port_tracker  # noqa: F401
+from valkey_bloom_test_case import TopkFixedSeedMixin
+
+class TestTopkReplication(TopkFixedSeedMixin, ReplicationTestCase):
+
+    @pytest.fixture(autouse=True)
+    def setup_test(self, setup):
```
zackcam, line 10: We can combine this with the setup_test already written in bloom replication test

tests/valkey_bloom_test_case.py
```diff
@@ -404,3 +404,13 @@ def verify_command_acl_categories(self, commands):
             for category in cmd[2]:
                 assert category in cmd_info[0][6]
 
+class TopkFixedSeedMixin:
```
zackcam: Lets disambiguate this from topk as other future data types might want this. Maybe can name skip_fixed_random_behaviour. Probably still not a great name tbh so if you have one feel free to use it

tests/test_topk_replication.py
detemmienation: Done in commit https://github.com/valkey-io/valkey-bloom/pull/108/commits/d1b5c25ef6620e14241827bf0d04fc27ca61d5a6

tests/valkey_bloom_test_case.py
detemmienation: It is a nice name ^^
Renamed in commit https://github.com/valkey-io/valkey-bloom/pull/108/commits/f27db0088652d99a601ca2cf162cd9164df72493 !

Cargo.toml
detemmienation: Ran a decay-path benchmark to check for regression from the RNG swap.

[criterion](https://docs.rs/criterion)  benchmark on the decay hot path (`decay_lobby_and_maybe_replace`), which draws one `rng.next_u64()` per unit of increment. Production params (`k=50 width=50 depth=4 decay=0.9`), 800 hot keys, varying increment to scale RNG draws per `add`. 
Same bench run on the parent commit (`SmallRng`) vs the RNG-change commit (`Xoshiro256PlusPlus`).

| increment | SmallRng | Xoshiro256PlusPlus | Δ |
|---|---|---|---|
| 1   | 50.38 µs | 50.79 µs | +0.68% |
| 7   | 59.60 µs | 59.93 µs | +0.74% |
| 64  | 96.91 µs | 97.29 µs | +0.30% |
| 512 | 460.9 µs | 461.6 µs | +0.43% |

All within ±1%, criterion flagged every case as "within noise threshold".

zackcam (approved):

PR #109, opened 2026-07-08
Add TopK defrag support

detemmienation:
## Summary

Adds defrag support for the TopK data type, so a TopK sketch's memory can be relocated out of fragmented allocator regions during the background defrag pass. The relocation itself is delegated to the [heavykeeper fork](https://github.com/detemmienation/heavykeeper-rs/tree/rdb-defrag)'s `Reallocator` trait.

It also adds a per-object memory-usage limit config, guarding against a single TOPK.RESERVE allocating unbounded memory.

### Defrag callback
- Wire up the `defrag` type callback.
- Add a `Defragger` implementing heavykeeper's `Reallocator`, backed by Valkey's defrag allocator. A null allocator result means "not relocated", so it keeps the original allocation.
- `topk_defrag` relocates the sketch's heap allocations via `realloc_large_heap_allocated_objects`, then the `TopKObject` itself, in a single pass. No cursor: unlike the bloom, relocation is cheap enough to finish in one pass.
- Reconcile the memory gauge across a defrag: relocation trims spare `Vec` capacity (`into_boxed_slice`), which `memory_usage()` counts, so `topk_defrag` recomputes it before/after and applies the delta to keep the gauge balanced against `Drop`.

### Config
- Add the `topk-defrag-enabled` boolean config (default `true`). When off, the callback is a no-op.
- Add the `topk-memory-usage-limit` config (default 128MB, mirroring `bloom-memory-usage-limit`). A `TOPK.RESERVE` whose sketch would exceed it is rejected before allocating. `TopKObject::estimated_size` estimates the up-front allocation from `k`, `width`, and `depth` (cell arrays, decay table, and the priority queue reserved to capacity `k`), saturating so oversized params cannot wrap under the limit.

### Metrics
- Add `bf_topk_defrag_hits` / `bf_topk_defrag_misses`, exposed via a new `topk_defrag_metrics` INFO section.

### Tests
- `test_topk_defrag.py`: create many TopK objects, delete half to fragment the heap, run active defrag, and assert the callback ran (`bf_topk_defrag_hits > 0`), fragmentation dropped (`allocator_frag_ratio` below its pre-defrag value), and data is unchanged (`DEBUG DIGEST-VALUE` / `TOPK.LIST` equal before/after and across an RDB round-trip).
- `test_topk_basic.py`:
  - `test_too_large_topk_obj` asserts an oversized `TOPK.RESERVE` (huge `width * depth`, or a huge `k` alone) is rejected, and that lowering the limit rejects a sketch that was previously allowed.
  - `test_memory_usage_cmd` asserts a TopK object's engine-reported `MEMORY USAGE` is at least its `TOPK.INFO` Size.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -185,6 +185,7 @@ valkey_module! {
         bool: [
             ["bloom-use-random-seed", &*configs::BLOOM_USE_RANDOM_SEED, configs::BLOOM_USE_RANDOM_SEED_DEFAULT, ConfigurationFlags::DEFAULT, None],
             ["bloom-defrag-enabled", &*configs::BLOOM_DEFRAG, configs::BLOOM_DEFRAG_DEFAULT,  ConfigurationFlags::DEFAULT, None],
+            ["topk-defrag-enabled", &*configs::TOPK_DEFRAG, configs::TOPK_DEFRAG_DEFAULT,  ConfigurationFlags::DEFAULT, None],
         ],
         enum: [
         ],
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -177,6 +177,7 @@ valkey_module! {
             ["bloom-capacity", &*configs::BLOOM_CAPACITY, configs::BLOOM_CAPACITY_DEFAULT, configs::BLOOM_CAPACITY_MIN, configs::BLOOM_CAPACITY_MAX, ConfigurationFlags::DEFAULT, None],
             ["bloom-expansion", &*configs::BLOOM_EXPANSION, configs::BLOOM_EXPANSION_DEFAULT, 0, configs::BLOOM_EXPANSION_MAX as i64, ConfigurationFlags::DEFAULT, None],
             ["bloom-memory-usage-limit", &*configs::BLOOM_MEMORY_LIMIT_PER_OBJECT, configs::BLOOM_MEMORY_LIMIT_PER_OBJECT_DEFAULT, configs::BLOOM_MEMORY_LIMIT_PER_OBJECT_MIN, configs::BLOOM_MEMORY_LIMIT_PER_OBJECT_MAX, ConfigurationFlags::DEFAULT, None],
+            ["topk-memory-usage-limit", &*configs::TOPK_MEMORY_LIMIT_PER_OBJECT, configs::TOPK_MEMORY_LIMIT_PER_OBJECT_DEFAULT, configs::TOPK_MEMORY_LIMIT_PER_OBJECT_MIN, configs::TOPK_MEMORY_LIMIT_PER_OBJECT_MAX, ConfigurationFlags::DEFAULT, None],
         ],
         string: [
             ["bloom-fp-rate", &*configs::BLOOM_FP_RATE, configs::BLOOM_FP_RATE_DEFAULT, ConfigurationFlags::DEFAULT, None, Some(Box::new(configs::on_string_config_set))],
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -14,6 +14,8 @@ lazy_static! {
     pub static ref TOPK_OBJECT_TOTAL_MEMORY_BYTES: AtomicUsize = AtomicUsize::new(0);
     pub static ref TOPK_TOTAL_ITEMS_ADDED_ACROSS_OBJECTS: AtomicU64 = AtomicU64::new(0);
     pub static ref TOPK_SUM_K_ACROSS_OBJECTS: AtomicU64 = AtomicU64::new(0);
+    pub static ref TOPK_DEFRAG_HITS: AtomicU64 = AtomicU64::new(0);
+    pub static ref TOPK_DEFRAG_MISSES: AtomicU64 = AtomicU64::new(0);
 }
 
 pub fn bloom_info_handler(ctx: &InfoContext) -> ValkeyResult<()> {
@@ -87,6 +89,16 @@ pub fn topk_info_handler(ctx: &InfoContext) -> ValkeyResult<()> {
             TOPK_SUM_K_ACROSS_OBJECTS
                 .load(Ordering::Relaxed)
                 .to_string(),
+        )?
+        .build_section()?
+        .add_section("topk_defrag_metrics")
+        .field(
+            "topk_defrag_hits",
+            TOPK_DEFRAG_HITS.load(Ordering::Relaxed).to_string(),
+        )?
+        .field(
+            "topk_defrag_misses",
+            TOPK_DEFRAG_MISSES.load(Ordering::Relaxed).to_string(),
         )?
         .build_section()?
         .build_info()?;
--- a/src/topk/utils.rs
+++ b/src/topk/utils.rs
@@ -1,3 +1,4 @@
+use crate::configs;
 use crate::metrics;
 use heavykeeper::CuckooTopK;
 use std::sync::atomic::Ordering;
@@ -9,11 +10,9 @@ pub const DEFAULT_WIDTH: u32 = 8;
 pub const DEFAULT_DEPTH: u32 = 7;
 pub const DEFAULT_DECAY: f64 = 0.9;
 
-/// Per-argument bounds.
-/// The minimums are 1,The maximums are placeholders covering the
-/// full u32 range until topk-specific configs are introduced; tighten them
-/// once dedicated config knobs land.
-// TODO: replace these with configurable bounds once topk configs exist.
+/// Per-argument bounds. The minimums are 1; the maximums span the full u32
+/// range because oversized sketches are bounded by the topk-memory-usage-limit
+/// config (see `TopKObject::estimated_size`), not by per-argument caps.
 pub const TOPK_K_MIN: u32 = 1;
 pub const TOPK_K_MAX: u32 = u32::MAX;
 pub const TOPK_WIDTH_MIN: u32 = 1;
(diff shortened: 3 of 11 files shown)


src/lib.rs
```diff
@@ -185,6 +185,7 @@ valkey_module! {
         bool: [
             ["bloom-use-random-seed", &*configs::BLOOM_USE_RANDOM_SEED, configs::BLOOM_USE_RANDOM_SEED_DEFAULT, ConfigurationFlags::DEFAULT, None],
             ["bloom-defrag-enabled", &*configs::BLOOM_DEFRAG, configs::BLOOM_DEFRAG_DEFAULT,  ConfigurationFlags::DEFAULT, None],
+            ["topk-defrag-enabled", &*configs::TOPK_DEFRAG, configs::TOPK_DEFRAG_DEFAULT,  ConfigurationFlags::DEFAULT, None],
```
zackcam, line 189: I wonder if we should make a full defrag enabled module config to turn off all module defrags. Might be overkill, thoughts?

src/metrics.rs
```diff
@@ -89,6 +91,16 @@ pub fn topk_info_handler(ctx: &InfoContext) -> ValkeyResult<()> {
                 .to_string(),
         )?
         .build_section()?
+        .add_section("topk_defrag_metrics")
```
zackcam, line 94: Same idea above, wonder if having total module defrag would be useful or not. This one probably even less so. Wouldn't need extra tracking could just be all loads combined. Putting this out to see what others think as well

tests/test_topk_defrag.py
zackcam, line 1: I think we should add some work done on the topk object once its been defragged. Basically adding some items and evicting some. This will help catch to make sure expected behaviour still works as intended on a defragged topk object

src/wrapper/topk_callback.rs
```diff
@@ -75,3 +80,59 @@ pub unsafe extern "C" fn topk_free_effort(
...
     1
 }
+
+/// Reallocator passed to the heavykeeper sketch during defrag. Routes each
+/// block through Valkey's defrag allocator; a null result means "not
+/// relocated", so we keep the original allocation.
+///
+/// Assumes `align_of::<T>() <= 16`, the alignment the defrag allocator
+/// guarantees. All element types the sketch passes through today
+/// satisfy this; revisit if a more strictly aligned type is added.
```
zackcam: nit: I don't know if we need the revisit line in the comment

tests/test_topk_defrag.py
detemmienation: Done in https://github.com/valkey-io/valkey-bloom/pull/109/commits/4ad8c0dafda33ac0fe07cd9ae43e0cb57ee75ac6 🙌

src/wrapper/topk_callback.rs
detemmienation: Removed, also in https://github.com/valkey-io/valkey-bloom/pull/109/commits/4ad8c0dafda33ac0fe07cd9ae43e0cb57ee75ac6

src/wrapper/topk_callback.rs
```diff
@@ -75,3 +80,56 @@ pub unsafe extern "C" fn topk_free_effort(
...
+    value: *mut *mut c_void,
+) -> i32 {
+    if !configs::TOPK_DEFRAG.load(Ordering::Relaxed) {
+        return 0;
+    }
+    // Relocate the sketch's internal heap allocations.
+    let topk_object: &mut TopKObject = &mut *(*value).cast::<TopKObject>();
+    topk_object
+        .sketch_mut()
+        .realloc_large_heap_allocated_objects(&mut Defragger);
```
zackcam, line 127: Nice little gotcha, in `realloc_large_heap_allocated_objects` we call for exmaple` realloc_vec(&mut self.heap, reallocator);`. In this method we do a `into_boxed_slice` where you correctly state we drop spare capacity. However this then could become an issue in memory tracking. In the memory usage we have `item.capacity()` which as we potentially just dropped some would make it so we could have the tracking differ if we sudden;y then delete the topk key. In reverse if we then add and go over the capacity (stll under k) we would most likely double the capacity due to rust vec rules and then deleting it could cause underflow

zackcam: Easy way to fix might be redo memory in this stage

detemmienation: Done in https://github.com/valkey-io/valkey-bloom/pull/109/commits/2d2b7dbaa84cbc1a0c1adf807e55c80184db5a60

src/wrapper/topk_callback.rs
```diff
@@ -117,11 +117,22 @@ pub unsafe extern "C" fn topk_defrag(
...
+    // spare Vec capacity (into_boxed_slice), so the memory gauge is reconciled
+    // by the before/after delta to keep it balanced against Drop's subtraction.
     let topk_object: &mut TopKObject = &mut *(*value).cast::<TopKObject>();
+    let mem_before = topk_object.memory_usage();
     topk_object
         .sketch_mut()
         .realloc_large_heap_allocated_objects(&mut Defragger);
+    let mem_after = topk_object.memory_usage();
+    if mem_before > mem_after {
+        metrics::TOPK_OBJECT_TOTAL_MEMORY_BYTES
```
zackcam: Can we confine all memory work in utils.rs will make it easier for tracking

src/wrapper/topk_callback.rs
```diff
@@ -117,11 +117,22 @@ pub unsafe extern "C" fn topk_defrag(
...
+            .fetch_sub(mem_before - mem_after, Ordering::Relaxed);
+    } else {
+        metrics::TOPK_OBJECT_TOTAL_MEMORY_BYTES
```
zackcam: Defrag can only ever decrease the memory its only the resulting add that could cause an increase. As we trim that will only ever lower or stay the same

src/topk/utils.rs
```diff
@@ -115,6 +115,27 @@ impl TopKObject {
...
+        (std::mem::size_of::<TopKObject>() as u64)
+            .saturating_add(width.saturating_mul(16))
+            .saturating_add(heavy)
+            .saturating_add(1024 * 8)
+            .saturating_add(k.saturating_mul(128))
+    }
+
+    /// Whether these params fit within the configured topk-memory-usage-limit.
+    pub fn validate_size(k: u32, width: u32, depth: u32) -> bool {
+        let limit = configs::TOPK_MEMORY_LIMIT_PER_OBJECT.load(Ordering::Relaxed) as u64;
```
zackcam: size_limit? instead of limit

src/topk/utils.rs
```diff
@@ -9,11 +10,9 @@ pub const DEFAULT_WIDTH: u32 = 8;
...
 pub const DEFAULT_DECAY: f64 = 0.9;
 
-/// Per-argument bounds.
-/// The minimums are 1,The maximums are placeholders covering the
-/// full u32 range until topk-specific configs are introduced; tighten them
-/// once dedicated config knobs land.
-// TODO: replace these with configurable bounds once topk configs exist.
+/// Per-argument bounds. The minimums are 1; the maximums span the full u32
+/// range because oversized sketches are bounded by the topk-memory-usage-limit
+/// config (see `TopKObject::estimated_size`), not by per-argument caps.
```
zackcam: Probably don't explicitly need this comment as names should be descriptive enough

src/topk/utils.rs
```diff
@@ -115,6 +115,27 @@ impl TopKObject {
...
+    /// Bytes the sketch allocates up front: `width` + `width * depth` cells
+    /// (16 bytes each), a 1024-entry u64 decay table, and the priority queue
+    /// reserved to capacity `k` (~128 bytes per entry, a conservative estimate).
+    pub fn estimated_size(k: u32, width: u32, depth: u32) -> u64 {
+        let (k, width, depth) = (k as u64, width as u64, depth as u64);
+        // Saturate: products can overflow u64 at u32::MAX, and wrapping would
+        // let an oversized sketch slip under the limit.
+        let heavy = width.saturating_mul(depth).saturating_mul(16);
+        (std::mem::size_of::<TopKObject>() as u64)
+            .saturating_add(width.saturating_mul(16))
```
zackcam: For this instead of numbers which might be hard to comprehend. Can we add a comment on each line eg:
`.saturating_add(width.saturating_mul(16)) //lobby cells: width cells × 16 bytes`

src/wrapper/topk_callback.rs
```diff
@@ -75,3 +80,67 @@ pub unsafe extern "C" fn topk_free_effort(
...
+        .realloc_large_heap_allocated_objects(&mut Defragger);
+    let mem_after = topk_object.memory_usage();
+    if mem_before > mem_after {
+        metrics::TOPK_OBJECT_TOTAL_MEMORY_BYTES
+            .fetch_sub(mem_before - mem_after, Ordering::Relaxed);
+    } else {
+        metrics::TOPK_OBJECT_TOTAL_MEMORY_BYTES
+            .fetch_add(mem_after - mem_before, Ordering::Relaxed);
+    }
+    // Relocate the TopKObject allocation itself.
```
zackcam: Can we change relocate to defrag/attempt to defrag same as above

src/wrapper/topk_callback.rs
detemmienation: Done in https://github.com/valkey-io/valkey-bloom/pull/109/commits/522372d143320114b19c2e02d13b99205cc74ea4

zackcam (commented):
I think now we can undercount as when we do the trim. If we then add an item the vecs will have to grow which we would not count again. So would have slightly less memory used in this case. However I don't think we'd want to recalculate this every add. This might need some different ideas on how this could be made fully accurate again

zackcam (commented):
Lets add a note with the overcount for now and say when we expand we use the same size before. Once done I will take one final look through it all. Once in the end we may take a look at introducing the unsafe code in heavykeeper crate but for now I think this is the best bet we have for recreation

zackcam (approved):
Approving and merging now. I think if we have time we should revisit the overcount in the future still

PR #110, opened 2026-07-09
Match bloom memory tracking by small over count of bytes after a defrag operation

zackcam:
This commit does a quick reallocation of size after a defrag on the vecs in bloom. This is needed due to dropping excess capacity which was initially tracked by memory metric. As we were silently dropping this we would have a very slight overcount in memory used by bloom in certain cases.

### DIFF
--- a/src/wrapper/bloom_callback.rs
+++ b/src/wrapper/bloom_callback.rs
@@ -229,7 +229,7 @@ pub unsafe extern "C" fn bloom_defrag(
     let bloom_object: &mut BloomObject = &mut *(*value).cast::<BloomObject>();
 
     let num_filters = bloom_object.num_filters();
-    let filters_capacity = bloom_object.filters().capacity();
+    let predefrag_filters_capacity = bloom_object.filters().capacity();
 
     // While we are within a timeframe decided from should_stop_defrag and not over the number of filters defrag the next filter
     while !defrag.should_stop_defrag() && cursor < num_filters as u64 {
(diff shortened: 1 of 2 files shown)


src/wrapper/bloom_callback.rs
```diff
@@ -229,6 +229,7 @@ pub unsafe extern "C" fn bloom_defrag(
     let bloom_object: &mut BloomObject = &mut *(*value).cast::<BloomObject>();
 
     let num_filters = bloom_object.num_filters();
+    let filters_capacity = bloom_object.filters().capacity();
```
KarthikSubbarao: maybe something like "predefrag_filters_capacity"

src/wrapper/bloom_callback.rs
```diff
@@ -314,6 +315,12 @@ pub unsafe extern "C" fn bloom_defrag(
             )
         };
     }
+    // As we drop some capcity we need to change the memory usage to avoid showing a higher memory usage than what is really used.
+    // This will be the difference in capacity and number of filters only * size of the box of bloom filters.
+    metrics::BLOOM_OBJECT_TOTAL_MEMORY_BYTES.fetch_sub(
+        (filters_capacity - num_filters) * std::mem::size_of::<Box<BloomFilter>>(),
+        std::sync::atomic::Ordering::Relaxed,
+    );
```
KarthikSubbarao: In the current implementation, for maintainability, we contained the `BLOOM_OBJECT_TOTAL_MEMORY_BYTES` metric to just the structs in the src/bloom/utils.rs] file. Is it possible to have this fit in there?

KarthikSubbarao (approved):

PR #111, opened 2026-07-13
Count Min Sketch Module: Implement CMS.INCRBY and CMS.QUERY

zmccoy:
This implements the increment by and query methods for the count min sketch and allows us to write the integration tests which test the creation of the CMS along with incrementing and querying.  

Please let me know what may need to change.  Thank you!

### DIFF
--- a/src/cms/cms_command_handler.rs
+++ b/src/cms/cms_command_handler.rs
@@ -179,10 +179,12 @@ pub fn cms_increment_by(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult 
     }
 
     let mut i = 2;
-    let mut pairs: Vec<(&[u8], &ValkeyString)> = Vec::new();
+    let mut pairs: Vec<(&[u8], u64)> = Vec::new();
     while i < args_count {
         let k = args[i].as_slice();
-        let v = &args[i + 1];
+        let v = args[i + 1]
+            .parse_unsigned_integer()
+            .map_err(|_| ValkeyError::Str(utils::BAD_INCREMENT))?;
         pairs.push((k, v));
         i += 2
     }
@@ -198,13 +200,7 @@ pub fn cms_increment_by(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult 
         None => Err(ValkeyError::nonexistent_key()),
         Some(v) => {
             for (item, increment) in pairs {
-                let item = &item;
-                let parsed_value = &increment.to_string_lossy().parse::<u64>();
-                let value = match parsed_value {
-                    Ok(v) => v,
-                    Err(_) => return Err(ValkeyError::Str(utils::BAD_INCREMENT)),
-                };
-                let count = v.increment_by(item, value.to_owned());
+                let count = v.increment_by(item, increment);
                 results.push(ValkeyValue::Integer(count as i64));
             }
             replicate_and_notify_events(ctx, key, Operation::Increment);
--- a/src/cms/cms_command_handler.rs
+++ b/src/cms/cms_command_handler.rs
@@ -179,9 +179,9 @@ pub fn cms_increment_by(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult 
     }
 
     let mut i = 2;
-    let mut pairs: Vec<(&ValkeyString, &ValkeyString)> = Vec::new();
+    let mut pairs: Vec<(&[u8], &ValkeyString)> = Vec::new();
     while i < args_count {
-        let k = &args[i];
+        let k = args[i].as_slice();
         let v = &args[i + 1];
         pairs.push((k, v));
         i += 2
@@ -198,7 +198,7 @@ pub fn cms_increment_by(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult 
         None => Err(ValkeyError::nonexistent_key()),
         Some(v) => {
             for (item, increment) in pairs {
-                let item = &item.to_string_lossy();
+                let item = &item;
                 let parsed_value = &increment.to_string_lossy().parse::<u64>();
                 let value = match parsed_value {
                     Ok(v) => v,
@@ -234,7 +234,7 @@ pub fn cms_query(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
             let estimates: Vec<ValkeyValue> = args[2..]
                 .iter()
                 .map(|item| {
-                    let estimate = v.estimate(&item.to_string_lossy());
+                    let estimate = v.estimate(&item);
                     ValkeyValue::Integer(estimate as i64)
                 })
                 .collect();
--- a/src/cms/cms_command_handler.rs
+++ b/src/cms/cms_command_handler.rs
@@ -1,4 +1,6 @@
-use valkey_module::{Context, NotifyEvent, ValkeyError, ValkeyResult, ValkeyString, VALKEY_OK};
+use valkey_module::{
+    Context, NotifyEvent, ValkeyError, ValkeyResult, ValkeyString, ValkeyValue, VALKEY_OK,
+};
 
 use crate::cms::data_type::CMS_TYPE;
 use crate::cms::utils::{self, CMSObject};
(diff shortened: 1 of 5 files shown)


src/cms/cms_command_handler.rs
```diff
@@ -0,0 +1,244 @@
...
+                let width_val = ValkeyString::create_from_slice(
+                    std::ptr::null_mut(),
+                    width.to_string().as_bytes(),
+                );
+                let depth_val = ValkeyString::create_from_slice(
+                    std::ptr::null_mut(),
+                    depth.to_string().as_bytes(),
+                );
+                let cmd = vec![&width_val, &depth_val];
+                ctx.replicate("CMS.INITBYDIM", cmd.as_slice());
```
zackcam, line 31: I think we are missing key here for the replicate. This would call `CMS.INITBYDIM width depth` which is misisng the key

src/cms/cms_command_handler.rs
```diff
@@ -0,0 +1,244 @@
...
+    let key_name = &args[1];
+
+    let existing_key = ctx.open_key(key_name);
+    let cms = match existing_key.get_value::<CMSObject>(&CMS_TYPE) {
+        Ok(v) => v,
+        Err(_) => return Err(ValkeyError::WrongType),
+    };
+
+    match cms {
+        None => Err(ValkeyError::Str("No CMS Exists")),
```
zackcam: Lets use the same error for cms_query and cms_incrby when no key exists. Currently one uses nonexistent_key while this is a string with "No CMS Exists"

src/cms/cms_command_handler.rs
```diff
@@ -0,0 +1,244 @@
...
+        Ok(v) => v,
+        Err(_) => return Err(ValkeyError::WrongType),
+    };
+
+    let mut results = Vec::new();
+    match value {
+        None => Err(ValkeyError::nonexistent_key()),
+        Some(v) => {
+            for (item, increment) in pairs {
+                let key = &item.to_string_lossy();
```
zackcam: This should be item not key i think for better readability

zackcam (commented):
Quick little review with some minor comments let me know if these don't make sense

src/cms/cms_command_handler.rs
zmccoy: Ah yep missed that in both of the cases.  Got that added now in the recent commit! [8274701](https://github.com/valkey-io/valkey-bloom/pull/111/commits/8274701f062b78642c74b82436145b4794f05fe5)

src/cms/cms_command_handler.rs
zmccoy: This should be fixed in [8274701](https://github.com/valkey-io/valkey-bloom/pull/111/commits/8274701f062b78642c74b82436145b4794f05fe5)

src/cms/cms_command_handler.rs
zmccoy: This should be fixed in [8274701](https://github.com/valkey-io/valkey-bloom/pull/111/commits/8274701f062b78642c74b82436145b4794f05fe5)

zmccoy:
@zackcam All of those make sense and are fixed up.  Thanks!

zmccoy:
This has been rebased after the previous merge and should be good to go into the feature branch.

tests/test_cms_command.py
```diff
@@ -41,6 +43,13 @@ def test_cms_command_error(self):
             ('CMS.INITBYPROB key 0.01', 'wrong number of arguments for \'CMS.INITBYPROB\' command'),
             ('CMS.INITBYPROB key 0.01 0.01 extra', 'wrong number of arguments for \'CMS.INITBYPROB\' command'),
 
+            ('CMS.INCRBY', 'wrong number of arguments for \'CMS.INCRBY\' command'),
```
zackcam: I wonder if we want to make this be " instead of ' just for not having to escape the ' in it

src/cms/utils.rs
```diff
@@ -100,6 +101,15 @@ impl CMSObject {
     pub fn total(&self) -> u64 {
         self.total
     }
+
+    pub fn increment_by(&mut self, item: &String, increment: u64) -> u64 {
```
zackcam: For this do we need to parse the output first so we don't fail a incrby part way through? This would cause inconsistency with replication if we don't copy what was done either

src/cms/utils.rs
```diff
@@ -100,6 +101,15 @@ impl CMSObject {
     pub fn total(&self) -> u64 {
         self.total
```
zackcam: Do we increment this anywhere yet? If nnto fine just want to make sure its know if so.

src/cms/cms_command_handler.rs
```diff
@@ -149,11 +150,95 @@ pub fn cms_initialize_by_probability(ctx: &Context, args: Vec<ValkeyString>) ->
...
+        Ok(v) => v,
+        Err(_) => return Err(ValkeyError::WrongType),
+    };
+
+    let mut results = Vec::new();
+    match value {
+        None => Err(ValkeyError::nonexistent_key()),
+        Some(v) => {
+            for (item, increment) in pairs {
+                let item = &item.to_string_lossy();
```
zackcam: Should we do lossy here? Could we pass `args[i].as_slice()` instead

src/cms/utils.rs
zmccoy: Can you explain what you're thinking there?  I'm a little fuzzy on what we may be touching on.

zackcam: Lets say we do a `CMS.INCRBY key a 10 b 42 c notanumber` should we reject this command before we do any work of actually adding anything as incrementing c can't be done properly? I think currently we do the a and b increment then check c and fail

src/cms/utils.rs
zmccoy: We don't, I can change it here as we're using the internal total method in the CMS struc itself. It was in another PR.

src/cms/cms_command_handler.rs
zmccoy: We can do that, would that be preferred in keeping all things in tact for all of these valkey strings that we take in?

zmccoy: Fixed in [Address comments, use as_slice instead of valkey string lossy, fix up…](https://github.com/valkey-io/valkey-bloom/pull/111/commits/47e043f83323b2d0c234494c603ed2b465bb428b) and we can do the same to others if you think that's the right choice.

src/cms/utils.rs
zmccoy: Fixed in [Address comments, use as_slice instead of valkey string lossy, fix up…](https://github.com/valkey-io/valkey-bloom/pull/111/commits/47e043f83323b2d0c234494c603ed2b465bb428b)

tests/test_cms_command.py
zmccoy: Fixed in [Address comments, use as_slice instead of valkey string lossy, fix up…](https://github.com/valkey-io/valkey-bloom/pull/111/commits/47e043f83323b2d0c234494c603ed2b465bb428b)

zackcam: Awesome thank you this looks resolved to me now!

src/cms/cms_command_handler.rs
zackcam: Awesome thank you this looks resolved to me now!

src/cms/utils.rs
zmccoy: Ah yes that makes total sense.  I've got that fixed now in [Parse arguments before doing any state manipulation](https://github.com/valkey-io/valkey-bloom/pull/111/commits/e909ac608506c5396b4d76f45a40564b05bb4654)

zackcam (approved):
Looks good to me! Thanks for doing the comments

src/cms/cms_command_handler.rs
KarthikSubbarao: Would be good to add a replication test. It would have caught this. But also, helpful to add a test for future regressions. You can refer to Bloom Replication Tests

src/cms/utils.rs
```diff
@@ -98,7 +92,16 @@ impl CMSObject {
     }
 
     pub fn total(&self) -> u64 {
-        self.total
+        self.cms.sketch.total_count()
```
KarthikSubbarao, line 95: Nice! The raw library supports it

src/cms/cms_command_handler.rs
```diff
@@ -149,11 +150,91 @@ pub fn cms_initialize_by_probability(ctx: &Context, args: Vec<ValkeyString>) ->
...
+        i += 2
+    }
+
+    let filter_key = ctx.open_key_writable(key);
+    let value = match filter_key.get_value::<CMSObject>(&CMS_TYPE) {
+        Ok(v) => v,
+        Err(_) => return Err(ValkeyError::WrongType),
+    };
+
+    let mut results = Vec::new();
```
KarthikSubbarao, line 198: `with_capacity` is nice to use if we know the size ahead of time, to avoid expansion

src/cms/utils.rs
```diff
@@ -98,7 +92,16 @@ impl CMSObject {
...
+    }
+
+    pub fn increment_by(&mut self, item: &[u8], increment: u64) -> u64 {
```
KarthikSubbarao, line 98: A small comment explaining what the APIs of the `CMSObject` do and return will be helpful

KarthikSubbarao (approved):
Approving with comments to address in follow up

PR #112, opened 2026-07-15
Add TopK AOF rewrite support (TOPK.LOAD)

detemmienation:
## Summary

Adds AOF rewrite support for the TopK data type. During an AOF rewrite, each TopK object is serialized into a single `TOPK.LOAD <key> <blob>` command; replaying that command on AOF load reconstructs a byte-identical sketch. 

Because a TopK sketch's `DEBUG DIGEST-VALUE` is computed over the exact serialized bytes (`sketch.to_bytes()`, which includes RNG state and cuckoo layout), replaying `TOPK.RESERVE` + `TOPK.ADD` cannot reproduce it. The rewrite therefore carries the full serialized blob rather than replaying the mutating commands.

### Shared encode/decode
- Extract `TopKObject::{encode_object, decode_object, from_serialized_bytes}`, factoring the seed/num_items/sketch (de)serialization and parameter validation out of the RDB load path so both the RDB and the AOF/`TOPK.LOAD` paths share one implementation.
- The object blob is `[TOPK_OBJECT_VERSION | seed | num_items | sketch bytes]`. No on-disk RDB format change.

### AOF rewrite callback
- Wire up the `aof_rewrite` type callback.
- `topk_aof_rewrite` serializes the object via `encode_object` and emits a single `TOPK.LOAD` command.

### TOPK.LOAD command
- Add `TOPK.LOAD <key> <blob>`, which reconstructs a TopK object via `decode_object`. It rejects if the key already exists and skips the per-object memory-limit check for primary/AOF clients, since an already-persisted object must load.
- Replicates verbatim: the blob already carries the full sketch state, so the command cannot be decomposed into `TOPK.RESERVE` replay. Publishes the `topk.reserve` keyspace event on the replica since it creates the object. Registered as a public command with an internal-use intent.

### Tests
- `test_topk_aofrewrite.py`: both tests disable `aof-use-rdb-preamble` so the rewrite actually emits `TOPK.LOAD`. With the default preamble the AOF base file is written in RDB format and persistence goes through `rdb_save`/`rdb_load`, so the new path would not be covered. Adds a basic `bgrewriteaof` + restart digest round-trip, and a metrics test asserting the `INFO bf` gauges survive the rewrite and reset to zero on `DEL`.
- Unit tests for the encode/decode round trip and the truncated, unsupported-version, and wrong-seed rejection paths.
- `test_bloom_aofrewrite.py`: the existing bloom AOF tests ran with the default preamble, so they went through the RDB path and never exercised `bloom_aof_rewrite` / `BF.LOAD`. Disable the preamble there too so they cover the intended command path.

### DIFF
--- a/src/topk/command_handler.rs
+++ b/src/topk/command_handler.rs
@@ -14,53 +14,60 @@ struct ReplicateArgs {
     seed: u64,
 }
 
+/// The write operation being propagated to replicas and published as a keyspace event.
+/// Each variant controls which command is propagated and which event is emitted.
+enum ReplicationOp {
+    /// TOPK.RESERVE: replayed as a deterministic TOPK.RESERVE on replicas.
+    Reserve(ReplicateArgs),
+    /// TOPK.ADD / TOPK.INCRBY: replicated verbatim and publishes `topk.add`,
+    /// safe because they are deterministic once the replica is seeded
+    /// identically via the replicated TOPK.RESERVE.
+    Add,
+    /// TOPK.LOAD (generated during AOF rewrite): replicated verbatim since the
+    /// serialized blob already carries the full sketch, and publishes the
+    /// `topk.reserve` event because it creates the object on the replica.
+    Load,
+}
+
 /// Helper function to replicate mutative commands to the replica nodes and publish keyspace events.
-/// There are three cases for replication:
-/// - RESERVE operation: replays a deterministic TOPK.RESERVE on replicas.
-/// - ADD operation: covers TOPK.ADD and TOPK.INCRBY; both replicate verbatim and publish
-///   the same `topk.add` event, safe because they are deterministic once the
-///   replica is seeded identically via the replicated TOPK.RESERVE.
-/// - LOAD operation: TOPK.LOAD (generated during AOF rewrite). Replicates verbatim
-///   since the serialized blob already carries the full sketch, and publishes the
-///   `topk.reserve` event because it creates the object on the replica.
-fn replicate_and_notify_events(
-    ctx: &Context,
-    key_name: &ValkeyString,
-    reserve_operation: bool,
-    add_operation: bool,
-    load_operation: bool,
-    args: Option<ReplicateArgs>,
-) {
-    if reserve_operation {
-        let args = args.expect("reserve replication requires ReplicateArgs");
-        let k_val =
-            ValkeyString::create_from_slice(std::ptr::null_mut(), args.k.to_string().as_bytes());
-        let width_val = ValkeyString::create_from_slice(
-            std::ptr::null_mut(),
-            args.width.to_string().as_bytes(),
-        );
-        let depth_val = ValkeyString::create_from_slice(
-            std::ptr::null_mut(),
-            args.depth.to_string().as_bytes(),
-        );
-        let decay_val = ValkeyString::create_from_slice(
-            std::ptr::null_mut(),
-            args.decay.to_string().as_bytes(),
-        );
-        let seed_str = ValkeyString::create_from_slice(std::ptr::null_mut(), "SEED".as_bytes());
-        let seed_val =
-            ValkeyString::create_from_slice(std::ptr::null_mut(), args.seed.to_string().as_bytes());
-        let cmd = vec![
-            key_name, &k_val, &width_val, &depth_val, &decay_val, &seed_str, &seed_val,
-        ];
-        ctx.replicate("TOPK.RESERVE", cmd.as_slice());
-        ctx.notify_keyspace_event(NotifyEvent::GENERIC, utils::RESERVE_EVENT, key_name);
-    } else if add_operation {
-        ctx.replicate_verbatim();
-        ctx.notify_keyspace_event(NotifyEvent::GENERIC, utils::ADD_EVENT, key_name);
-    } else if load_operation {
-        ctx.replicate_verbatim();
-        ctx.notify_keyspace_event(NotifyEvent::GENERIC, utils::RESERVE_EVENT, key_name);
+fn replicate_and_notify_events(ctx: &Context, key_name: &ValkeyString, op: ReplicationOp) {
+    match op {
+        ReplicationOp::Reserve(args) => {
+            let k_val = ValkeyString::create_from_slice(
+                std::ptr::null_mut(),
+                args.k.to_string().as_bytes(),
+            );
+            let width_val = ValkeyString::create_from_slice(
+                std::ptr::null_mut(),
+                args.width.to_string().as_bytes(),
+            );
+            let depth_val = ValkeyString::create_from_slice(
+                std::ptr::null_mut(),
+                args.depth.to_string().as_bytes(),
+            );
+            let decay_val = ValkeyString::create_from_slice(
+                std::ptr::null_mut(),
+                args.decay.to_string().as_bytes(),
+            );
+            let seed_str = ValkeyString::create_from_slice(std::ptr::null_mut(), "SEED".as_bytes());
+            let seed_val = ValkeyString::create_from_slice(
+                std::ptr::null_mut(),
+                args.seed.to_string().as_bytes(),
+            );
+            let cmd = vec![
+                key_name, &k_val, &width_val, &depth_val, &decay_val, &seed_str, &seed_val,
+            ];
+            ctx.replicate("TOPK.RESERVE", cmd.as_slice());
+            ctx.notify_keyspace_event(NotifyEvent::GENERIC, utils::RESERVE_EVENT, key_name);
+        }
+        ReplicationOp::Add => {
+            ctx.replicate_verbatim();
+            ctx.notify_keyspace_event(NotifyEvent::GENERIC, utils::ADD_EVENT, key_name);
+        }
+        ReplicationOp::Load => {
+            ctx.replicate_verbatim();
+            ctx.notify_keyspace_event(NotifyEvent::GENERIC, utils::RESERVE_EVENT, key_name);
+        }
     }
 }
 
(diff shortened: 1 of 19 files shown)


src/topk/command_handler.rs
```diff
@@ -14,16 +15,20 @@ struct ReplicateArgs {
...
 ///   replica is seeded identically via the replicated TOPK.RESERVE.
+/// - LOAD operation: TOPK.LOAD (generated during AOF rewrite). Replicates verbatim
+///   since the serialized blob already carries the full sketch, and publishes the
+///   `topk.reserve` event because it creates the object on the replica.
 fn replicate_and_notify_events(
     ctx: &Context,
     key_name: &ValkeyString,
     reserve_operation: bool,
     add_operation: bool,
+    load_operation: bool,
```
zackcam: Could maybe do an enum here instead to make it easier and less unwieldily for the different operations

detemmienation: Done in commit https://github.com/valkey-io/valkey-bloom/pull/112/commits/74ecdc71efe73d97326f6d8e80390ebdd5f1cde2 🫡

src/topk/utils.rs
```diff
@@ -431,4 +504,80 @@ mod tests {
...
+        let topk = TopKObject::new_reserved(3, 16, 4, 0.9, 42);
+        let blob = topk.encode_object();
+        let decoded = TopKObject::decode_object(&blob, false).expect("round trip should succeed");
+        assert_eq!(decoded.num_items(), 0);
+        assert_eq!(decoded.sketch().to_bytes(), topk.sketch().to_bytes());
+    }
+
+    #[test]
+    fn test_topk_decode_when_bytes_is_truncated_should_failed() {
+        // A blob shorter than the 17-byte header is rejected.
```
Frank-Gu-81: nit: minor grammar fix - `*_should_fail`. Same for the 2 tests below.

src/topk/utils.rs
```diff
@@ -164,6 +167,76 @@ impl TopKObject {
...
+        }
+        if blob[0] != TOPK_OBJECT_VERSION {
+            return Err(DECODE_UNSUPPORTED_VERSION);
+        }
+        let seed = u64::from_le_bytes(blob[1..9].try_into().expect("8 bytes"));
+        let num_items = u64::from_le_bytes(blob[9..17].try_into().expect("8 bytes"));
+        let sketch = CuckooTopK::<Vec<u8>>::from_bytes(&blob[17..], seed)
+            .map_err(|_| DECODE_TOPK_OBJECT_FAILED)?;
+        Self::from_serialized_bytes(seed, num_items, sketch, validate_size_limit)
+    }
```
Frank-Gu-81, line 279: My concern here is that it looks like `TOPK_K_MAX`/`TOPK_WIDTH_MAX`/`TOPK_DEPTH_MAX` are `u32::MAX`, so the range checks in `from_serialized_bytes` are no-ops and the only real cap is `validate_size`, which runs after `CuckooTopK::from_bytes` has already allocated the sketch. 

On the RDB path that's fine because the blob came from disk. But for `TOPK.LOAD`, it is user callable. 

We should 1. reorder so that validation happens before sketch is build and 2. fix the bound

tests/test_topk_aofrewrite.py
```diff
@@ -0,0 +1,93 @@
+from valkeytestframework.util.waiters import *
```
Frank-Gu-81: Nit: Could you take a look and see if we actually need to import *? Normally, we should try to avoid importing everything.

tests/test_topk_aofrewrite.py
Frank-Gu-81, line 1: Nice coverage on decode_object in `src/topk/utils.rs`! Can we also add a few test around checking `KEY_EXISTS`, `WRONGTYPE`, and error-message wire format here as well?

Frank-Gu-81 (changes requested):

src/topk/utils.rs
detemmienation: Fixed in commit https://github.com/valkey-io/valkey-bloom/pull/112/commits/903bbeaf93b6d7912cdff48fe6f03a2697541710 and also fixed the same issue from bloom's tests

src/topk/utils.rs
detemmienation: Addressed both in commit https://github.com/valkey-io/valkey-bloom/pull/112/commits/d411ed8ea6e67bb94c0c00642d8d1e8cd30f31a0: reordered `decode_object` to peek the sketch header and run `validate_size` before from_bytes allocates, but only on the user-callable TOPK.LOAD path (RDB stays unchecked so a tightened limit can't reject persisted objects). 

Left `TOPK_*_MAX at u32::MAX` since the real cap is `validate_size` (topk-memory-usage-limit, 128MB default) same as TOPK.RESERVE. (Bloom similarly relies on a per-object memory limit as its size cap.)

Added `test_topk_decode_when_size_exceeds_limit_should_fail` covering the pass case plus oversized top_items and width/depth rejections.

tests/test_topk_aofrewrite.py
detemmienation: Thanks!! Added TOPK.LOAD error-path coverage to `test_topk_command.py` in commit https://github.com/valkey-io/valkey-bloom/pull/112/commits/5e2719364ee23dc7a6fbd88b943a8208572ed35e 

One question though: TOPK.LOAD isn't really a user-facing command (it only exists so AOF rewrite can replay sketches, same as BF.LOAD), so in normal operation nothing hits these error paths. Do we consider this level of coverage necessary? Asking because BF.LOAD currently has no equivalent tests, should we also add some?

Frank-Gu-81: Thanks! Fair point. The reason I suggested it was: `TOPK.LOAD` is registered as "write deny-oom" with no flag restricting callers, so technically any authenticated client can invoke it whether we intended that or not. But if bloom doesn't have it, I'm okay with dropping them for consistency. 

Separately, I think it is worth deciding whether TOPK.LOAD should be user-callable at all. If we restrict it, most of this coverage becomes optional. This is something we can discuss with @zackcam next week.

Frank-Gu-81 (approved):

tests/test_topk_aofrewrite.py
detemmienation: Bloom's tests and all our previous tests use the same import * here, if we wanna change it, we may want to change them all together so they dont diverge. @zackcam what do you think?

src/topk/utils.rs
```diff
@@ -164,6 +167,97 @@ impl TopKObject {
...
+
+        let sketch_blob = &blob[17..];
+
+        // Pre-validate: peek width/depth/top_items from the sketch header.
+        if validate_size_limit {
+            const SKETCH_PARAMS_END: usize = 4 + 1 + 1 + 8 + 8 + 8 + 8 + 8; // 46
+            if sketch_blob.len() < SKETCH_PARAMS_END {
+                return Err(DECODE_TOPK_OBJECT_FAILED);
+            }
+            let width = u64::from_le_bytes(sketch_blob[14..22].try_into().expect("8 bytes"));
```
zackcam: Lets make a note on this for these offsets coming from the heavykeeper crate and where they are in the heavykeeper crate

src/topk/utils.rs
```diff
@@ -164,6 +167,97 @@ impl TopKObject {
...
+            let depth = u64::from_le_bytes(sketch_blob[22..30].try_into().expect("8 bytes"));
```
zackcam: Also for these maybe we should make a variable for these to make it more understanble

tests/test_topk_aofrewrite.py
zackcam: Yes I don't see any reason not to do this if its quick and easy. We only ever use wait_for_equal at the moment tbh as is so might be easy as replace all with just that import

tests/test_topk_aofrewrite.py
zackcam: TOPK.LOAD should stay user-callable. The fact that a authenticated client is needed means that the client could already mess with topk or other parts of the server.

detemmienation: Thanks @zackcam! Bloom's BF.LOAD has no error-path test coverage like the tests I added for topk. Since we're keeping them user-callable, do you think it's worth keeping the current tests and adding matching ones for BF.LOAD, or should we drop them all?

zackcam: I think its fine to keep the tests we have, we can decide later and not something that you should focus on for doing the equivalent bloom tests for now

tests/test_topk_aofrewrite.py
detemmienation: Done in https://github.com/valkey-io/valkey-bloom/pull/112/commits/4767e565ba5fc872f07b8e1ac0f393c8596b59a6

src/topk/utils.rs
detemmienation: Done in commit https://github.com/valkey-io/valkey-bloom/pull/112/commits/807616af76a9d3e2d08eabf73d10b1dea702e280

tests/test_topk_aofrewrite.py
```diff
@@ -0,0 +1,94 @@
...
+        # cmd debug digest
+        server_digest = client.execute_command('DEBUG', 'DIGEST')
+        assert server_digest != None or 0000000000000000000000000000000000000000
+        object_digest = client.execute_command('DEBUG DIGEST-VALUE testSave')
+
+        # save aof, restart server
+        client.bgrewriteaof()
+        self.server.wait_for_action_done(ValkeyAction.AOF_REWRITE)
+        # Keep the server running for 1 second more to have a larger uptime.
+        time.sleep(1)
```
zackcam: Is this wait needed?

zackcam (approved):

PR #113, opened 2026-07-23
Add TopK valkey-py compatibility test

detemmienation:
## Summary

Fixes TOPK.INFO field-name casing to match the valkey-py client contract, and adds a valkey-py compatibility test for the TopK data type.

### INFO field-name fix
- Lowercase the `TOPK.INFO` response labels from `K`/`Width`/`Depth`/`Decay`/`Size`/`Total items added` to `k`/`width`/`depth`/`decay`/`size`/`total items added`.
- The single-field selector tokens (`TOPK.INFO key K`) remain case-insensitive, so existing callers are unaffected.
- Update assertions in `test_topk_command.py` and `test_topk_basic.py`.

### Compatibility test
- Add `test_topk_valkeypy_compatibility.py` adapted from [valkey-py](https://github.com/valkey-io/valkey-py/blob/main/tests/test_bloom.py) (`test_topk` and `test_topk_incrby`), exercising all commands(`reserve`/`add`/`incrby`/`query`/`count`/`list`/`info`) through valkey-py's high-level client API.
- Assertions verify protocol-level compatibility (return types, lengths, field-name parsing).

### DIFF
--- a/src/topk/command_handler.rs
+++ b/src/topk/command_handler.rs
@@ -339,17 +339,17 @@ pub fn topk_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
     }
 
     let result = vec![
-        ValkeyValue::SimpleStringStatic("K"),
+        ValkeyValue::SimpleStringStatic("k"),
         ValkeyValue::Integer(topk.k() as i64),
-        ValkeyValue::SimpleStringStatic("Width"),
+        ValkeyValue::SimpleStringStatic("width"),
         ValkeyValue::Integer(topk.width() as i64),
-        ValkeyValue::SimpleStringStatic("Depth"),
+        ValkeyValue::SimpleStringStatic("depth"),
         ValkeyValue::Integer(topk.depth() as i64),
-        ValkeyValue::SimpleStringStatic("Decay"),
+        ValkeyValue::SimpleStringStatic("decay"),
         ValkeyValue::Float(topk.decay()),
-        ValkeyValue::SimpleStringStatic("Size"),
+        ValkeyValue::SimpleStringStatic("size"),
         ValkeyValue::Integer(topk.memory_usage() as i64),
-        ValkeyValue::SimpleStringStatic("Total items added"),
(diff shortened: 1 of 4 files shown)


src/topk/command_handler.rs
```diff
@@ -339,17 +339,17 @@ pub fn topk_info(ctx: &Context, input_args: &[ValkeyString]) -> ValkeyResult {
...
+        ValkeyValue::SimpleStringStatic("width"),
         ValkeyValue::Integer(topk.width() as i64),
-        ValkeyValue::SimpleStringStatic("Depth"),
+        ValkeyValue::SimpleStringStatic("depth"),
         ValkeyValue::Integer(topk.depth() as i64),
-        ValkeyValue::SimpleStringStatic("Decay"),
+        ValkeyValue::SimpleStringStatic("decay"),
         ValkeyValue::Float(topk.decay()),
-        ValkeyValue::SimpleStringStatic("Size"),
+        ValkeyValue::SimpleStringStatic("size"),
```
zackcam, line 350: Size and total are just in ours so technically we don't have to lowercase these but think its good for consistency

tests/test_topk_valkeypy_compatibility.py
```diff
@@ -0,0 +1,71 @@
...
+from valkey_bloom_test_case import SkipSeedParameterizationMixin, ValkeyBloomTestCaseBase
+from valkeytestframework.conftest import resource_port_tracker  # noqa: F401
+
+class TestValkeyTopKCompatibility(SkipSeedParameterizationMixin, ValkeyBloomTestCaseBase):
+    """
+        Tests TopK compatibility with valkey-py's high-level client API.
+        Adapted from https://github.com/valkey-io/valkey-py/blob/main/tests/test_bloom.py
+    """
+
+    def test_topk(self):
```
zackcam, line 11: Makes sense that we don't have the same fixed seed so can't have same exact check thats done in the original tests

zackcam (approved):

PR #114, opened 2026-07-27
Update documentation for TopK commands

detemmienation:
- Update README and QUICK_START to document TopK data type alongside Bloom Filter
- Add command JSON files for all 8 TopK commands (TOPK.RESERVE, TOPK.ADD, TOPK.INCRBY, TOPK.QUERY, TOPK.COUNT, TOPK.LIST, TOPK.INFO, TOPK.LOAD) with time complexities aligned to Redis docs
- Update opK ACL categories and  ACL category test to match the new classifications
- Use u32 fingerprint and counter to reduce memory usage

### DIFF
--- a/QUICK_START.md
+++ b/QUICK_START.md
@@ -1,6 +1,6 @@
 # Quick Start
 
-Follow these steps to set up, build, and run the Valkey server with the valkey-bloom module. This guide covers both Bloom Filter and Top-K data types.
+Follow these steps to set up, build, and run the Valkey server with the valkey-bloom module. This guide will walk you through creating a Bloom Filter and a Top-K sketch, inserting items, and querying them.
 
 ## Step 1: Install Valkey and valkey-bloom
 
--- a/QUICK_START.md
+++ b/QUICK_START.md
@@ -1,6 +1,6 @@
 # Quick Start
 
-Follow these steps to set up, build, and run the Valkey server with the valkey-bloom module. This guide will walk you through creating a bloom filter, inserting items, and checking for items in the filters.
+Follow these steps to set up, build, and run the Valkey server with the valkey-bloom module. This guide covers both Bloom Filter and Top-K data types.
 
 ## Step 1: Install Valkey and valkey-bloom
 
@@ -19,13 +19,15 @@ In case of Linux:
 
 You should see the Valkey server start, and it will be ready to accept commands.
 
-## Step 3: Create a Bloom Filter
-
 Start a Valkey CLI session:
 
 ```bash
 valkey-cli
 ```
+
+## Bloom Filter
+
+### Step 3: Create a Bloom Filter
 
 Create a bloom filter using the BF.ADD, BF.INSERT, BF.RESERVE or BF.MADD commands. For example:
 
@@ -36,7 +38,7 @@ BF.ADD filter-key item-val
 - `filter-key` is the name of the bloom filter we will be operating on
 - `item-val` is the item we are inserting into the bloom filter
 
-## Step 4: Insert some more items
+### Step 4: Insert some more items
 
 To insert items on an already created filter, use the `BF.ADD`, `BF.MADD` or `BF.INSERT` commands:
 
@@ -47,7 +49,7 @@ BF.MADD filter-key example1 example2
 
 Replace the example with the actual items you want to add.
 
-## Step 5: Check if items are present
+### Step 5: Check if items are present
 
 Now that you've created a bloom filter and inserted items, you can check what items have been added. Use the `BF.EXISTS` or `BF.MEXISTS` commands to check for items:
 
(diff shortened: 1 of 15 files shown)


src/lib.rs
```diff
@@ -170,12 +170,12 @@ valkey_module! {
...
         ["TOPK.RESERVE", topk_reserve_command, "write fast deny-oom", 1, 1, 1, "fast write topk"],
-        ["TOPK.ADD", topk_add_command, "write fast deny-oom", 1, 1, 1, "fast write topk"],
-        ["TOPK.INCRBY", topk_incrby_command, "write fast deny-oom", 1, 1, 1, "fast write topk"],
+        ["TOPK.ADD", topk_add_command, "write deny-oom", 1, 1, 1, "write topk"],
+        ["TOPK.INCRBY", topk_incrby_command, "write deny-oom", 1, 1, 1, "write topk"],
         ["TOPK.INFO", topk_info_command, "readonly fast", 1, 1, 1, "fast read topk"],
         ["TOPK.LIST", topk_list_command, "readonly", 1, 1, 1, "read topk"],
-        ["TOPK.COUNT", topk_count_command, "readonly fast", 1, 1, 1, "fast read topk"],
-        ["TOPK.QUERY", topk_query_command, "readonly fast", 1, 1, 1, "fast read topk"],
+        ["TOPK.COUNT", topk_count_command, "readonly", 1, 1, 1, "read topk"],
```
zackcam: Should these be slow now instead?

src/lib.rs
```diff
@@ -170,12 +170,12 @@ valkey_module! {
         ["BF.LOAD", bloom_load_command, "write deny-oom", 1, 1, 1, "write bloom"],
         // TOPK Commands
         ["TOPK.RESERVE", topk_reserve_command, "write fast deny-oom", 1, 1, 1, "fast write topk"],
-        ["TOPK.ADD", topk_add_command, "write fast deny-oom", 1, 1, 1, "fast write topk"],
-        ["TOPK.INCRBY", topk_incrby_command, "write fast deny-oom", 1, 1, 1, "fast write topk"],
+        ["TOPK.ADD", topk_add_command, "write deny-oom", 1, 1, 1, "write topk"],
```
zackcam: ACL and tag may have different definitions might be good to double check tag as well

src/commands/topk.reserve.json
```diff
@@ -0,0 +1,46 @@
...
+                "name": "key",
+                "type": "key",
+                "key_spec_index": 0
+            },
+            {
+                "name": "topk",
+                "type": "integer"
+            },
+            {
+                "name": "width",
```
zackcam: Do you know if this will appear as the optional arguments need to be all done together? i.e we would want `[width depth decay]` not `[width] [depth] [decay]`

src/commands/topk.add.json
```diff
@@ -0,0 +1,25 @@
+{
+    "TOPK.ADD": {
+        "summary": "Adds one or more items to a Top-K sketch",
```
zackcam: Should we mention what gets expelled?

src/commands/topk.count.json
```diff
@@ -0,0 +1,25 @@
+{
+    "TOPK.COUNT": {
+        "summary": "Returns the estimated count for one or more items in a Top-K sketch",
```
zackcam: Should we note the inconsistency that can occur here?

src/commands/topk.incrby.json
```diff
@@ -0,0 +1,35 @@
+{
+    "TOPK.INCRBY": {
+        "summary": "Increases the count of one or more items in a Top-K sketch by the specified increments",
+        "complexity": "O(N * k * incr) where N is the number of items, k is the depth, and incr is the increment",
```
zackcam: Is the complexity really gated on incr? do we just increase bu the number instead of doing an incrmeent 1 at a time

QUICK_START.md
```diff
@@ -1,6 +1,6 @@
 # Quick Start
 
-Follow these steps to set up, build, and run the Valkey server with the valkey-bloom module. This guide will walk you through creating a bloom filter, inserting items, and checking for items in the filters.
+Follow these steps to set up, build, and run the Valkey server with the valkey-bloom module. This guide covers both Bloom Filter and Top-K data types.
```
zackcam: Maybe still have "This guide will walk you through creating" then make it generic after that

src/lib.rs
detemmienation: Fixed in https://github.com/valkey-io/valkey-bloom/pull/114/commits/eb5d4245f3c254f4ca4cfd42c8f20d54b92853f6

src/lib.rs
detemmienation: Fixed in https://github.com/valkey-io/valkey-bloom/pull/114/commits/eb5d4245f3c254f4ca4cfd42c8f20d54b92853f6

src/commands/topk.incrby.json
detemmienation: Incr does change the time, see https://github.com/pmcgleenon/heavykeeper-rs/blob/ced9dc002f1ed528a4368e7ad29d878a1554166f/src/cuckoo.rs#L715 
On the decay path we loop incr times so it is O(incr) here. But cuckoo is not nested, so depth and incr add rather than multiply, the overall time complexity will be like O(n *(depth + incr)), fixed in https://github.com/valkey-io/valkey-bloom/commit/eb5d4245f3c254f4ca4cfd42c8f20d54b92853f6

src/commands/topk.reserve.json
detemmienation: Done in https://github.com/valkey-io/valkey-bloom/pull/114/commits/218bb99ff7a8d31e3362cd1e779385a7440ff5b1

QUICK_START.md
detemmienation: Done in https://github.com/valkey-io/valkey-bloom/pull/114/commits/31b2dc8862c4b2c34cba8acb97b02831e5f2dae6

README.md
```diff
@@ -1,12 +1,18 @@
 # valkey-bloom
 
-Valkey-Bloom (BSD-3-Clause) is a Rust based Valkey-Module which brings a Bloom Filter (Module) data type into Valkey and supports versions >= 8.0. With this, users can create bloom filters (space efficient probabilistic data structures) to add elements, check whether elements exists, auto scale their filters, customize bloom filter properties, perform RDB Save and load operations, etc.
+Valkey-Bloom (BSD-3-Clause) is a Rust based Valkey-Module which brings a Bloom Filter and a Top-K (Module) data type into Valkey and supports versions >= 8.0.
+
+With Bloom Filters, users can create space efficient probabilistic data structures to add elements, check whether elements exist, auto scale their filters, customize bloom filter properties, perform RDB Save and load operations, etc.
+
+With Top-K, users can create probabilistic data structures to track the most frequent items in a data stream. Users can add items (with optional increment weights), query whether items are currently in the top-k set, retrieve estimated counts, list the current top-k items ranked by frequency, customize sketch properties, perform RDB Save and load operations, etc.
```
zackcam: I don't know if the `with optional increment weights` makes sense in this context might be confusing to people I think keeping it simple with add

QUICK_START.md
```diff
@@ -56,3 +58,38 @@ BF.EXISTS filter-key example
 ```
 
 This command checks if an item is present in a bloom filter. Bloom filters can have false positives, but no false negatives. This means that if the BF.EXISTS command returns 0, then the item is not present. But if the BF.EXISTS command returns 1, there is a possibility (determined by false positive rate) that the item is not actually present.
+
+## Top-K
+
+### Step 3: Create a Top-K Sketch
```
zackcam: Maybe should be like step 3a etc same for above. Or Topk should be 4a more precisely

zackcam (approved):

zackcam:
Merging this in. Doing the vend pr as the one we will make changes to if needed

PR #116, opened 2026-08-05
CMS Feature: Add the INFO command 

zmccoy:
Branched from: https://github.com/valkey-io/valkey-bloom/pull/111 

This adds the INFO command to the list of supported commands for CMS.

### DIFF
--- a/src/cms/cms_command_handler.rs
+++ b/src/cms/cms_command_handler.rs
@@ -256,11 +256,11 @@ pub fn cms_info(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     match cms {
         Some(cms) => {
             let result = vec![
-                ValkeyValue::SimpleStringStatic("Width"),
+                ValkeyValue::SimpleStringStatic("width"),
                 ValkeyValue::Integer(cms.width as i64),
-                ValkeyValue::SimpleStringStatic("Depth"),
+                ValkeyValue::SimpleStringStatic("depth"),
                 ValkeyValue::Integer(cms.depth as i64),
-                ValkeyValue::SimpleStringStatic("Count"),
+                ValkeyValue::SimpleStringStatic("count"),
                 ValkeyValue::Integer(cms.total() as i64),
             ];
             Ok(ValkeyValue::Array(result))
--- a/src/cms/cms_command_handler.rs
+++ b/src/cms/cms_command_handler.rs
@@ -242,7 +242,8 @@ pub fn cms_query(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
 //Function that implements logic to handle the CMS.INFO command.
 pub fn cms_info(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     let args_count = args.len();
-    if args_count != 2 {
+
+    if !(2..=3).contains(&args_count) {
         return Err(valkey_module::ValkeyError::WrongArity);
     }
 
@@ -254,7 +255,7 @@ pub fn cms_info(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     };
 
     match cms {
-        Some(cms) => {
+        Some(cms) if args_count == 2 => {
             let result = vec![
                 ValkeyValue::SimpleStringStatic("width"),
                 ValkeyValue::Integer(cms.width as i64),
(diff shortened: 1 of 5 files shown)


src/cms/utils.rs
zackcam: With width and depth now being public do we need these getters?

src/cms/cms_command_handler.rs
```diff
@@ -242,3 +242,33 @@ pub fn cms_query(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
...
+    let key_existing = ctx.open_key(key);
+    let cms = match key_existing.get_value::<CMSObject>(&CMS_TYPE) {
+        Ok(v) => v,
+        Err(_) => return Err(ValkeyError::WrongType),
+    };
+
+    match cms {
+        Some(cms) => {
+            let result = vec![
+                ValkeyValue::SimpleStringStatic("Width"),
```
zackcam: I think these should be fully lower case to maintain compatibility with redis, was looking at example in https://redis.io/docs/latest/commands/cms.info/ for the fields returned

src/cms/cms_command_handler.rs
```diff
@@ -242,3 +242,33 @@ pub fn cms_query(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
...
+                ValkeyValue::Integer(cms.width as i64),
+                ValkeyValue::SimpleStringStatic("Depth"),
+                ValkeyValue::Integer(cms.depth as i64),
+                ValkeyValue::SimpleStringStatic("Count"),
+                ValkeyValue::Integer(cms.total() as i64),
+            ];
+            Ok(ValkeyValue::Array(result))
+        }
+        None => Err(ValkeyError::Str(utils::NOT_FOUND)),
```
zackcam: Should this be `ValkeyError::nonexistent_key()` to match the other commands as well

src/cms/cms_command_handler.rs
```diff
@@ -242,3 +242,33 @@ pub fn cms_query(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
         }
     }
 }
+
+//Function that implements logic to handle the CMS.INFO command.
+pub fn cms_info(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
+    let args_count = args.len();
+    if args_count != 2 {
```
zackcam: Weird that specific info fields are not allowed to be called individually to be got. I wonder if we would want to expand this as it wouldn't break functionality and would make it consistent for the other probabilistic data types as well. Let me know what you think

src/cms/utils.rs
zmccoy: Those should be able to be removed, good catch!

src/cms/cms_command_handler.rs
zmccoy: Agreed, i'll get those changed.

src/cms/cms_command_handler.rs
zmccoy: Yes, that would definitely be more correct.  I'll get that updated.

src/cms/cms_command_handler.rs
zmccoy: I'd be up for trying that out.  Do you have an example that you're thinking of?  I assume another parameter that is optional that is one of the 3 field names and the return then just being that specific one?

src/cms/cms_command_handler.rs
zmccoy: Fixed in [Fix comments. Use existing error for conformity. lowercase output. no…](https://github.com/valkey-io/valkey-bloom/pull/116/commits/6d0d364f3758116a08e5380970cbebb674df234c)

src/cms/cms_command_handler.rs
zmccoy: Fixed in [Fix comments. Use existing error for conformity. lowercase output. no…](https://github.com/valkey-io/valkey-bloom/pull/116/commits/6d0d364f3758116a08e5380970cbebb674df234c)

src/cms/utils.rs
zmccoy: Fixed in [Fix comments. Use existing error for conformity. lowercase output. no…](https://github.com/valkey-io/valkey-bloom/pull/116/commits/6d0d364f3758116a08e5380970cbebb674df234c)

zmccoy:
Looks like the tests failed on an unrelated change (BF command test) and passed on my repo / local branch.

zackcam:
> Looks like the tests failed on an unrelated change (BF command test) and passed on my repo / local branch.

Yeah unrelated failures, we have a fix merged already which looks to just be missing on the CMS branch we have here:
https://github.com/valkey-io/valkey-bloom/commit/74d7eb17c9ef25d36223120f46fd406ec5d52862. I've hit rerun so should be fine

KarthikSubbarao:
@coderabbitai review

src/cms/cms_command_handler.rs
zmccoy: Also happy to enhancements into another branch, just like the extra changes for INCRBY etc to help close loops.

zackcam: It would kinda be like bloom here, https://github.com/valkey-io/valkey-bloom/blob/unstable/src/bloom/command_handler.rs#L733. Where we give the optional argument so like the one extra parameter like you said

zmccoy: This should be done here now:  [Add params for info single value grabs, add tests around it and failures](https://github.com/valkey-io/valkey-bloom/pull/116/commits/e9e08d6b9700dedc144d32c026558772b6f013c3) 

Let me know if there's anything else

zmccoy:
Build failure was unrelated to the changes, and from a flakey pull.  Let me know if there's anything else we should fix up here.   Thanks again!

zackcam (approved):
Will have a few follow up asks for tests but as its a side branch merging in to unblock

PR #117, opened 2026-08-05
Vendor heavykeeper and optimize the TopK sketch

detemmienation:
## Summary

Vendors the `heavykeeper` crate into `heavykeeper/` as a local path dependency, then applies a series of memory and lookup optimizations to the TopK sketch (`CuckooTopK`): u32 cells, single-copy priority queue, u32 sequence, no decay table, and a default linear-scan lookup.

See [heavykeeper/MODIFICATIONS.md](heavykeeper/MODIFICATIONS.md) for the change-by-change rationale and per-commit links.


heavykeeper/src/cuckoo.rs
```diff
@@ -772,6 +771,140 @@ impl<T: Ord + Clone + Hash> CuckooTopK<T> {
...
+
+        // Rebuild the seed-derived state, then graft the cells and queue on top.
+        // Reserve the original `top_items` capacity so a round-trip is identical
+        // in size to the original.
+        //
+        // OOM TRADEOFF: unlike the cell/key allocations (gated by `take`), this
+        // is sized by the unbounded `top_items` header, so a tampered stream
+        // could force a huge reserve. Fine for restoring our own dumps (RDB); if
+        // ever fed untrusted bytes, bound `top_items` first.
+        let mut sketch = Self::with_seed(top_items, width, depth, decay, seed);
```
zackcam, line 887: We should bound this for the reserve. RDB's can't always be trusted. Can reinforce this on the module side eg around here we should check: 
```
        let sketch = match CuckooTopK::<Vec<u8>>::from_bytes(sketch_bytes.as_ref(), seed) {

```

heavykeeper/src/cuckoo.rs
```diff
@@ -772,6 +771,140 @@ impl<T: Ord + Clone + Hash> CuckooTopK<T> {
...
+    /// ```
+    ///
+    /// The seed is not stored; the hasher is rebuilt from the seed passed to
+    /// [`from_bytes`](CuckooTopK::from_bytes). A `hasher_probe` guards against a
+    /// wrong seed: it is the hash of a fixed value, re-checked on load. The RNG
+    /// position is stored (`rng_state`) and restored exactly.
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let cell_count = self.lobbies.len() + self.heavy.len();
+        let pq_len = self.priority_queue.len();
+        // Capacity hint: header (magic + variant + version + probe + 5 u64s) + cells + pq_len u64 + pq estimate + rng.
```
zackcam, line 797: Might be 6 u64s here

heavykeeper/src/serialization.rs
```diff
@@ -0,0 +1,333 @@
...
+    TrailingBytes { count: usize },
+}
+
+/// Magic tag at the start of every serialized sketch (`b"HVYK"`).
+pub(crate) const MAGIC: [u8; 4] = *b"HVYK";
+/// On-disk format version. Bump whenever the byte layout changes.
+pub(crate) const VERSION: u8 = 1;
+/// Probe hashed at serialize time to detect a wrong seed on load.
+///
+/// `ahash` output is not stable across CPU architectures or `ahash` versions,
```
zackcam, line 50: ahashcausing rdbs to be architecture specific isn't ideal. We may need to switch to a seeded hash

zackcam (commented):
Adding notes on areas to check before release: serialization comments

zackcam (commented):
Add heavykeeper defrag support - Looks good

zackcam (commented):
Deserialization changes might need to come in for this where we allow different sized cells - Add generic Fingerprint/Counter types for CuckooTopK

zackcam (commented):
Refactor TopKQueue to single-copy storage, removing the free-slot

Seems good

zackcam (commented):
Narrow TopKQueue sequence to u32
 

Also seems good

PR #120, opened 2026-08-31
CMS :  Merge Command

zmccoy:
Cherry picked from the older branches that were in a rebase stack, this adds the merge command and testing around it for the Count Min Sketch.  This is the last command we were missing.  

Questions:  
1. I'm a bit unclear on replication, and it seems like I'd not want to do verbatim, but am curious as to how to think about that.  

Notes: 
1. We are still missing the replication testing as a whole, I'm happy to add that here or in another PR after this is done and fix up from there.

### DIFF
--- a/src/cms/cms_command_handler.rs
+++ b/src/cms/cms_command_handler.rs
@@ -256,7 +256,10 @@ pub fn cms_merge(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
     //This must already be initialized.
     let destination_key = &args[1];
 
-    let number_of_keys_value = args[2].to_string_lossy().parse::<usize>()?;
+    let number_of_keys_value = args[2]
+        .to_string_lossy()
+        .parse::<usize>()
+        .map_err(|_| ValkeyError::Str("ERR invalid number of keys value"))?;
 
     //Indexes 3 -> 3 + N-1 are keys to merge
     let sketch_end_index = 3 + number_of_keys_value - 1;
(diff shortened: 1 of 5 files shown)


src/cms/cms_command_handler.rs
```diff
@@ -238,3 +245,88 @@ pub fn cms_query(ctx: &Context, args: Vec<ValkeyString>) -> ValkeyResult {
...
+    let sketches_result: Result<Vec<&CMSObject>, ValkeyError> = source_key_handles
+        .iter()
+        .map(|key_handle| {
+            key_handle
+                .get_value::<CMSObject>(&CMS_TYPE)
+                .and_then(|opt| opt.ok_or_else(|| ValkeyError::Str("ERR key does not exist")))
+        })
+        .collect();
+
+    let sketches: Vec<&CMSObject> = sketches_result?;
```
zmccoy: I may be able to turbo shark like the above to get this on line 305 as well...

zmccoy: Fixed [Use turbo shark to use ? on the same line as collect](https://github.com/valkey-io/valkey-bloom/pull/120/commits/3249f681f79f8400f0b7233b22a2e0d22277edfa)

zackcam:
I owuld think verbatim replication is what we want as consistency is key across replica and primaries. Was there a place where you saw this wasn't the case?

zmccoy:
@zackcam Nothing specific to why it wouldn't work.  I think I was pattern matching off the previously done INITBYDIM and INITBYPROB which were not verbatim and followed a similar structure of not having to omit anything.  

For my own edification, would INITBYDIM and INITBYPROB also be able to be verbatim?  For example: https://github.com/valkey-io/valkey-bloom/blob/count-min-sketch/src/cms/cms_command_handler.rs#L46-L47 ends up recreating the `CMS.INITBYPROB key error probability`

